Issue #160, opened 2015-02-27
Enhancement: add ability to swipe with offset (MobileElement).


TikhomirovSergey:
[SwipeElementDirection](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/SwipeElementDirection.java) and [MobileElement](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/MobileElement.java) here should be implemented additional swipe-method with the offset as a parameter.

TikhomirovSergey:
#159

TikhomirovSergey:
Changes have been merged.

PR #163, opened 2015-03-06
#160 fix

TikhomirovSergey:
Change list:
- Enum SwipeElementDirection is redesigned. The ability to perform the swiping was with offset was added.
- The swiping using offset from element borders was added to MobileElement.

### DIFF
--- a/src/main/java/io/appium/java_client/TouchableElement.java
+++ b/src/main/java/io/appium/java_client/TouchableElement.java
@@ -55,13 +55,15 @@ public interface TouchableElement extends WebElement {
 	   * 
 	   * @param direction direction UP, DOWN, LEFT, RIGHT
 	   * 
-	   * @param offset1 is the offset from the border of the element. If direction is UP then
+	   * @param offsetFromEndBorder is the offset from the border of the element where the swiping should be started. 
+	   * If direction is UP then
 	   * this is offset from the bottom of the element. If direction is DOWN then
 	   * this is offset from the top of the element. If direction is RIGHT then
 	   * this is offset from the left border of the element. If direction is LEFT then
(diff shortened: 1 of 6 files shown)


TikhomirovSergey:
#160

src/main/java/io/appium/java_client/TouchableElement.java
```diff
@@ -1,52 +1,76 @@
...
+      * @param direction UP, DOWN, LEFT, RIGHT
+      * 
+      * @param duration amount of time in milliseconds for the entire swipe action to
+		 *            take
+      */
+	  public void swipe(SwipeElementDirection direction, int duration);	  
+	  
+	  
+	  /**
+	   * Convenience method for swiping on the given element to the given direction
```
Jonahss: What's the difference between offset1 and offset2? Do you mean to say that offset1 is offset from the element that swipe starts on, and offset2 is from the element that the swipe ends on?

TikhomirovSergey: offset1 and offset2 are offsets from borders of the element.  
offset1 is the offset from the border where the swiping should be started. 
offset2 is the offset from the border where the swiping should be ended.
If parameter names are not clear how can I rename them?

TikhomirovSergey:
Ok. I am going to rename __offset1__ and __offset2__ to __offsetFromStartBorder__ and __offsetFromEndBorder__. Is it ok?

Fanch-:
For me it's clear.

TikhomirovSergey:
Ok.

TikhomirovSergey:
I've improved parameter names and comments.

Jonahss:
Great :)

Issue #162, opened 2015-03-04
Make findElement return MobileElement


alexec:
I noticed in a presentation yesterday that they almost always casted to MobileElement. As MobileElement appears to be a sub-class of WebElement, you can modified the return type of findElement* to return MobileElement.

This would be backwards compatible. 

If we want to get technical, we are dealing with covarient return types:

http://c2.com/cgi/wiki?CovariantReturnTypes

Jonahss:
I *totally* agree!
For some reason when I tried to override the method originally I ran into compilation errors, but looking at it now it seems fine.

Care to submit a Pull Request?

TikhomirovSergey:
I think the better way is to propose PR to Selenium project. I've got the great idea how to implement this. If this PR will be rejected then it is easy to override __find__ methods here.

So I think we should work on two these directions (I'm sorry for my English).

TikhomirovSergey:
PR proposed to Selenium: https://github.com/SeleniumHQ/selenium/pull/327
If you consider that it is good please vote for this.

Jonahss:
Great!
I've found that mentioning your pull request on the #selenium irc can get it merged quicker ^.^

TikhomirovSergey:
I thick this ticket should be closed

PR #167, opened 2015-03-14
#162 fix

TikhomirovSergey:
- interfaces FindsByAndroidUIAutomator, FindsByAccessibilityId and FindsByIosUIAutomation were changed 
- findElement(By) and findElementBy* were overriden. Now they return MobileElement or subclass.

I have tried to override findElements(By) and findElementsBy*. But there are casting problems of java-collections. Also there is a problem of backward compatibility. Sometimes I hate Java :)

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -71,7 +71,43 @@ public abstract class AppiumDriver extends RemoteWebDriver implements MobileDriv
 		return dc;
 	}
 
-	/**
+    public MobileElement findElement(By by){
+        return (MobileElement) super.findElement(by);
+    }
+
+    public MobileElement findElementById(String using){
+        return (MobileElement) super.findElementById(using);
+    }
+
+    public MobileElement findElementByClassName(String using){
+        return (MobileElement) super.findElementByClassName(using);
+    }
+
+    public MobileElement findElementByName(String using){
+        return (MobileElement) super.findElementByName(using);
+    }
+
+    public MobileElement findElementByTagName(String using){
+        return (MobileElement) super.findElementByTagName(using);
+    }
+
+    public MobileElement findElementByCssSelector(String using){
+        return (MobileElement) super.findElementByCssSelector(using);
+}
+
+    public MobileElement findElementByLinkText(String using){
+        return (MobileElement) super.findElementByLinkText(using);
+    }
+
+    public MobileElement findElementByPartialLinkText(String using){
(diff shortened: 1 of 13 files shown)


TikhomirovSergey:
#162

Jonahss:
Ha! I knew that I had tried to solve #162 when I first wrote the java-client and then decided it wasn't possible, but #162 convinced me that I was just being dumb in the past.

You don't think it will be confusing to users if indElements(By) and findElementsBy* aren't consistent with the others?

TikhomirovSergey:
Ok. I've just proposed it. 

If you consider that this PR is not consistent please close it without merging. Actually I think it is not crucial that users can't get MobileElement without casting.

>but #162 convinced me that I was just being dumb in the past.
I am sorry if I made you think so. I was too optimistic. :)

Jonahss:
Hm. The convenience of not having to cast to MobileElement vs the confusion of having to cast when using plural finders. What does @bootstraponline think?

bootstraponline:
I like it. I think the code is cleaner without having to cast all the time.

TikhomirovSergey:
@bootstraponline 
...but the necessity to cast an element of the list is still here. Please pay attention.

```java
List<WebElement> I = driver.findElements(by);
MobileElement m = (MobileElement ) l.get(0); 
```

The architecture of RemoteWebDriver solution and Java by itself are both prevent the following use case

```java
List<MobileElement> I = driver.findElements(by);
```

and 

casting

bootstraponline:
I think overall, it'll be less casting for the common use case of finding and acting upon an element.

```diff
-	  MobileElement e = (MobileElement) driver.findElement(MobileBy.AccessibilityId("App"));
+	  MobileElement e = driver.findElement(MobileBy.AccessibilityId("App"));
```

Casting in the list.get(0) case is pretty common in Java.

bootstraponline:
It's your PR so if you think it's better not to merge, that's fine. I don't feel strongly either way.

TikhomirovSergey:
Ok @bootstraponline 
I think this PR is could be merged because these changes really WORK. But I don't mind if it won't be merged because this solution LOOKS like something one-legged. 

So, let @Jonahss make a desicion.

Actually, today or tomorrow I am going to propose the similar changes to C#. And it is possible to provide both usecases:

```java
MobileElement e = driver.findElement(by);
```    

```java
List<MobileElement> l = driver.findElements(by);
```

Jonahss:
This is exactly the sort of thing I don't like about Java, but I'm ok with merging this ^.^
Thanks for all the hard work @TikhomirovSergey

TikhomirovSergey:
Thank you @Jonahss

PR #177, opened 2015-03-24
junit dependency is not required for compile scope

NozomiIto:
Now I develop a Java test tool which depends on the appium java-client library.
My test tool supports both JUnit3 and JUnit4, and I don't want the appium java-client to include junit4 dependency since this dependency may cause conflict.

I think appium java-client requires junit4 dependency only for test scope.


Jonahss:
Ah wonderful. I'm sure this will help others as well. Thanks!

PR #181, opened 2015-04-01
Documentation issue fix

lukian-tabandzhov:
Hi,

Just a minor documentation fix in case we need to exclude the selenium version on which the java-client depends on.

Cheers


Jonahss:
thanks :)

PR #186, opened 2015-04-16
make sure that finalize calls on annotated fields in page objects don't send findElement calls

aluedeke:
page objects do tons of findElement(s) calls to appium server because they also forward finalize method invocations.

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/ElementInterceptor.java
+++ b/src/main/java/io/appium/java_client/pagefactory/ElementInterceptor.java
@@ -23,6 +23,10 @@ class ElementInterceptor implements MethodInterceptor {
 	
 	public Object intercept(Object obj, Method method, Object[] args,
 			MethodProxy proxy) throws Throwable {
+        if(Object.class.getDeclaredMethod("finalize").equals(method)){
+            return proxy.invokeSuper(obj, args);  //invokes .finalize of the proxy-object
+        }
+
 		WebElement realElement = locator.findElement();
 		return method.invoke(realElement, args);
 	}
--- a/src/main/java/io/appium/java_client/pagefactory/ElementListInterceptor.java
+++ b/src/main/java/io/appium/java_client/pagefactory/ElementListInterceptor.java
@@ -12,10 +12,12 @@ import net.sf.cglib.proxy.MethodInterceptor;
 import net.sf.cglib.proxy.MethodProxy;
 
 /**
+ *
  * Intercepts requests to the list of {@link MobileElement}
  *
  */
 class ElementListInterceptor implements MethodInterceptor{
+
 	private final ElementLocator locator;
 	
 	ElementListInterceptor(ElementLocator locator){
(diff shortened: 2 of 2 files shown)


TikhomirovSergey:
@Jonahss  It is the useful PR. I've never faced any problem with objects proxied (lazy instantiation like that) by CGLIB and  garbage collector. But now I've got the message that some problems are possible.

I have some remarks:
-  @aluedeke Maybe It's better to make improvement like that

```java
...
if(Object.class.getDeclaredMethod("finalize").equals(method)){
    locator.finalize();
    return proxy.invokeSuper(obj, args);  //invokes .finalizy of the proxy-object
}
...
```

aluedeke:
@TikhomirovSergey updated the PR

actually i didn't had any functional issues with it. It created just more load on the appium serve  and created bloat logs which probably no one will understand

TikhomirovSergey:
Have you tried to launch tests or check your scripts when your project depends on locally prepared snapshot? Does this improvement really help?

Jonahss:
Any reason we shouldn't merge @TikhomirovSergey ? Looks good to me, but I don't rightly understand the code around page object models :)

aluedeke:
@TikhomirovSergey yes improved it a lot. Had PageObjects with 20+ annotated fields and every finalize call created a findElement call (so 20+). Now i have zero. So its much better/cleaner.

Hope no one will ever put this stuff into hashmaps so that equals, hashcode or toString are frequently called.

TikhomirovSergey:
@Jonahss 

Lazy instantiation of elements is the one of the main principles of page object model in Appium as well as in Selenium. It doesn't create elements directly invoking  __find*__. It creates proxies - synthetic objects.
They implement interfaces of desired class (it is done in Selenium by default using java.reflect.Proxy API) or (it looks like) their class extends the target class (it is done in Appium). Inside these proxies some additional logic is implemented and the real target method is invoked.

Let's look at this:

```java
WebElement realElement = locator.findElement(); 
return method.invoke(realElement, args);
```
Here it finds the real element and invokes the target method. __locator__ contains the given search context (WebElement or WebDriver). Locator-object contains once defined SearhContext all its lifetime. This implementation is similar as in Selenium. 

I'm suspecting that when GC invokes _finalize_ of the proxy it still tries to find an element (!!!) and invoke its _finalize_. This can throw an exception when defined Searchcontext (especially WebDriver) is already dead (__quit()__ method was invoked before) on some conditions.  But why nobody has found it? 

Let me check my assumption. If I will proof it then we can merge this PR. 

@Jonahss There are lots of things to explain. Have you worked with __java.reflect.Proxy__, __Spring__, __AspectJ__, __CGlib__?

TikhomirovSergey:
It looks like @aluedeke have prooved my assumption.

aluedeke:
@TikhomirovSergey your assumption is right GC calls finalize on proxy, proxy calls findElement.

BTW i can put the proxy object after creation into a Map (which hopefully no one will ever do)

Jonahss:
@TikhomirovSergey I haven't worked with java.reflect.Proxy, Spring, AspectJ, or CGlib and I don't really consider myself a professional Java developer. It's fine though, I defer to your judgement when it comes to handling this aspect of the java-client ^.^

This is why open source software works! I don't need to learn all that if you know it and I trust you :)

TikhomirovSergey:
@aluedeke If you have a static screen (element set is not frequently refreshed) you can avoid the excessive searchings this way. 

```java
@CacheLookup //**<-allows to cache the element that once was found
@AndroidFindBy(yourLocator) 
RemoteWebElement someElement;
```
Also I think there are special setting of your environment. Your GC cleaning is being run frequently. Or test is being run so long time :)

@Jonahss let's merge it. All that I've suspected really happens. But this doesn't affect users just because generally GH cleaning goes in another thread generally. So it doesn't affect users...generally :)

PR #190, opened 2015-04-28
added a method "scrollTobyCellIndex" to scroll by index on IOSDriver and IOSElement class

sahajamit:
Actually in iOS, ScrollTo method is defined to scroll over a TableView which has multiple cells. It is always recommended to use it over the UIATableView and it scroll to the respective cell of this table depending the search text is present in the cell name.

If we have cells with duplicate names then there is no way to scroll to each of these cells. So I have added a new method called "scrollTobyCellIndex" in the IOSElement class where you can pass the index of the cell and it will scroll to that specific cell.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 
     <groupId>io.appium</groupId>
     <artifactId>java-client</artifactId>
-    <version>2.2.0</version>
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
I'm closing it according to #385

Issue #201, opened 2015-06-09
Update selenium dependency to 2.46


Jonahss:

Jonahss:
fixed by #206

PR #205, opened 2015-06-15
update to Selenium 2.46 + Some improvements.

TikhomirovSergey:
Changes:
- update dependencies to Selenium 2.46
-  I have found a problem recently. Page object fields couldn't be instantiated because a customized implementation of the SearhContext was used. So now it is possible to use classes like that. But user's SearchContext implementations have to implement WrapsElement or WrapsDriver too.

PR #206, opened 2015-06-16
#201 fix update to Selenium 2.46 + Some improvements.

TikhomirovSergey:
Changes:

- update dependencies to Selenium 2.46

-I have found a problem recently. Page object fields couldn't be populated because a customized implementation of the SearhContext was used. So now it is possible to use classes like that as well as
WebDriver and WebElement. But user's SearchContext implementations have to implement WrapsElement or WrapsDriver too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Depends upon the Selenium Java client library, available [here](http://docs.sele
 </dependency>
 ```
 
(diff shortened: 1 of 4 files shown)


TikhomirovSergey:
#201 
It duplicates #205. This one was unchancy

Jonahss:
Nice!

PR #213, opened 2015-07-05
Fix #184 Fix #211

### LINKED ISSUE #184: @AndroidFindBy and @iOSFindBy for native app automation
As suggested in https://github.com/appium/java-client/pull/68#issuecomment-93260061:

We should have the annotations @AndroidFindBy and @iOSFindBy support for native apps also. Currently it supports only for the web view.

I would also suggest from my point of view to make AppiumDriver as interface instead of keeping it as abstract.

### LINKED ISSUE #211: iOSFindBy & AndroidFindBy annotations are ignored
I've been trying for a couple of days to get iOSFindBy & AndroidFindBy annotations working with page objects and the PageFactory. I've tried versions 2.2.0 and 3.0.0 of the java-client and tests for Android and iOS, all with the same result - which is the same as not including any annotations.

Here is a truncated version of one of my page objects

```java
public class HomePhonePage extends BaseHomePage {

    public HomePhonePage(AppiumDriver driver) {
        super(driver);

        PageFactory.initElements(new AppiumFieldDecorator(driver,
                        IMPLICIT_WAIT,
                        TimeUnit.SECONDS),
                this
        );
    }

    @iOSFindBy(accessibility = "home_landing_my_favorites_count_label")
    private WebElement myFavoritesCountLabel;

    public String getMyFavoritesCountLabel() {
        return myFavoritesCountLabel.getText();
    }
}
```

My test code:
```java
    @Test
    public void testHomeIsValid() {

        homePhonePage = new HomePhonePage(driver);
        globalNavPage = new GlobalNavigationPage(driver);

        homePhonePage.clickHamburgerMenu(); //FAILS HERE because element was not found
        globalNavPage.clickHome();

    }
```
When I debug into PageFactory.initElements, I see [unknown locator] for each property that is a WebElement. Then the PageFactory defaults to looking for each page object's property name as the id or name of the WebElement and fails to find anything.

unknown-locator2

I suspect "I'm doing it wrong" but I can't figure out where. It is behaving as if the annotations are not even there.

TikhomirovSergey:
Change list:
- fixed #184 and #211. Now the choosing is by-strategy is depending on only driver class. There was outdated code. Now it is updated. Summary:
So when here is __AnroidDriver__ and current context is native app then will be choosen By-strategy defined via __AndroidFindBy/AndroidFindBys/AndroidFindAll__ annotations.  When here is __IOSDriver__ and current context is native app then will be choosen By-strategy defined via __iOSFindBy/iOSFindBys/iOSFindAll__. annotations.
By-strategy that is built using defaul __FindBy/FindBys/FindAll__ will be used for browser/webview content or when there is no specific annotations for Android/iOS native apps.

There is one specific remark. If somebody desires to use their own Webdriver implementation then it has to implement HasCapabilities.

- AutomationName has been added in order to define the desired automation type correctly in Capabilities.

- Now the class ContentMappedBy has overridden __toString()__ method

- I fixed a problem that is not reported yet. 

How to reproduce:

```java
//Right now we are testing an Android app

@iOSFindBy(iOS locator)
@FindBy(css = "e.e1.e2")
private e;
```

or

```java
//Right now we are testing an iOS app

@AndroidFindBy(className = "android.widget.TextView")
@FindBy(css = "e.e1.e2")
private WebElement e;
```

At these cases By-locator defined via __FindBy__ is used. If we are testing a native app the test is failed because of UnsupportedCommandException. CSS and link text are not supported by automation tools. 

So now NoSuchElementException will be thrown an this case insted of UnsupportedCommandException. This minor bug here since 1.7.0 version.

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -106,8 +106,9 @@ class AppiumElementLocator implements ElementLocator {
         if (IOSDriver.class.isAssignableFrom(driverClass))
             return MobilePlatform.IOS;
 
-        //it is possible that here customers implementation of WebDriver
-        //At this case it has to implement HasCapabilities
+        //it is possible that somebody uses RemoteWebDriver or their
+        //own WebDriver implementation. At this case capabilities are used
+        //to detect platform
         if (HasCapabilities.class.isAssignableFrom(driverClass))
             return String.valueOf(((HasCapabilities) d).getCapabilities().
                     getCapability(MobileCapabilityType.PLATFORM_NAME));
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -1,5 +1,7 @@
 package io.appium.java_client.pagefactory;
 
+import io.appium.java_client.android.AndroidDriver;
+import io.appium.java_client.ios.IOSDriver;
 import io.appium.java_client.remote.MobileCapabilityType;
 
 import java.lang.reflect.Field;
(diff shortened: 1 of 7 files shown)


src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -88,6 +93,41 @@ private WaitingFunction(SearchContext searchContext) {
...
+            return null;
+
+        Class<?> driverClass = d.getClass();
+        if (AndroidDriver.class.isAssignableFrom(driverClass))
+            return MobilePlatform.ANDROID;
+
+        if (IOSDriver.class.isAssignableFrom(driverClass))
+            return MobilePlatform.IOS;
+
+        //it is possible that here customers implementation of WebDriver
```
Simon-Kaz: Not sure what you mean here - is this if statement handling custom drivers?

Jonahss:
Great :)

PR #230, opened 2015-08-25
Add abstract sendKeyEvent into AppiumDriver

laviua:
It needs, because we're working with parent AppiumDriver and don't have ability to use child's sendKeyEvent method for Android

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -717,12 +717,14 @@ public abstract class AppiumDriver<RequiredElementType extends WebElement> exten
 		return remoteAddress;
 	}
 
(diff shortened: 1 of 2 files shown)


Jonahss:
Can't you cast your AppiumDriver as an AndroidDriver?
```
AppiumDriver driver = new AndroidDriver();
((AndroidDriver)driver).sendKeyEvent(5);
```
?

laviua:
I can, but it looks not pretty good as I think. I thought that you missed
this method in parent class :)
Thanks

Jonahss:
Ah. Well I don't think we missed it, because iOSDriver does not have that method, so there is no need to have it on the parent. The parent AppiumDriver class should only have methods which are implemented by both Android and iOS

PR #235, opened 2015-08-31
Exposed replaceValue which is available in appium and python client i…

ahansen1:
…n the java-client.

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -38,6 +38,7 @@ import static io.appium.java_client.MobileCommand.PULL_FILE;
 import static io.appium.java_client.MobileCommand.PULL_FOLDER;
 import static io.appium.java_client.MobileCommand.PUSH_FILE;
 import static io.appium.java_client.MobileCommand.REMOVE_APP;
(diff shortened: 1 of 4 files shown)


TikhomirovSergey:
Hi, @ahansen1 
Your contribution is appreciated but I can't make a decision. Let's wait for @Jonahss 

Now your changes are in conflict with the master. The conflicted code:
https://github.com/appium/java-client/pull/235/files#diff-c60f84c0cc90f37a9a4513eeebd4896dR221

The command map filling has been moved here:
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/AppiumDriver.java#L666

ahansen1:
I believe that I've resolved the conflict and it is mergable again.  

Another approach would be that setValue(or /value) does replacement and that sendKeys (or /keys) does appending.  Though, I thought that may be a bigger deal because it would not be backward compatible with what is in the wild.  I found in the appium server code and the python client that replaceValue had been added which is why I was exposing it here.

TikhomirovSergey:
Yep. Everything is allright. Let's wait for @Jonahss

Jonahss:
Looks great, thanks very much @ahansen1 
I'll add you to the authors list :)

PR #237, opened 2015-09-01
adding SEARCH keycode to AndroidKeyCode

bgordon-applause:


bgordon-applause:
for reference:  http://developer.android.com/reference/android/view/KeyEvent.html#KEYCODE_SEARCH

bootstraponline:
`int KEYCODE_SEARCH = 84;` already exists. Is that not sufficient?

bgordon-applause:
Perfectly sufficient, apologies for not noticing before issuing PR.  Will close the PR, now.  Thanks!

PR #246, opened 2015-09-16
maven surefire to run unit tests

GiannisPapadakis:
1. Surefire plugin added to run unit tests on integration-test phase
2. ReentrantLock used in AppiumDriverLocalService


TikhomirovSergey:
Hi @GiannisPapadakis 
I like the new maven configuration :+1: 

Please provide more details about the reason why you use the lock instead of synchronized methods.

GiannisPapadakis:
I think a ReentrantLock is unstructured, unlike synchronized constructs for we don't need to use a block structure for locking and can even hold a lock across methods. Furthermore ReentrantLock  is more scalable (in heavy load of ThreadPoolExecutor it performs better).

TikhomirovSergey:
Ok. I have nothing abainst change like that.

There is a minor issue. Please provide the locking here:
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java#L75

And everything will be completed :)

GiannisPapadakis:
Done!!!

TikhomirovSergey:
Ok. Today I'm going to check and merge this. Thanks!

Jonahss:
Awesome.

Issue #239, opened 2015-09-08
SendKeys with accent and others characters is not working.


Vanilton18:
Hello,

I am sending texts in Portuguese in which they have accents like: á, ão, ª, º, ° , à, ¨, soon identified problems sending the SendKeys method words or text containing the values already mentioned.

Example:

Input:
String word = "ão";
for(int i=0; i<word.length(); i++){
driver.findElementById("foo").sendKeys("letter"+word.charAt(i));
}

Output

""

No exception is thrown, only the text is not sent.

Appium java-client Version: 3.1.0
Device tested: Samsung_SM-T210 
Android Version: 4.4.2

Fanch-:
Did you try with 
```java
capabilities.setCapability("unicodeKeyboard", true); 
```
?
Care it modifies your keyboard settings.

Vanilton18:
Thank you! I did not know this capability, she could be in MobileCapabilityType interface. Apparently solved the problem. Another problem identifying return here.

TikhomirovSergey:
@Vanilton18 I can add this to MobileCapabilityType soon.

Fanch-:
@Vanilton18 I got issue with this capability though, like filling phone number fields. Otherwise it was fine. To retrieve previous state : 
```java
capabilities.setCapability("resetKeyboard", true);
```

Vanilton18:
Ok @Fanch- I will perform the test battery that I have to check but there were problems.

@TikhomirovSergey Thanks for disposal.

Thank you for your help.

PR #251, opened 2015-09-29
Update MobileCapabilityType.java

eliasnogueira:
capability to accept accent from keyboard to edittext component
see https://github.com/appium/java-client/issues/239


TikhomirovSergey:
I like this change :+1: 
@Jonahss What's your opinion?

Jonahss:
Thanks for your contribution :)

PR #261, opened 2015-10-23
Updated startActivity() method to allow starting activities without stopping target app.

deadmoto:
The reason behind is that Appium always stops target app when startActivity() method is called.
That makes impossible to start activities that belong to app under test without killing it.
This PR introduces new optional backward-compatible parameter to control this behavior.


TikhomirovSergey:
:+1: 
@deadmoto  Can you add some tests?
@Jonahss What's your opinion?

deadmoto:
@TikhomirovSergey, I assume test must be isolated and use only ApiDemos-debug.apk?

TikhomirovSergey:
@deadmoto 

Yep. They should be isolated (not dependent). We usually use ApiDemos-debug.apk. Somewhere default activities of a device/emulator are used. You can look at 
https://github.com/appium/java-client/blob/master/src/test/java/io/appium/java_client/android/AndroidDriverTest.java

deadmoto:
@TikhomirovSergey It looks like there is a test related to this PR - startActivityInThisAppTest().
However, it seems like the test does not actually verify anything but the fact that activity has started.
I will invest some time to get a better solution.

TikhomirovSergey:
ok.

Jonahss:
Does this work? the `dontStopAppOnReset` is a desired capability that has to be added when a session is started. Adding it to the `startActivity` command won't affect Appium unless this capability has been implemented there. Is this in the current Appium, or are you working on a Pull Request?

deadmoto:
@Jonahss, `startActivity` method uses `startApp` method and just passes request parameters as an argument.
`startApp` is implemented as following:
```
Android.prototype.startApp = function (args, cb) {
  if (args.androidCoverage) {
    ...
  } else {
    this.adb.startApp({
      ...
      stopApp: args.stopAppOnReset || !args.dontStopAppOnReset
    }, cb);
  }
};
```
This pair of arguments is being used to control the behavior.
And yes, I need extra time to implement test for that.

Jonahss:
Cool, looks good then.

deadmoto:
@TikhomirovSergey, @Jonahss I covered both behaviors with elaborate test.
I had to update ApiDemos-debug.apk to run original tests so I pushed it as well.

TikhomirovSergey:
:+1: 
It looks ok. But I have a remark. It looks like the new version of the test ApiDemos-debug.apk breaks two tests:
https://github.com/appium/java-client/blob/master/src/test/java/io/appium/java_client/android/AndroidGestureTest.java#L76
https://github.com/appium/java-client/blob/master/src/test/java/io/appium/java_client/android/AndroidUIAutomatorTest.java#L102

because ids of elemends are probably were changed. Can you look at them and synchronize them with the new .apk if you are facing the same problem?

I have nothing this change. It is expected it will be published soon. I'm working on one more interesting feature too :)

@Jonahss Please look at this :)

deadmoto:
@TikhomirovSergey, @Jonahss I have updated these tests.

TikhomirovSergey:
@deadmoto 
It is ok. :+1: 
Let's @Jonahss take a look at your changes and merge them.

Jonahss:
Cool.
I guess after @TikhomirovSergey's next big feature we should publish a version.

Issue #264, opened 2015-11-02
AppiumLocalService needs to have OutputStream parameter for logging


smartinovic:
Currently AppiumLocalService logs to System.err and it is not changeable. It would be better if there is OutputStream parameter for streaming the logs.

TikhomirovSergey:
@smartinovic 
Do you want to save logs to file? Do I clearly understand your problem?

smartinovic:
@TikhomirovSergey 
I don't have problem saving logs to file, for this I use existing method withLogFile().
Problem is that I don't have freedom to set output stream for appium server logs, now they are always logging to error stream and cluttering output of my application.
If that is parametrised, I would be able to choose to write logs into System.err, System.out, log4j or choose any other way that I want to handle these logs.

TikhomirovSergey:
Ok. I'll add the requiled abilty soon.
Do you want this method

> servise.setOutputStream(OutputStream stream)

?

smartinovic:
That looks right. Thanks

PR #266, opened 2015-11-11
#264 fix

TikhomirovSergey:
Change list:
- an abilty to use few output streams during a server session was added

```java
service.addOutPutStream(stream);
```

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.ReentrantLock;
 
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
@Jonahss @bootstraponline please take a look at this

TikhomirovSergey:
#264

smartinovic:
@TikhomirovSergey Could you also add method: setOutPutStream(stream) that will reset ListOutputStream() and add that stream:
```java
public void setOutPutStream(OutputStream outputStream){
        checkNotNull(outputStream);
        stream = new ListOutputStream().add(outputStream);
}
```

It is valuable if I don't want for stream to be outputted in System.err, and you can use it on line 47:
```java
private final ListOutputStream stream = setOutPutStream(System.err);
```

TikhomirovSergey:
@smartinovic 
It means that by default it will print server output to console. When you'll add some another outputs when information will go to console and to another ...files, pipelines and so on. 

I think that majority of users want to see output on console at all cases. What if I'll:
- replace System.err by System.out 
- add some new methods

```java
service.addOutPutStreams(List<OutputStream> streams); 
```

?

smartinovic:
Sounds good, it will do the trick for me.

TikhomirovSergey:
Additional changes:
- System.err was replaced by System.out
- The additional method addOutPutStreams(List<OutputStream> outputStreams)

bootstraponline:
Looks reasonable. :+1:

TikhomirovSergey:
So I'm merging it

Issue #272, opened 2015-11-20
A problem of the not common Node.js starting


TikhomirovSergey:
There are problems with the starting of Node.js apps such as [NSolid](https://nodesource.com/products/nsolid)

Jonahss:
Fixed and merged :)

Added environment variable for Node.js binary path: NODE_BINARY_PATH

PR #273, opened 2015-11-23
#272 fix

TikhomirovSergey:
Change list:
- the main change is the bug fix of #272. This thing was breaking the starting of a Node.js instances which differ from default (e.g. [NSolid](https://nodesource.com/products/nsolid))

- a new usecase was provided

There are problems related to local environment which could break this:
```java
AppiumDriverLocalService service = AppiumDriverLocalService.buildDefaultService();
```
Also there are situations when should be used an another Node.JS instance, e.g. the instance which is installed in the directory that differs from one defined at the PATH environmental variable.

At this case user is able to set up a value (it should be a full path to desired Node.JS executable) of the __appium.node.js.exec.path__ environmental variable. Also it is possible to define it programmatically: 

```java
//appium.node.js.exec.path
System.setProperty(AppiumServiceBuilder.NODE_JS_EXECUTABLE_PROPERTY, 
"the path to the desired node.js executable");

AppiumDriverLocalService service = AppiumDriverLocalService.buildDefaultService();
```
 
- the refactoring of AppiumServiceBuilder API. It is expected that it is more robust now.  

- scripts which run __npm__ for specific purposes were moved from code to the _resources_ folder. There are no specific actions for users to do. These files will be included in *.jar and used by code of the artefact.

- pom.xml format issues were improved.

- tests of compatipility with desktop browsers now work on Chrome

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ If you are using the Eclipse IDE, make sure you are using verison Luna or later.
 - bug fix and enhancements of io.appium.java_client.service.local.AppiumDriverLocalService
     - FIXED bug which was found and reproduced with Eclipse for Mac OS X. Please read about details here: [#252](https://github.com/appium/java-client/issues/252)
     Thanks to [saikrishna321](https://github.com/saikrishna321) for the bug report
+    - FIXED bug which was found out by [Jonahss](https://github.com/Jonahss). Thanks for the reporting. Details: [#272](https://github.com/appium/java-client/issues/272)
     - The ability to set additional output streams was provided
 - The additional __startActivity()__ method was added to AndroidDriver. It allows to start activities without the stopping of a target app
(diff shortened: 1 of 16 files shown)


TikhomirovSergey:
One more not a big improvement will be commited soon... Please wait...

TikhomirovSergey:
Change list:
- the main change is the bug fix of #272. This thing was breaking the starting of a Node.js instances which differ from default (e.g. [NSolid](https://nodesource.com/products/nsolid))

- the refactoring of AppiumServiceBuilder API. It is expected that it is more robust now.  

- scripts which run __npm__ for specific purposes were moved from code to the _resources_ folder. There are no specific actions for users to do. These files will be included in *.jar and used by code of the artefact.

- pom.xml format issues were improved.

TikhomirovSergey:
@Jonahss @bootstraponline 
Please rewiev this PR :)

PR #279, opened 2015-12-05
Source code format issues fix

TikhomirovSergey:
There is no serious changes but:
- source code format issues fix
- the unnecessary "public" modifier was removed from interfaces


Jonahss:
nice

PR #284, opened 2015-12-28
Fixed Typo getAppString-> getAppStrings

rompic:
it would also be great to have a proper example how this can be harnessed in practice. I couldn't find anything on this topic.


rompic:
would it be possible to get an array of strings instead of a string? otherwise I'm not sure how to use this function properly.

rompic:
see discussion in https://discuss.appium.io/t/getappstrings/7797

Issue #288, opened 2016-01-06
zoom gestures are not working


kkhaidukov:
I'm trying to use AppiumDriver's `zoom` methods and they seem broken:
1. `zoom(WebElement el)` does nothing (i.e. the expected zoom gesture is not performed)
2. `zoom(int x, int y)` causes an exception, probably due to an incorrect usage of moveTo call

kkhaidukov:
I've created a pull request for this -- https://github.com/appium/java-client/pull/289

Jonahss:
Merged :)
Will publish in next version.

kkhaidukov:
thanks!

PR #289, opened 2016-01-06
fixed zoom gestures in AppiumDriver.java (issue #288)

kkhaidukov:


TikhomirovSergey:
@kkhaidukov 
Thank you! I'll reviev this as soon as possible.
Have you run tests? Does it work?

@Jonahss What do you think about this change?

kkhaidukov:
@TikhomirovSergey 
that's a bit tricky
I've run the Android tests in AndroidGestureTest, and they are green, but, as far as I can see, there are no asserts in elementGestureTest (which has a `zoom` call), and the element that is being zoomed doesn't seem to be really zoomable, so there we just make sure that no exception occurs.

As for the iOS tests, ZoomTest in iOSGestureTest passes too (however, there are no asserts there too). But the map does zoom in the simulator.

Besides that, tests for `zoom(int x, int y)` are nowhere to be found, but that's kind of another story.

I've checked both updated `zoom` methods with the iOS app that I'm testing against, and both work fine.

TikhomirovSergey:
@kkhaidukov 
ok. I'll review it soon. May be I'll find a way to add assertions to tests.

kkhaidukov:
@TikhomirovSergey I'd like to help with the tests as well, but that would require some changes in the TestApp, e.g. adding a view with an image that would change its size when being zoomed/pinched -- that could help with assertions

TikhomirovSergey:
@kkhaidukov 
Ok. I'll ask @Jonahss where are sources of test apps for Android/iOS.

TikhomirovSergey:
@kkhaidukov 
You can add required changes to:
https://github.com/appium/ios-test-app
https://github.com/appium/android-apidemos

If you'll add these changes please attach a new PR to this one.

kkhaidukov:
@TikhomirovSergey hmmmm, I meant that I could write the missing tests, but adding something to test apps would require me much more time, since I'm not a iOS/Android developer in any way

TikhomirovSergey:
@kkhaidukov 
Well. I'll add the missing functionality or find it. And then you'll add assertions.
Ok?

kkhaidukov:
@TikhomirovSergey sure

Jonahss:
You're going to have trouble adding assertions. We would manually watch these, because the only other option is to take screenshots and use image regression testing.

It's ok though. The tests for java-client are just to test wether the http commands generated by the client are in the right format, it's not meant to test Appium's functionality.

Tests for wether appium is behaving correctly should be in the Appium test suite.

PR #290, opened 2016-01-07
getAppStrings method returns a map instead of an string

rgonalo:
With this change getAppStrings returns a Map<String, String>, making it easier to be used:
```
Map<String, String> strings = driver.getAppStrings();
String string_value = strings.get('string_id');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,7 @@ If you are using the Eclipse IDE, make sure you are using version Luna or later.
 ##Changelog##
 *4.0.0 (still not released)*
 - `getAppStrings()` methods now return a map with the app strings keys and values, instead of an string
+- Add `getAppStrings(String language, String stringFile)` method to allow searching app strings in the specified file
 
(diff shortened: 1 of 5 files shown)


TikhomirovSergey:
:+1: 
I like it!
@Jonahss What's your opinion?

rompic:
+1

Jonahss:
this is a breaking change, so will require a major version release. I think it's good though.
Can you add a node to the changelog?

rgonalo:
I have added the note to the changelog in version 4.0.0 ¿it is ok the version number?

I have also added to the PR an additional commit with a new getAppStrings() method to search app strings in the specified file, that is already allowed by appium server: https://github.com/appium/appium/blob/cdadd16d216abbe9838ed48a37af7d0aaed67150/lib/server/controller.js#L1120

Jonahss:
Cool! Thanks.

Issue #283, opened 2015-12-24
AppiumDriverLocalService dosent spawn on MultiThreading


saikrishna321:
@TikhomirovSergey  i was trying to execute the appium android test in parallel where i wanted to start the appium-server programmatically for every thread. 

TestNG by default provides  a way were u can run methods in parallel and set the thread-count. When i trigger the appium server with AppiumLocalSerive it always hooks to a single thread. 

When i start the appium server as we do from command line it works fine. I have a sample project to reproduce it. 

https://github.com/saikrishna321/TestNGParallelThreads

When i set the ParallelMode.CLASSES  i see the the appium server and the application launched on two thread(two devices) as expected. Please check the gist for logs 

https://gist.github.com/saikrishna321/8a2e49ccc080a3faf014

When i set the ParallelMode.METHODS i dont see the methods executing in parallel, check the gist for logs and also i noticed the AfterClass for one of the has not completed, i see the process still running

https://gist.github.com/saikrishna321/6592f680dee7d70be59c

TikhomirovSergey:
@saikrishna321 Thank you for the detailed report. I'll run into this problem as soon as possible.

saikrishna321:
@TikhomirovSergey  More info: i'm running my test with TestNG framework and below the testng.xml used to trigger the tests

```
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
<suite thread-count="2" name="Suite" parallel="methods">
	<test name="Test">
		<classes>
			<class name="com.test.classes.TestNGMethodSelector">
				<methods>
					<include name="methodOne" />
					<include name="methodTwo" />
				</methods>
			</class>
			<class name="com.test.classes.DemoClassWithManyMethodsTest">
				<methods>
					<include name="methodOne_1" />
					<include name="methodTwo_2" />
				</methods>
			</class>
		</classes>
	</test> <!-- Test -->
</suite> <!-- Suite -->


```

saikrishna321:
@TikhomirovSergey  any updates in this.. i have updated the repo link .. all that i'm trying to do is just trigger appium server session as per the testng.xml

TikhomirovSergey:
@saikrishna321 
I'm working on .Net client, today or tomorrow going to finish this work.

The main idea is that it is not a bug (however, who knows :) ). AppiumDriverLocalService was designed as thread safe feature...

saikrishna321:
[...]

PR #291, opened 2016-01-13
#283 fix: Improvement of AppiumDriverLocalService for the multithreading.

TikhomirovSergey:
#283 fix Changes:
- Work with _java.util.concurrent.locks.ReentrantLock_ was incorect. It was causing deadlocks/hangings in multithreading. Now it is worked out and uses "fair" synchronization.
- Improvement of the service stopping.
- New tests were provided.

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
@@ -44,7 +44,7 @@ public final class AppiumDriverLocalService extends DriverService {
     private final String ipAddress;
     private final long startupTimeout;
     private final TimeUnit timeUnit;
-    private final ReentrantLock lock = new ReentrantLock(true);
+    private final ReentrantLock lock = new ReentrantLock(true); //uses "fair" thread ordering policy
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
@Jonahss @bootstraponline 
Please take a look at this. 
@saikrishna321 You are invited too. :) I've used your sample code with this (https://github.com/saikrishna321/TestNGParallelThreads/pull/2, it is not necessary to merge :)) little improvement for the stress testing. It was green!

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -44,7 +44,7 @@
     private final String ipAddress;
     private final long startupTimeout;
     private final TimeUnit timeUnit;
-    private final ReentrantLock lock = new ReentrantLock();
+    private final ReentrantLock lock = new ReentrantLock(true);
```
bootstraponline: Maybe add a comment here mentioning true means `fair ordering policy`? It's not obvious what `true` does in the absence of a comment.

bootstraponline:
Code looks good to me. It's great you have tests!

Issue #305, opened 2016-01-27
AppiumServiceBuilder throws InvalidServerInstanceException with Appium node 1.5.0.x


TikhomirovSergey:
AppiumServiceBuilder throws InvalidServerInstanceException because the location of the file appium.js is changed with the new major version. 

[This pattern](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java#L45) should be more fluent and compatible with higher and lower appium-node versions.

TikhomirovSergey:
#302

rompic:
@TikhomirovSergey when will this be released?

PR #306, opened 2016-02-01
#305 fix

TikhomirovSergey:
Change list:
- The #305 fix;
- update to Selenium 2.50.0
- changes of HasAppStrings interface. The old 'getAppStrings()' was returned with __Deprecated__ mark. Appium java client has wide user community. So let's care about users and let them prepare for serious API change. New methods were renamed to 'getAppStringMap'. @rgonalo Are you ok with this change?

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
@@ -36,14 +36,36 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 public final class AppiumServiceBuilder extends DriverService.Builder<AppiumDriverLocalService, AppiumServiceBuilder> {
 
+    /**
+     * The property of environmental variable used to define
+     * the path to executable appium.js (node server v<=1.4.x) or
+     * main.js (node server v>=1.5.x)
+     */
     public static final String APPIUM_PATH = "APPIUM_BINARY_PATH";
+
+    /**
+     * The property of environmental variable used to define
+     * the path to executable NodeJS file (node.exe for WIN and
+     * node for Linux/MacOS X)
+     */
     public static final String NODE_PATH = "NODE_BINARY_PATH";
 
     private static final String APPIUM_FOLDER = "appium";
+
     private static final String BIN_FOLDER = "bin";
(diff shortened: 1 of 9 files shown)


TikhomirovSergey:
@imurchie @bootstraponline 
Could you look at this PR before the merging?

rgonalo:
:+1:
But, to be a non breaking change, we should return also as deprecated the old method `public String getAppStrings(String language)`

TikhomirovSergey:
@rgonalo ok. Thanks for the reminding.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -36,14 +36,36 @@
...
     private static final String BIN_FOLDER = "bin";
+    private static final String BUILD_FOLDER = "build";
+    private static final String LIB_FOLDER = "lib";
+
     private static final String APPIUM_JS = "appium.js";
-    private static final String APPIUM_NODE_MASK =  File.separator +
-            APPIUM_FOLDER + File.separator + BIN_FOLDER + File.separator + APPIUM_JS;
+    private static final String MAIN_JS = "main.js";
+
+    private static final String ERROR_TEXT_WHEN_DEFAULT_APPIUM_NODE_IS_NOT_FOUND = "There is no installed nodes! Please " +
```
bootstraponline: `ERROR_TEXT_WHEN_DEFAULT_APPIUM_NODE_IS_NOT_FOUND` is too long. How about:

`ERROR_NODE_NOT_FOUND`

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -36,14 +36,36 @@
 
 public final class AppiumServiceBuilder extends DriverService.Builder<AppiumDriverLocalService, AppiumServiceBuilder> {
 
+    /**
+     * The property of environmental variable used to define
```
bootstraponline: `The property of` can be deleted and then it'll read more easily.

bootstraponline:
code looks good to me :+1:

TikhomirovSergey:
@rgonalo @bootstraponline 
ok. 
I'll fix it and merge it soon.

Issue #311, opened 2016-02-04
Update page object features. Server node 1.5.x


TikhomirovSergey:
These things should be done:

- mark  _name_ parameter _Decrecated_ as _By.name_ selector is not supported by native app automation

- _By.id_ selector should be used by default (native content) when not any selector strategy defined.

- _org.openqa.selenium.InvalidSelectorException_ should be handled.

TikhomirovSergey:
https://github.com/appium/java-client/issues/302

PR #313, opened 2016-02-04
removed name locator strategy from android and ios drivers along with tests

SrinivasanTarget:

### DIFF
--- a/src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
+++ b/src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
@@ -100,11 +100,17 @@ abstract class DefaultGenericMobileElement<T extends WebElement> extends RemoteW
         return super.findElementsByTagName(using);
     }
 
-    public T findElementByName(String using) {
+    /**
+     * @throws WebDriverException This method doesn't work against native app UI.
+     */
+    public T findElementByName(String using) throws WebDriverException{
         return (T) super.findElementByName(using);
     }
 
-    public List findElementsByName(String using) {
+    /**
+     * @throws WebDriverException This method doesn't work against native app UI.
+     */
+    public List findElementsByName(String using) throws WebDriverException{
         return super.findElementsByName(using);
     }
 
--- a/src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
+++ b/src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
@@ -100,17 +100,11 @@ abstract class DefaultGenericMobileElement<T extends WebElement> extends RemoteW
         return super.findElementsByTagName(using);
     }
 
-    /**
-     * @throws WebDriverException This method doesn't work against native app UI.
-     */
-    public T findElementByName(String using) throws WebDriverException{
+    public T findElementByName(String using) {
         return (T) super.findElementByName(using);
     }
 
-    /**
-     * @throws WebDriverException This method doesn't work against native app UI.
-     */
-    public List findElementsByName(String using) throws WebDriverException{
+    public List findElementsByName(String using) {
         return super.findElementsByName(using);
     }
 
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -35,6 +35,12 @@ public @interface AndroidFindBy {
 	String uiAutomator() default "";
 	String accessibility()  default "";
 	String id() default "";
+	@Deprecated
+	/**
+	 * By.name selector is not supported by Appium server node since 1.5.x.
+	 * So this option is going to be removed further. Be careful.
+	 */
+    String name() default "";
 	String className() default "";
 	String tagName() default "";
 	String xpath()  default "";
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -27,7 +27,7 @@ import java.lang.annotation.Target;
  * element or a list of elements. Used in conjunction with
  * {@link org.openqa.selenium.support.PageFactory}
  * this allows users to quickly and easily create PageObjects.
- * using Android UI selectors, accessibility, id, class name, tag and xpath
+ * using Android UI selectors, accessibility, id, name, class name, tag and xpath
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.TYPE})
@@ -40,7 +40,7 @@ public @interface AndroidFindBy {
 	 * By.name selector is not supported by Appium server node since 1.5.x.
 	 * So this option is going to be removed further. Be careful.
 	 */
-    String name() default "";
+        String name() default "";
 	String className() default "";
 	String tagName() default "";
 	String xpath()  default "";
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -39,8 +39,7 @@ public @interface AndroidFindBy {
 	/**
 	 * By.name selector is not supported by Appium server node since 1.5.x.
 	 * So this option is going to be removed further. Be careful.
-	 */
-        String name() default "";
+	 */String name() default "";
 	String className() default "";
 	String tagName() default "";
 	String xpath()  default "";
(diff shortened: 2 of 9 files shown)


src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
```diff
@@ -100,11 +100,17 @@ public List findElementsByTagName(String using) {
         return super.findElementsByTagName(using);
     }
 
-    public T findElementByName(String using) {
```
TikhomirovSergey: Should be reverted

src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
```diff
@@ -100,11 +100,17 @@ public List findElementsByTagName(String using) {
...
+    /**
+     * @throws WebDriverException This method doesn't work against native app UI.
+     */
+    public T findElementByName(String using) throws WebDriverException{
```
TikhomirovSergey: Should be reverted

src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
```diff
@@ -100,11 +100,17 @@ public List findElementsByTagName(String using) {
...
         return (T) super.findElementByName(using);
     }
 
-    public List findElementsByName(String using) {
+    /**
+     * @throws WebDriverException This method doesn't work against native app UI.
+     */
+    public List findElementsByName(String using) throws WebDriverException{
         return super.findElementsByName(using);
     }
```
TikhomirovSergey: Should be reverted

src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
```diff
@@ -27,7 +27,7 @@
  * element or a list of elements. Used in conjunction with
```
TikhomirovSergey: Should be reverted

src/main/java/io/appium/java_client/pagefactory/SelendroidFindBy.java
```diff
@@ -27,13 +27,12 @@
  * element or a list of elements. Used in conjunction with
```
TikhomirovSergey: Let's keep it for some time

src/test/java/io/appium/java_client/pagefactory_tests/SelendroidModeTest.java
```diff
@@ -50,15 +50,6 @@
     @SelendroidFindBy(id = "text1")
     private WebElement textSelendroidId;
 
-    @SelendroidFindBy(name = "Accessibility")
```
TikhomirovSergey: If Selendroid mode supports by name  then this change should be reverted

src/test/java/io/appium/java_client/pagefactory_tests/iOSPageObjectTest.java
```diff
@@ -118,13 +118,13 @@
 	
 	@iOSFindAll({
```
TikhomirovSergey: Is this test passing?

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -124,7 +124,7 @@ public void hideKeyboardWithParametersTest() {
```
TikhomirovSergey: @bootstraponline @imurchie 
By.name is deprecated? iOS automation

src/test/java/io/appium/java_client/ios/iOSGestureTest.java
```diff
@@ -134,7 +134,7 @@ public void TapSingleFingerTest() {
 
   @Test
```
TikhomirovSergey: @bootstraponline @imurchie 
By.name is deprecated? iOS automation

src/main/java/io/appium/java_client/pagefactory/iOSFindBy.java
```diff
@@ -27,7 +27,7 @@
  * element or a list of elements. Used in conjunction with
```
TikhomirovSergey: @bootstraponline @imurchie 
By.name is deprecated? iOS automation

This change should be reverted anyway

src/test/java/io/appium/java_client/ios/iOSGestureTest.java
SrinivasanTarget: @TikhomirovSergey https://github.com/appium/appium-ios-driver/blob/master/lib/driver.js#L75

This is removed from ios driver as well.Please look into this.Correct me if i am wrong.

TikhomirovSergey:
Hi @SrinivasanTarget 
First of all let's add descriptions of PR. Ok?

I'm against these changes as they are because: 
- it is breaking and harmfull for users.

- I'm not sure that tests will pass;
Take a look at these
https://github.com/appium/java-client/pull/313/files#diff-54b126e009942041858b50db621bc3aeR121
https://github.com/appium/java-client/pull/313/files#diff-54b126e009942041858b50db621bc3aeR127

- Also I'm not sure that _By.name_ is deprecated for Selendroid. Also I couldn't check check it because of some current problems with 1.5.x. @bootstraponline @imurchie Is _By.name_ not supported by Selendroid mode. It can be used instead of _accessibility_?

@SrinivasanTarget 
But you can make some improvements and help ne to resolve #311 

What should be done if you want this PR to be merged: 
- revert changes which require reversion. Read my comments. 

- The _name_ option of annotatations should be annotated _Deprecated_. You can do it this way 

```java
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.FIELD, ElementType.TYPE})
public @interface AndroidFindBy {
	String uiAutomator() default "";
	String accessibility()  default "";
	String id() default "";
	@Deprecated
	/**
	 * By.name selector is not supported by Appium server node since 1.5.x.
	 * So this option is going to be removed further. Be careful.
	 */
	String name() default "";
	String className() default "";
	String tagName() default "";
	String xpath()  default "";
}
```
So users will be warned and they will have time to improve their projects. We will remove it later.  But I still don't know which _name_ options should be deprecated. @bootstraponline @imurchie is _By.name_ deprecated for Selendroid? One more question. Why _By.name_ is deprecated for iOS? iOS-elements have the _name_ tag, I think.

So the point 
> mark name parameter Decrecated as By.name selector is not supported by native app automation

will be finished.

- You could finish this point  

>org.openqa.selenium.InvalidSelectorException should be handled.

The target code is here: 
https://github.com/appium/java-client/blob/6f7615af626e201bb6431fb79945488c3c0a5743/src/main/java/io/appium/java_client/pagefactory/ThrowableUtil.java#L26. The exception 

 ```java
org.openqa.selenium.InvalidSelectorException: Locator Strategy 'css selector' is not supported for this session (WARNING: The server did not provide any stacktrace information)
```

is thrown and it is not being handled here. We should keep it backward compatible with Appium node 1.4x for some time. So you could improve this method the following way: 
 
    1) firstly it checks the excepion type. If it is _org.openqa.selenium.InvalidSelectorException_ then method should return true. 

   2) if 1) has a negative result then it checks exception message. It could use regexps. If message is convenient to the previous pattern 
https://github.com/appium/java-client/blob/6f7615af626e201bb6431fb79945488c3c0a5743/src/main/java/io/appium/java_client/pagefactory/ThrowableUtil.java#L24

or   
 
 ```java
Locator Strategy 'xxxr' is not supported bla-bla-bla
```

then then method should return true. 

 3) if both 1) and 2) have negative results then these steps are repeated recursively with the cause of exception.

src/main/java/io/appium/java_client/DefaultGenericMobileElement.java
TikhomirovSergey: ...By.nane is still supported by browser automation tools

SrinivasanTarget:
Thanks @TikhomirovSergey for your time.Will update the descriptions going forward. And will update the PR based on your review comments.

bootstraponline:
> @bootstraponline @imurchie is By.name deprecated for Selendroid?

[No.](https://github.com/selendroid/selendroid/blob/c2ec0d49f18279e751ce0500f2454e7c494657f0/selendroid-server/src/main/java/io/selendroid/server/model/internal/NativeAndroidBySelector.java#L38) Even if it was deprecated (it isn't), we'd still want to wait for support to be removed from the server. If the server supports it then I don't see why clients should remove the feature.

SrinivasanTarget:
Thanks for clarification. Will update my piece of code.

TikhomirovSergey:
@SrinivasanTarget 
Ok. I've seen your changes. They are looking much better now :). I'm going to review and merge them soom.

SrinivasanTarget:
I'm still working on this. will update once it's done.

TikhomirovSergey:
ok

SrinivasanTarget:
@TikhomirovSergey I have handled both the scenario's mentioned above.Please review.

TikhomirovSergey:
@SrinivasanTarget I'm ok. :+1: 
@imurchie?

imurchie:
Conceptually it looks fine. I can't comment on the Java code.

Issue #340, opened 2016-03-15
Widget class cannot be in Groovy


aznime:
## Description

When creating a widget, code must be in Java. An illegal access exception is thrown if it is coded in Groovy.

## Environment

* java client build version or git revision if you use some shapshot: **3.3.0**
* Appium server version or git revision you is use some shapshot: **1.14.16**
* Desktop OS/version used to run Appium if necessary: **OSX El Capitan (10.11.1)**
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: **0.12.7**
* Mobile platform/version under test: **iOS 9.2**
* Real device or emulator/simulator: **Simulator**

## Exception stacktraces

https://gist.github.com/aznime/54263754d2a312ffb1bc

TikhomirovSergey:
@aznime 
Could you share a gist with code of this class?
_com.jumbo.shared.widgets.PlayWidget_

It looks like a bug but I don't know how to reproduce it. Java client widget tests use objects with protected constructors and everything is ok.

TikhomirovSergey:
@aznime 
Also.
How should I check it? Have you a Groovy project?

aznime:
Sorry for the late reply. I was trying to create a demo project for this purpose.

You can find it here: https://github.com/aznime/jumbo-mobile-tests

TikhomirovSergey:
@aznime 
Thank you. I've started to search for the root cause of the error.

TikhomirovSergey:
@aznime 
Ok. It seems I know the root cause. I think that cglib and reflection are not used properly at some cases and I'm going to fix it. 

As a workaround I can advice you to change class modifiers to public. 

```java
public class UICatalogWidget extends Widget {
...
}
```

Can you do it and send a message about the result.

aznime:
@TikhomirovSergey 
I got the same exception after explicitly using the public modifier.

TikhomirovSergey:
@aznime 
It needs for some redesign. It is going to be included at 3.4.1

TikhomirovSergey:
@aznime 
The fix is going to be included at 3.4.1 very soon!

TikhomirovSergey:
The fix has been merged and it is going to be published

PR #344, opened 2016-03-24
#340 fix

TikhomirovSergey:
## Change list

- #340 fix
-  dependency update:
   - all dependencies were updated
   - the dependency on org.apache.commons commons-lang3 v3.4 was added
   - the fix of Widget method invocation.#340. A class visibility was taken into account.
 
## Types of changes

##Details

- #340 fix. Incorrect work with  _java.lang.reflect_ (method visibility) was improved.
- all dependencies were updated
- the dependency on org.apache.commons commons-lang3 v3.4 was added

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -25,12 +25,12 @@
         <dependency>
             <groupId>com.google.code.gson</groupId>
             <artifactId>gson</artifactId>
-            <version>2.2.4</version>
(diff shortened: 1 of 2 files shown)


TikhomirovSergey:
@aznime
Can you take a look at these changes?

Issue #342, opened 2016-03-21
scrollTo does NOT work on named element


autoaim800:
## Description
Method `public MobileElement scrollTo(String text)` from https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/ios/IOSElement.java does NOT work on named element (once the element has been assigned with accessibilityIdentifier).

THIS IS A RE-OPEN OF https://github.com/appium/java-client/issues/250, previously closed by @TikhomirovSergey 
THIS ISSUE EXISTS ON iOS ONLY.

### Preliminary analysis:
Attribute of accessibilityIdentifier will assign static value to the 'name', e.g. my-cell-title, hence 'name' can not be used as identifier of changing text, see current implementation below
```return (IOSElement) findElementByIosUIAutomation(".scrollToElementWithPredicate(\"name CONTAINS '" + text + "'\")"); ```

### Likely solution
```return (IOSElement) findElementByIosUIAutomation(".scrollToElementWithPredicate(\"value CONTAINS '" + text + "'\")");```

## Environment
* Issue exists on java client till now - git revision ad9f757  on Dec 6, 2015

## Details
If an element is assigned with accessibilityIdentify=myCellTitle, then the name is always equal to 'myCellTitle', but its 'value' is always changing according to its assignment, hence 'value' should be used as a query parameter instead of 'name'.

## Code To Reproduce Issue [ Good To Have ]
This issue is found during code-static-review, not applicable with code-to-reproduce.

## Exception stacktraces
This issue is found during code-static-review, not applicable with exception.

## Link to Appium logs
This issue is found during code-static-review, not applicable with appium log.

TikhomirovSergey:
@autoaim800 
Now I'm seeing the great description and I'll start the investigation

SrinivasanTarget:
appium/appium#2942

SrinivasanTarget:
appium/appium#4633

SrinivasanTarget:
Android Gestures API: http://developer.android.com/training/gestures/scroll.html

TikhomirovSergey:
I'm closing it according to #385

Issue #341, opened 2016-03-17
ScrollToExact inconsistent in it's behavior


nuggit32:
## Description

I'm not sure this is the expected behavior of ScrollToExact, but I would expect it to stop at the text once it is present on the screen.

I've seen ScrollToExact work correctly sometimes and incorrectly sometimes.

Incorrectly:
1. Scrolls the whole page to the bottom (passes by the text)
2. Then all the way back to the top (passes by the text)
3. Then back to the middle where it needed to stop. (now it stop on the text)

## Environment

* java client build version or git revision if you use some shapshot: 3.4.0
* Appium server version or git revision you is use some shapshot: 1.4.16 and Source 1.5
* Desktop OS/version used to run Appium if necessary:  MAC OSX El Capitan 10.11.3
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: 5.5.0
* Mobile platform/version under test: Android
* Real device or emulator/simulator: API 21 Android 5.1.1

## Details

I would expect the scrolling to stop mid way down where the text is present, but I see it scan the whole page, then back to the top then back down to the text field. It passes by the text field 2 times.

## Code To Reproduce Issue [ Good To Have ]

   ((AndroidDriver<?>) driver).scrollToExact("7-31 Day CD");

## Exception stacktraces

There is no stacktrace to provide as eventually it finds the text.

## Link to Appium logs

https://gist.github.com/nuggit32/b959e7dd241ae3546614

TikhomirovSergey:
I'm closing it according to #385

Issue #226, opened 2015-08-07
UiScrollable Within A Context - Scrolling First Element


anders-g-hansen:
Sending a UiScrollable command to a page that contains multiple scrollable elements always appears to act upon the first scrollable element even when a context is provided.

Here is an example using test I plugged into the java-client AndroidUIAutomatorTest class to demonstrate the issue.  Note, the test navigates to a page in the test app which has multiple scrollables.  When it attempts to scroll to find an element it scrolls the wrong scrollable (first on the page) rather than the element in the provided context.

	  @Test
	  public void scrollBars() {
		driver.scrollToExact("Views").click();
		driver.scrollToExact("ScrollBars").click();
		driver.scrollToExact("3. Style").click();

		// find the middle ScrollView (second one since we're 0 based index)
		AndroidElement element = driver.findElementByAndroidUIAutomator("new UiSelector().className(android.widget.ScrollView).index(1)");
		assertTrue(element.isDisplayed());

		AndroidElement found;

		// use a UiScrollable selector with something we know will be found within the context of the middle scrollview and print the text to ensure it works
		found = element.findElementByAndroidUIAutomator("new UiScrollable(new UiSelector().scrollable(true)).scrollIntoView(className(android.widget.TextView))");
		assertTrue(found.isDisplayed());
		System.out.println("text: " + found.getText());

		// use a UiScrollable selector with something ew know will NOT be found to see if it will properly scroll to find it, note it is scrolling the first Scrollable on the page
		// not the one with the "element" context
		found = element.findElementByAndroidUIAutomator("new UiScrollable(new UiSelector().scrollable(true)).scrollIntoView(resourceId(\"notFoundId\"))");
		assertTrue(found.isDisplayed());
		System.out.println("text: " + found.getText());

	  }

Here are the appium logs for the above test run.

The first log provided is from the first find within the context (4 from the log).  This call works as expected.

		info: --> POST /wd/hub/session/84c164f6-fc7c-433f-ac46-7cae26739c05/element/4/element {"id":"4","using":"-android uiautomator","value":"new UiScrollable(new UiSelector().scrollable(true)).scrollIntoView(className(android.widget.TextView))"}
		info: [debug] Waiting up to 0ms for condition
		info: [debug] Pushing command to appium work queue: ["find",{"strategy":"-android uiautomator","selector":"new UiScrollable(new UiSelector().scrollable(true)).scrollIntoView(className(android.widget.TextView))","context":"4","multiple":false}]
[...]

Issue #197, opened 2015-05-28
iOS - scrollTo and scrollToExact should parameterise the class name


email2vimalraj:
Currently, the `scrollTo` and `scrollToExact` method has the `UIATableView` hardcoded in the method under the `IOSDriver` class. The `UIATableView` is specific to a particular type of application.

The application which I test has implemented the `UIAScrollView` instead of `UIATableView`. Hence I would recommend to parameterise the value of the view class name like as follows:

```java
public MobileElement scrollTo(String className, String text) {
     return ((IOSElement) findElementByClassName(className)).scrollTo(text);
}
```

Same for `scrollToExact`. Please refer [UIAScrollView Documentation](https://developer.apple.com/library/prerelease/ios/documentation/ToolsLanguages/Reference/UIAScrollViewClassReference/index.html).

Jonahss:
:+1:

TikhomirovSergey:
I'm closing it according to #385

PR #366, opened 2016-04-24
#342 Fix

SrinivasanTarget:
## Change list

Added ScrollToUsingPredicates method for android and iOS
 
## Types of changes

##Details

ScrollTo  -  Existing method doesn't work for UIScrollableView.It works fine only for UITableView since it is hardcoded. This will again be re-fractored in upcoming commits in same PR.

Now introduced a new method "ScrollToUsingPredicates" for both android and iOS that enables user to define their own predicates that works fine for any scrollable views.

For Predicates in iOS refer: 
1)https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/ios_predicate.md
2)https://developer.apple.com/library/mac/documentation/Cocoa/Reference/Foundation/Classes/NSPredicate_Class/

For Android Scroll refer:
1)http://developer.android.com/reference/android/support/test/uiautomator/UiScrollable.html#scrollTextIntoView(java.lang.String)

@TikhomirovSergey Your thoughts on this?

As we discussed,this PR will continue to fix existing ScrollTo method.Will work on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,8 @@ Locators:
 - findElementsByIosUIAutomation()
 - findElementByAndroidUIAutomator()
 - findElementsByAndroidUIAutomator()
+- findElementByIosNsPredicate()
+- findElementsByIosNsPredicate()
 
 ## Note to developers! ##
 If you are working on this project and use Intellij Idea, you need to change the compiler to the Eclipse compilers instead of the default.
@@ -128,6 +130,7 @@ a subclass of `io.appium.java_client.internal.JsonToMobileElementConverter`. Con
 - the integration with `org.owasp dependency-check-maven` was added. Thanks to [@saikrishna321](https://github.com/saikrishna321) 
 for the work.
 - the integration with `org.jacoco jacoco-maven-plugin` was added. Thanks to [@SrinivasanTarget](https://github.com/SrinivasanTarget) for the contribution.
+- the searching for elements by nspredicate string was added. It requires the [XCUITest mode](https://github.com/appium/java-client/blob/master/docs/Installing-xcuitest-driver.md). Thanks to [@rafael-chavez](https://github.com/appium/java-client/pull/352) for the contribution
 
 ...
 
--- a/docs/Installing-xcuitest-driver.md
+++ b/docs/Installing-xcuitest-driver.md
@@ -1,0 +1,19 @@
+## Setting Up XCUITest
+---
+
+**Install Appium XCUITest Driver**
+
+Clone Appium XCUITest Driver repo https://github.com/appium/appium-xcuitest-driver
+
+```
+git clone --recursive https://github.com/appium/appium-xcuitest-driver.git
+cd appium-xcuitest-driver
+npm install
+```
+
+Set your appium path to the appropriate location (Where you cloned appium-xcuitest-driver) and you can 
+start the driver by doing:
+
+```
+node .
+```
--- a/pom.xml
+++ b/pom.xml
@@ -15,7 +15,8 @@
   ~ limitations under the License.
   -->
 
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
     <groupId>io.appium</groupId>
@@ -62,6 +63,16 @@
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-lang3</artifactId>
             <version>3.4</version>
+        </dependency>
+        <dependency>
+            <groupId>net.lingala.zip4j</groupId>
+            <artifactId>zip4j</artifactId>
+            <version>1.3.2</version>
+        </dependency>
+        <dependency>
+            <groupId>commons-io</groupId>
+            <artifactId>commons-io</artifactId>
+            <version>2.1</version>
         </dependency>
     </dependencies>
     <packaging>jar</packaging>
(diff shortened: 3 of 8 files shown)


SrinivasanTarget:
https://github.com/appium/java-client/issues/342

TikhomirovSergey:
@SrinivasanTarget 
I'm against the proposed PR is it is now.  It is much easier to make scrolling methods _deprecated_ and remove it later and make users to use **findBySomeAutomation** instead. May be it is more correct solution but it is not convenient I think. 

I'm looking forward for more complex solution:

- 1. There is no need to modify  io.appium.java_client.ScrollsTo. Let it stay as it is.

- 2. MobileBy needs for some refacoring I think. It would be cool if it looked like the sample below after the refactoring:

```java
...
public abstract class MobileBy extends By implements Serializable {
    private final String automatorText;

    public MobileBy(String automatorText) {
        this.automatorText = automatorText;
    }

    protected String getAutomatorText() {
        return automatorText;
    }

    ...
    
    public static class ByIosUIAutomation extends MobileBy  {
        public ByIosUIAutomation(String uiautomationText) {
            super(uiautomationText);
        }

        @SuppressWarnings("unchecked")
        @Override
        public List<WebElement> findElements(SearchContext context) {
            return (List<WebElement>) ((FindsByIosUIAutomation<?>) context)
                    .findElementsByIosUIAutomation(getAutomatorText());
        }

        @Override public WebElement findElement(SearchContext context) {
            return ((FindsByIosUIAutomation<?>) context)
                .findElementByIosUIAutomation(getAutomatorText());
        } 

        @Override public String toString() {
            return "By.IosUIAutomation: " + getAutomatorText();
        } 
    }
  
    public static class ByAndroidUIAutomator extends By implements Serializable {
        public ByAndroidUIAutomator(String uiautomatorText) {
            super(uiautomationText);
        }

        @SuppressWarnings("unchecked")
        @Override
        public List<WebElement> findElements(SearchContext context) {
            return (List<WebElement>) ((FindsByAndroidUIAutomator<?>) context)
                    .findElementsByAndroidUIAutomator(getAutomatorText());
        }

        @Override public WebElement findElement(SearchContext context) {
            return ((FindsByAndroidUIAutomator<?>) context)
                .findElementByAndroidUIAutomator(getAutomatorText());
        }

        @Override public String toString() {
            return "By.AndroidUIAutomator: " + automatorText;
        }

        @Override public String getAutomatorText() {
            return automatorText;
        }
    }

    public static class ByAccessibilityId extends By implements Serializable {

        private final String id;

        public ByAccessibilityId(String id) {
            this.id = id;
        }

        @SuppressWarnings("unchecked")
        @Override
        public List<WebElement> findElements(SearchContext context) {
            return (List<WebElement>) ((FindsByAccessibilityId<?>) context)
                .findElementsByAccessibilityId(id);
        }

        @Override public WebElement findElement(SearchContext context) {
            return ((FindsByAccessibilityId<?>) context).findElementByAccessibilityId(id);
        }

        @Override public String toString() {
            return "By.AccessibilityId: " + id;
        }
    }
}
```

TikhomirovSergey:
@SrinivasanTarget what do you think about this solution?

@bootstraponline @Jonahss you can join to this discussion if you want.
@autoaim800 @nuggit32 you are invited too.

#341

SrinivasanTarget:
@TikhomirovSergey Above suggested Android changes looks fine for me,

Have few queries w.r.t iOS,

1) Why are we not gonna refractor existing ScrollTo method? It works fine only for "UITableView" but not for other scrollable views e.g "UIScrollView"
2) I see above suggested iOS changes also uses "UITableView" hard coded, this serves the same way as existing ScrollTo i.e may work only for UITableView.

```
public T scrollToUsingPredicate(By scrollableElementBy, String iOSpredicate) {
        return (T) findElementBy(scrollableElementBy)
            .scrollToUsingPredicate(iOSpredicate);
   }

   public T scrollToUsingPredicate(String iOSpredicate) {
        return scrollToUsingPredicate(By.className("UIATableView"), iOSpredicate);
   }
```
I think we need to use ByIosUIAutomation instead of By. Still not clear with this portion.Please clarify.

autoaim800:
Hi 
Thanks for inviting me.

I'm just thinking aloud:

#341 supposed to be fixed by #342, and #342 is supposed to be fixed by #366

For #341 the issue is scrollTo() and/or scrollToExact() can not be stopped at first occurrence at first glance, what's the root cause?
I thought it's not fix-able, as on Android, ListView has to scroll top->bottom to build a complete list of cells - as cells are not loaded at first/once (not like iOS), that's why only on its way back bottom->top, it stops at the first occurrence; correct me if I'm mistaken - I only have observation, no evidence at all to back up myself.

For #342, the solution seems reasonable from my perspective, either stick/correct the old-fashion implementation or introduce new method/feature works fine for me.

For #366, it's more like a new feature - can I say I like both two new methods to be added.

-Bill

TikhomirovSergey:
https://github.com/appium/java-client/issues/197

SrinivasanTarget:
https://github.com/appium/java-client/issues/226

SrinivasanTarget:
@autoaim800 Thanks for your comments. Your observations are correct and the above tagged/discussed issues will be addressed.

@TikhomirovSergey Have updated another possible iOS solution that works for all scrollable views now.Please have a look.Waiting for your comments:)

SrinivasanTarget:
@TikhomirovSergey  

**Second Solution:**
Have altered your solution here below, please check which one would be good to have,

*MobileBy will be re-fractored as you said above
*AndroidDriver will have below changes,
```
    static String uiScrollablenew(String uiSelector) {
        return "new UiScrollable(new UiSelector().scrollable(true).instance(0))"
            + ".scrollTextIntoView(\"" + uiSelector + "\");";
    }

    public T scrollToUsingAutomator(ByAndroidUIAutomator by) {
        return findElementByAndroidUIAutomator(uiScrollablenew(by.getAutomatorText()));
    }
```
*IOSDriver will have below changes,
```
    static String scrollToVisible(ByIosUIAutomation scrollableElementBy) {
        return  scrollableElementBy.getAutomatorText() + ".scrollToVisible();";
    }

    public T scrollToUsingAutomator(ByIosUIAutomation scrollableElementBy) {
        return (T) findElementByIosUIAutomation(scrollToVisible(scrollableElementBy));
    }
```

Your thoughts on solution comitted and the second one?

SrinivasanTarget:
https://github.com/appium/appium/issues/2942

SrinivasanTarget:
https://github.com/appium/appium/issues/4633

SrinivasanTarget:
appium/appium#4633

TikhomirovSergey:
I'm closing it according to #385

PR #375, opened 2016-04-28
Fixed doc issues in jdk 8

SrinivasanTarget:
## Change list

To skip doc issues on java 8 compiler
 
## Types of changes

## Details

Added parameter to skip compilation issues related to doc on java 8 compiler.Also this is downward compatible.Works on both 7 & 8 compilers.

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -33,8 +33,8 @@ import java.lang.annotation.Target;
 public @interface AndroidFindBy {
     /**
      * It is an is Android UIAutomator string.
-     * Read {@link http://developer.android.com/intl/ru/tools/testing-support-library/
-     * index.html#uia-apis}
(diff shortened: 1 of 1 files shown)


SrinivasanTarget:

saikrishna321:
@SrinivasanTarget i do this locally to build. With this lint added, do you think we are stopping java doc creation for JDK 7 as well ?

SrinivasanTarget:
We can solve this by creating a profile for java 8 and ignore doc but that's not the right way. So fixed this now.

Able to generate doc's on both jdk 7  & 8 now.

java 8 : https://gist.github.com/SrinivasanTarget/96454da9b9af507ca340a6797c1b03c5

java 7 : https://gist.github.com/SrinivasanTarget/c26fef3f8f017bb4d8a43a63564868c5

TikhomirovSergey:
@SrinivasanTarget 
I think it was more correct to remove tags which are non supported by java 8. Please take a look at 
https://github.com/appium/java-client/commit/d8cd86d2149d861a6a7d423e9c0460518523163b

SrinivasanTarget:
Cool Fixed in d8cd86d

Thanks @TikhomirovSergey

PR #384, opened 2016-05-07
Removed deprecated method's

SrinivasanTarget:
## Change list

Removed deprecated method's:
1)getMouse()
 
## Types of changes

@TikhomirovSergey ping


TikhomirovSergey:
I'wont merge this PR. This method was added with the purpose to warn users. This method works fine against desktop browsers and it doesn't work on mobile OSs. If I will remove it then this warning will disappear but the method will be present because it is declared by RemoteWebDriver.

PR #386, opened 2016-05-09
scrollToChildElement and scrollWithDirection

rafael-chavez:
## Change list

- IOSElement currently implements scrollToExact and and scrollTo, both of which use UIAutomation framework. Added scrollToChildElement and scrollWithDirection (names can change if we dont like those), using the new XCUITest Framework (WebDriverAgent).  
 
## Types of changes

## Details

ScrollWithDirection usage:
IOSElement element;
... //set element somewhere
element.scrollWithDirection(ScrollDirection.Down)

scrollToChildElement usage:
IOSElement element;
... //set element somewhere
String childElementIdentifier = getChildIdentifierToScrollTo();
element.scrollToChildElement(childElementIdentifier);


rafael-chavez:
@TikhomirovSergey

SrinivasanTarget:
@rafael-chavez Can you add some tests please? That would give us a clear picture about the feature.

src/main/java/io/appium/java_client/IOSMobileCommand.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
TikhomirovSergey: I don't like this class. I think that it is better to just add the new command to MobileCommand.

src/main/java/io/appium/java_client/MobileCommand.java
```diff
@@ -72,10 +72,7 @@ static CommandInfo postC(String url) {
         return new CommandInfo(url, HttpMethod.POST);
     }
 
-    private static Map<String, CommandInfo> getMobileCommands() {
-        if (commandRepository != null) {
-            return commandRepository;
-        }
+    protected static ImmutableMap.Builder<String, CommandInfo> getCommandsBuilder() {
```
TikhomirovSergey: What the reason to change the modifier to _protected_

TikhomirovSergey:
Hi @rafael-chavez 

There are some remarks 

1.
>IOSElement currently implements scrollToExact and and scrollTo, both of which use UIAutomation framework.

These methods are deprecated and they are going to be removed. The situation is not clear. It is possible that some scroll commands would be implemented on the server side using iOS UI Automation framework. If it will be real then they will require to add specific commands too. So then there will be method duplication.

2. Yes. Tests should be present there.

3. I have a question. How many changes (related to XCUITest) are going to be proposed. May be there is sense to splip IOSDriver to IOSAutomationDriver and IOSXCUITDriver. @SrinivasanTarget @rafael-chavez What do you think?

src/main/java/io/appium/java_client/IOSMobileCommand.java
rafael-chavez: Sounds good. I just didnt want to add uiaElement/:elementId/scroll to MobileCommands since this seemed more like a IOS command. I have no problem moving this to MobileCommand, actually make this easier.

src/main/java/io/appium/java_client/MobileCommand.java
rafael-chavez: Ah that was because i was overriding/hiding this method in the IOSMobileCommand class. I will remove when i put this command in the MobileCommand class.

rafael-chavez:
@TikhomirovSergey here are some of the pros and cons that i can think of to splitting up the code base to different branches.

Pros:
1. Code base will be cleaner as only code pertaining to each driver will live on that branch.
2. Less chance of conflicts as we will only have to worry about the specific driver in each branch and not worry about breaking the other driver when changes are introduced.

Cons:
1. The branches might diverge so much that when its time to merge them there will be a ton of conflicts (if we add features to UIAutomation branch that are not in XCUITesting branch) .
2. More maintenance as there will be to branches to try to keep feature parity. (Unless we decide to not add any more features to UIAutomation driver branch).

If we dont add any more features to the UIAutomation branch then i think it might make sense to split this up since it will be cleaner to work with. Im not sure if there is a plan to stop adding to UIAutomation @TikhomirovSergey @SrinivasanTarget

SrinivasanTarget:
@rafael-chavez We are not talking about splitting the code base to different branches in our repo.Its just having the driver's specific to UIAutomation/XCUITest and all these implementations will continue in same branch.Correct me if i'm wrong @TikhomirovSergey ? Also @TikhomirovSergey I like the change you have suggested above reg. splitting the drivers.

Also we don't have any plans of stop supporting/adding new features/functionality to UIAutomation? I don't think so XCUITest is completely matured enough to handle things standalone now.

TikhomirovSergey:
>Correct me if i'm wrong @TikhomirovSergey?

@SrinivasanTarget yes, you are right.

@rafael-chavez It is very risky to merge changes which may confuse other user. But I have an idey which can be acceptable to you.

What if this constructor would be public:
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/AppiumDriver.java#L101
?

Than you can do these things:

```java
Map<String, CommandInfo> yourMap = new HashMap<>();
yourMap.putAll(MobileCommand.commandRepository);
yourMap.put("scroll", MobileCommand.postC("/session/:sessionId/uiaElement/:id/scroll"));

AppiumCommandExecutor executor = new AppiumCommandExecutor(yourMap, new URL("your URL"));
driver = new IOSDriver<>(executor, capabilities);

...

public class YourXCUITUtil {
    public static void scroll(AppiumDriver driver, otherArgs) {
         ImmutableMap.Builder builder = ImmutableMap.builder();
         ...
         execute("scroll", builder.build());
    }
}
```

These methods should be public too:

https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/MobileCommand.java#L67
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/MobileCommand.java#L71

@SrinivasanTarget @rafael-chavez What do you think about this improvement?

I would like to keep this PR opened. I'm thinking about the driver splitting... But the consistency of XCUIT mode is needed on the Appium server-side.

rafael-chavez:
I like this proposed change @TikhomirovSergey. This would increase our flexibility without having too many constructors and like you said confuse less people. If we agree upon this change i can make the appropriate changes and open a pull request.

SrinivasanTarget:
I like your changes :+1

TikhomirovSergey:
>if we agree upon this change i can make the appropriate changes and open a pull request.

@rafael-chavez 
Sure. We can wait for some time. Changes proposed by the new one PR will be included in v4.0.0.

TikhomirovSergey:
#398

PR #389, opened 2016-05-14
Android - Network APIs Data Wifi and Airplane mode Implementation

SrinivasanTarget:
## Change list

Network API's to toggle Data,Wifi and AirplaneModes are implemented for AndroidDriver.
 
## Types of changes

## Details

Implemented Android API's to toggle Data,Wifi and Airplane Mode's are implemented in this PR.
**Results:**https://gist.github.com/SrinivasanTarget/a2310159ebf491bef9fcc7a21a2a4dcf

@TikhomirovSergey Please review

### DIFF
--- a/src/main/java/io/appium/java_client/android/AndroidDriver.java
+++ b/src/main/java/io/appium/java_client/android/AndroidDriver.java
@@ -28,7 +28,10 @@ import static io.appium.java_client.MobileCommand.PRESS_KEY_CODE;
 import static io.appium.java_client.MobileCommand.PUSH_FILE;
 import static io.appium.java_client.MobileCommand.SET_NETWORK_CONNECTION;
 import static io.appium.java_client.MobileCommand.START_ACTIVITY;
+import static io.appium.java_client.MobileCommand.TOGGLE_AIRPLANE_MODE;
+import static io.appium.java_client.MobileCommand.TOGGLE_DATA;
 import static io.appium.java_client.MobileCommand.TOGGLE_LOCATION_SERVICES;
+import static io.appium.java_client.MobileCommand.TOGGLE_WIFI;
 import static io.appium.java_client.MobileCommand.UNLOCK;
 
 import com.google.common.collect.ImmutableMap;
@@ -403,6 +406,18 @@ public class AndroidDriver<T extends WebElement>
 
     public void toggleLocationServices() {
         execute(TOGGLE_LOCATION_SERVICES);
+    }
+
+    public void toggleFlightMode() {
+        execute(TOGGLE_AIRPLANE_MODE);
+    }
+
+    public void toggleData() {
+        execute(TOGGLE_DATA);
+    }
+
+    public void toggleWiFi() {
+        execute(TOGGLE_WIFI);
     }
 
     /**
--- a/src/test/java/io/appium/java_client/android/AndroidDriverTest.java
+++ b/src/test/java/io/appium/java_client/android/AndroidDriverTest.java
@@ -75,18 +75,6 @@ public class AndroidDriverTest extends BaseAndroidTest {
         driver.toggleLocationServices();
     }
 
-    @Test public void toggleFlightModeTest() {
-        driver.toggleFlightMode();
-    }
-
-    @Test public void toggleDataTest() {
-        driver.toggleData();
-    }
-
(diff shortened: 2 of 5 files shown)


src/main/java/io/appium/java_client/android/AndroidDriver.java
```diff
@@ -405,6 +408,18 @@ public void toggleLocationServices() {
         execute(TOGGLE_LOCATION_SERVICES);
     }
 
+    public void toggleFlightMode() {
```
TikhomirovSergey: I think it would be cool if new methods were defined at the HasNetworkConnection interface

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
```diff
@@ -75,6 +75,18 @@
         driver.toggleLocationServices();
     }
 
+    @Test public void toggleFlightModeTest() {
```
TikhomirovSergey: Why did not you add them to AndroidConnectionTest? :)

TikhomirovSergey:
@SrinivasanTarget 
There are some remarks. But it looks ok.
There is #390. I've changed engines which use the connection setting/getting via bitmask. It would be cool to know the difference betwee proposed API and current. 

@bootstraponline Can you clarify? Is something deprecated?

bootstraponline:
> @bootstraponline Can you clarify? Is something deprecated?

Here's [the spec for network type](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#connectiontype). Appium is trying to implement the mobile jsonwp spec. I'd make sure there are tests for this to make sure the code actually works.

SrinivasanTarget:
Sounds Interesting as both the API's are listed under mobile wire here: 

https://github.com/appium/node-mobile-json-wire-protocol/blob/master/lib/routes.js#L278
https://github.com/appium/node-mobile-json-wire-protocol/blob/master/lib/routes.js#L347

src/main/java/io/appium/java_client/android/AndroidDriver.java
SrinivasanTarget: Defined now

SrinivasanTarget:
I have just gone through the server code now both API's are implemented in same way here, 

https://github.com/appium/appium-android-driver/blob/master/lib/commands/network.js#L34
https://github.com/appium/appium-android-driver/blob/master/lib/commands/network.js#L72

Also it seems API is not implemented for Real devices.

TikhomirovSergey:
I will keep it opened for some time till the reason why API is duplicated on the server side is clear.

imurchie:
The `toggle...` commands were how it used to be done, before we implemented the `setNetworkConnection` with bitmask, which is how the [spec defines it](https://github.com/SeleniumHQ/mobile-spec/blob/master/spec-draft.md#connectiontype) (you'll notice that the toggle commands are all behind the `appium` vendor ([see here](https://github.com/appium/node-mobile-json-wire-protocol/blob/master/lib/routes.js#L344-L354))).

Since implementing the latter, the former methods basically became aliases (except that they toggle, so they need to find the current state and change it to the opposite). They seem less generally useful. As far as I know we never deprecated them, and I haven't heard of any plans to remove them.

SrinivasanTarget:
@imurchie @TikhomirovSergey Now I have tested both the APIs in emulators.
setNetworkConnection API doesn't work fine for me in API levels > 19. But toggle API's works fine for me in almost all API levels 19,21,22 (tested so far).

API Level 21 Logs:- https://gist.github.com/SrinivasanTarget/5427a31d8443713005db22a25076605c

SrinivasanTarget:
@TikhomirovSergey I have run through the APIs today on real devices, above findings matches with real devices as well.

Here we go with real devices,
SetNetworkConnection APIs are not working for API levels > 19
toggleAPIs works fine on almost all the APIs till API level 22 (Haven't tested API level 23 yet)

Real Device Failure Logs: https://gist.github.com/SrinivasanTarget/f1047375b448285663ce68a41f0253a3

SrinivasanTarget:
Have fixed setNetworkConnections Endpoints in Server: https://github.com/appium/appium-android-driver/pull/148.

@TikhomirovSergey I'm closing this PR here.

PR #397, opened 2016-05-28
The GetsNamedTextField interface has become deprecated

TikhomirovSergey:
## Change list

The GetsNamedTextField interface has become deprecated

## Types of changes

## Details
I think that the using of hardcoded locators is the bad practice. Also I think that user can handle these actions the better way.


SrinivasanTarget:
@TikhomirovSergey LGTM 👍

PR #432, opened 2016-07-01
Use weaker interface HasIdentity instead of class RemoteWebElement

asolntsev:
## Change list
* io.appium.java_client.TouchAction
* io.appium.java_client.pagefactory_tests.widgets.ios.simple.IOSMovie
 
## Types of changes

## Details

In java_client code, `WebElement` parameter is casted to `RemoteWebElement`. 
It makes hard to create proxy object for `WebElement` (like Selenide and probably some other frameworks do). 

Actually only method `getId()` is used. This method is declared in interface `HasIdentity`. 

This pull request changes few places where `WebElement` is casted to `RemoteWebElement` so that it's not casted to `HasIdentity` instead of `RemoteWebElement`.

Old code:

```java
new ActionParameter("press", (RemoteWebElement) el);
```

new code:

```java
new ActionParameter("press", (HasIdentity) el);
```

### DIFF
--- a/src/main/java/io/appium/java_client/TouchAction.java
+++ b/src/main/java/io/appium/java_client/TouchAction.java
@@ -20,7 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
 import org.openqa.selenium.WebElement;
-import org.openqa.selenium.remote.RemoteWebElement;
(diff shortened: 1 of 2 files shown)


TikhomirovSergey:
@asolntsev Thank you. I'm merging...

asolntsev:
Wow, that was fast!

PR #459, opened 2016-08-16
Alert API Implementation for iOS

SrinivasanTarget:
## Change list

Alert Handling for iOS.Added below methods,
```
1)getAlertText
2)setAlertText(String value)
3)acceptAlert
4)dismissAlert
```

API is here: https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L403

API's were not implemented for Android Driver on server side.We can port it to android driver on client side post server implementation.

New W3C API's here: https://github.com/appium/appium-base-driver/blob/master/lib/mjsonwp/routes.js#L414 are not been adapted by Appium (Legacy drivers on server) yet.
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/ios/IOSDriver.java
+++ b/src/main/java/io/appium/java_client/ios/IOSDriver.java
@@ -16,6 +16,7 @@
 
 package io.appium.java_client.ios;
 
+import static io.appium.java_client.MobileCommand.prepareArguments;
 import static io.appium.java_client.ios.IOSMobileCommandHelper.hideKeyboardCommand;
 import static io.appium.java_client.ios.IOSMobileCommandHelper.lockDeviceCommand;
 import static io.appium.java_client.ios.IOSMobileCommandHelper.shakeCommand;
@@ -27,15 +28,19 @@ import io.appium.java_client.ios.internal.JsonToIOSElementConverter;
 import io.appium.java_client.remote.MobilePlatform;
 import io.appium.java_client.service.local.AppiumDriverLocalService;
 import io.appium.java_client.service.local.AppiumServiceBuilder;
-
+import org.openqa.selenium.Alert;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.WebElement;
+import org.openqa.selenium.remote.DriverCommand;
 import org.openqa.selenium.remote.HttpCommandExecutor;
+import org.openqa.selenium.remote.Response;
 import org.openqa.selenium.remote.http.HttpClient;
+import org.openqa.selenium.security.Credentials;
 
 import java.net.URL;
 import java.util.List;
+
 
 /**
  * @param <T> the required type of class which implement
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
@SrinivasanTarget 
Like this solution but I am not agree with this implementation. It duplicates methods of the `org.openqa.selenium.Alert` API. Why could we use Selenium API more high? Like the sample below.

```java
public class IOSDriver<T extends WebElement>
    extends AppiumDriver<T>
    implements IOSDeviceActionShortcuts,
        FindsByIosUIAutomation<T> {
...

    @Override public TargetLocator switchTo() {
        return new InnerTargetLocator();
    }

    private class InnerTargetLocator extends RemoteTargetLocator {
        @Override public Alert alert() {
            execute(DriverCommand.GET_ALERT_TEXT);
            return new IOSAlert();
        }
    }

    class IOSAlert implements Alert {
        @Override
        public void dismiss() {
            ...
        }

        @Override
        public void accept() {
            ...
        }

        @Override
        public String getText() {
            ...
        }

        @Override
        public void sendKeys(String keysToSend) {
            ...
        }

        @Override
        public void setCredentials(Credentials credentials) {
            ...
        }

        @Override
        public void authenticateUsing(Credentials credentials) {
            ...
        }
    }
}
```

TikhomirovSergey:
There are new checkstyle issues:
image

@SrinivasanTarget Please get it fixed

src/test/java/io/appium/java_client/ios/IOSAlertTest.java
```diff
@@ -16,25 +16,36 @@
...
+        wating.until(alertIsPresent());
+        driver.switchTo().alert().dismiss();
+    }
+
+    @Test public void getAlertTextTest() {
+        driver.findElement(MobileBy
+            .IosUIAutomation(".elements().withName(\"show alert\")")).click();
+        WebDriverWait wating = new WebDriverWait(driver, 10000);
+        wating.until(alertIsPresent());
+        assertNotNull(driver.switchTo().alert().getText());
```
TikhomirovSergey: @SrinivasanTarget 
I would like to advive you to
```java
assertTrue(!StringUtils.isBlank(driver.switchTo().alert().getText()));
```
What if the empty string would be returned? :)

src/main/java/io/appium/java_client/ios/IOSDriver.java
```diff
@@ -220,4 +224,50 @@ public T findElementByIosUIAutomation(String using)
...
+    class IOSAlert implements Alert {
+
+        private final Alert alert;
+
+        IOSAlert(Alert alert) {
+            this.alert = alert;
+        }
+
+        @Override public void dismiss() {
+            execute(DriverCommand.DISMISS_ALERT);
```
TikhomirovSergey: Why don't you use 
```java
@Override public void dismiss() {
    alert.dismiss();
}
```
instead?

src/main/java/io/appium/java_client/ios/IOSDriver.java
```diff
@@ -220,4 +224,50 @@ public T findElementByIosUIAutomation(String using)
...
+        }
+
+        @Override public void accept() {
+            execute(DriverCommand.ACCEPT_ALERT);
```
TikhomirovSergey: Why don't you use 
```java
@Override public void dismiss() {
    alert.accept();
}
```
instead?

src/main/java/io/appium/java_client/ios/IOSDriver.java
SrinivasanTarget: @TikhomirovSergey ```alert.dismiss()``` cannot be used as ```RemoteWebDriver``` uses ```RemoteWebDriver.this.execute("dimissAlertW3C")``` i.e W3C API which we didn't adopt yet. It is the same for ```alert.accept()``` as well

SrinivasanTarget:
@TikhomirovSergey Thanks.I have incorporated the changes now and squashed the commits.

PR #468, opened 2016-09-09
fix element timeout changing page object timeout issue

### LINKED ISSUE #467: page object pattern, field with "@WithTimeout" would change the global implicit wait time
## Description
page object pattern, field with "@WithTimeout" would change the global implicit wait time

## Environment
Always exists.  see code analysis below

## Details
in method [AppiumElementLocatorFactory#createLocator](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocatorFactory.java#L47):
* if the page object field contains annoation `WithTimeout`, it would create a new `TimeOutDuration` instance.
* This instance would be passed to `AppiumElementLocator`.
* When element tries to find the element. It would invoke method [AppiumElementLocator#waitFor](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java#L92). See the line in `finally` block. it would change the global implicit wait time using the `WithTimeout` value.

So this would make the page object and global appium driver implicit wait both change to that value. It would mess the implicit timeout when there are many `WithTimeout` elements in page objects that contains `driver.findElement` methods.

There is no getter for implicit wait by design. so it is not easy to fix this.

Solution suggestion: 
IMO changing constructors that also passes original timeout duration as argument, in finally block it would set back to original timeout duration.

truebit:
## Change list

fix element timeout changing page object timeout issue 
 
## Types of changes

## Details
this bug fixes #467

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -42,6 +42,7 @@ class AppiumElementLocator implements CacheableLocator {
     final boolean shouldCache;
     final By by;
     final TimeOutDuration timeOutDuration;
+    private final TimeOutDuration originalTimeOutDuration;
     final WebDriver originalWebDriver;
     private final SearchContext searchContext;
     private final WaitingFunction waitingFunction;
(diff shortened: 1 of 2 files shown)


TikhomirovSergey:
Ok.
Please do not forget to add a test next time

Issue #471, opened 2016-09-12
Windows Mobile automation support. 


TikhomirovSergey:
## Description

Windows Mobile automation support.

PR #410, opened 2016-06-08
add support to set UiAutomator Congfigurator values

truebit:
## Change list
This is a sub PR for https://github.com/appium/appium/pull/6561
 
## Types of changes

What types of changes are you proposing/introducing to Java client?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details
This is a sub PR for https://github.com/appium/appium/pull/6561

PR #477, opened 2016-09-25
Android setting engine

TikhomirovSergey:
## Change list

- `io.appium.java_client.AppiumSetting` became deprecated. It was moved to `io.appium.java_client.android.Setting`.

- the method `io.appium.java_client.AppiumDriver#getSettings()` was completely moved to `io.appium.java_client.android.AndroidDriver`.

- the new interface `io.appium.java_client.android.HasSettings` was added. It has default implementation. And this interface is implemented by AndroidDriver.
Methods:
  - `void setSetting(Setting setting, Object value) `
  - `JsonObject getSettings() `
  - `void ignoreUnimportantViews(Boolean compress)`
  - `void configuratorSetWaitForIdleTimeout(int timeout) `
  - `void configuratorSetWaitForSelectorTimeout(int timeout) `
  - `void configuratorSetScrollAcknowledgmentTimeout(int timeout) `
  - `void configuratorSetKeyInjectionDelay(int delay)`
  - `void configuratorSetActionAcknowledgmentTimeout(int timeout)`
This change was proposed at #410 by @truebit 
 
## Types of changes

#471

### DIFF
--- a/src/main/java/io/appium/java_client/android/Setting.java
+++ b/src/main/java/io/appium/java_client/android/Setting.java
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package io.appium.java_client.android.settings;
+package io.appium.java_client.android;
 
 /**
  * Enums defining constants for Appium Settings which can be set and toggled during a test session.
@@ -32,31 +32,4 @@ public enum Setting {
     public String toString() {
         return this.name;
     }
-
-    /**
-     * Contains possible values of the `ignoreUnimportantViews` setting.
-     * It helps to define whether Android devices should use `setCompressedLayoutHeirarchy()`
-     * which ignores all views that are marked IMPORTANT_FOR_ACCESSIBILITY_NO
-     * or IMPORTANT_FOR_ACCESSIBILITY_AUTO (and have been deemed not important
-     * by the system), in an attempt to make things less confusing or faster.
-     */
-    public static enum IgnoreUnimportantViews {
-        YES(true),
-        NO(false);
-
-        private final Boolean ignore;
-
-        IgnoreUnimportantViews(Boolean ignore) {
-            this.ignore = ignore;
-        }
-
-        /**
-         * @return value which means to ignore unimportant views if true,
-         * doesn't ignore otherwise.
-         */
-        public Boolean value() {
-            return ignore;
-        }
-    }
-
 }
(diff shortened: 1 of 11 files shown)


TikhomirovSergey:
@SrinivasanTarget @truebit 
Could you review this PR?

truebit:
the settings api has been changed to top level as requested by @jlipps in https://github.com/appium/appium-android-driver/pull/153
you could find the doc in https://github.com/truebit/appium/blob/master/docs/en/advanced-concepts/settings.md#supported-settings

So these settings api requests should be the same as `ignoreUnimportantViews` in top level instead of `configurator/xxxTimeout`

TikhomirovSergey:
@truebit 
Yes and proposed code actualy does that:
https://github.com/appium/java-client/pull/477/files#diff-7daaa6d2742d959db92558292282ea56R41
https://github.com/appium/java-client/pull/477/files#diff-7daaa6d2742d959db92558292282ea56R78
https://github.com/appium/java-client/pull/477/files#diff-cebdf3eaee01b491314c7101b3103f7bR298
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/MobileCommand.java#L92

Isn't it?

src/main/java/io/appium/java_client/android/Setting.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.appium.java_client.android;
+
+/**
+ * Enums defining constants for Appium Settings which can be set and toggled during a test session.
+ */
+public enum Setting {
+
+    IGNORE_UNIMPORTANT_VIEWS("ignoreUnimportantViews"),
+    CONFIGURATOR("configurator");
```
truebit: no, this code is still the two level configurator. Now it should be like `ignoreUnimportantViews`. Every setting here for configurators should be individivual

truebit:
all below settings are at same level like `ignoreUnimportantViews`. there is no 'configurator' sub setting
> "actionAcknowledgmentTimeout" - Int which is the same as setActionAcknowledgmentTimeout. If a negative value is given, it would set to default(3 * 1000 milliseconds)
> 
> "keyInjectionDelay" - Int which is the same as setKeyInjectionDelay. If a negative value is given, it would set to default(0 milliseconds)
> 
> "scrollAcknowledgmentTimeout" - Int which is the same as setScrollAcknowledgmentTimeout. If a negative value is given, it would set to default(200 milliseconds)
> 
> "waitForIdleTimeout" - Int which is the same as setWaitForIdleTimeout. If a negative value is given, it would set to default(10 * 1000 milliseconds)
> 
> "waitForSelectorTimeout" - Int which is the same as setWaitForSelectorTimeout. If a negative value is given, it would set to default(10 * 1000 milliseconds)

TikhomirovSergey:
Ok. Will iptrove this a bit later.

PR #509, opened 2016-11-07
TouchID Implementation [iOS Sim Only]

dpgraham:
## Change list

Added touchId command
 
## Types of changes

## Details

Touch ID is a feature in Apple products (see https://support.apple.com/en-ca/HT201371). This endpoint simulates the behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ More can be found in the docs, but here's a quick list of features which this pr
 - endTestCoverage()
 - isLocked()
 - shake()
+- touchId()
 - getSessionDetails()
 - openNotifications()
 - Context Switching: .context(), .getContextHandles(), getContext())
--- a/README.md
+++ b/README.md
@@ -53,7 +53,6 @@ More can be found in the docs, but here's a quick list of features which this pr
 - endTestCoverage()
 - isLocked()
 - shake()
-- touchId()
 - getSessionDetails()
 - openNotifications()
 - Context Switching: .context(), .getContextHandles(), getContext())
--- a/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
+++ b/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
@@ -18,7 +18,6 @@ package io.appium.java_client.ios;
 
 import static io.appium.java_client.ios.IOSMobileCommandHelper.hideKeyboardCommand;
 import static io.appium.java_client.ios.IOSMobileCommandHelper.shakeCommand;
-import static io.appium.java_client.ios.IOSMobileCommandHelper.touchIdCommand;
 
 import io.appium.java_client.CommandExecutionHelper;
 import io.appium.java_client.DeviceActionShortcuts;
@@ -56,15 +55,6 @@ public interface IOSDeviceActionShortcuts extends DeviceActionShortcuts {
      */
     default void shake() {
         CommandExecutionHelper.execute(this, shakeCommand());
-    }
-    
-    /**
-     * Simulate touchId event
-     *
-     * @param match If true, simulates a successful fingerprint scan. If false, simulates a failed fingerprint scan
-     */
-    default void touchId(boolean match) {
-    	CommandExecutionHelper.execute(this, touchIdCommand(match));
     }
 
 }
--- a/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
+++ b/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
@@ -61,7 +61,7 @@ public interface IOSDeviceActionShortcuts extends DeviceActionShortcuts {
     /**
      * Simulate touchId event
      *
-     * @param match Are we simulating a successful fingerprint scan?
+     * @param match If true, simulates a successful fingerprint scan. If false, simulates a failed fingerprint scan
      */
     default void touchId(boolean match) {
     	CommandExecutionHelper.execute(this, touchIdCommand(match));
--- a/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
+++ b/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
@@ -18,6 +18,7 @@ package io.appium.java_client.ios;
 
 import static io.appium.java_client.ios.IOSMobileCommandHelper.hideKeyboardCommand;
 import static io.appium.java_client.ios.IOSMobileCommandHelper.shakeCommand;
+import static io.appium.java_client.ios.IOSMobileCommandHelper.touchIdCommand;
 
 import io.appium.java_client.CommandExecutionHelper;
 import io.appium.java_client.DeviceActionShortcuts;
@@ -55,6 +56,15 @@ public interface IOSDeviceActionShortcuts extends DeviceActionShortcuts {
      */
     default void shake() {
         CommandExecutionHelper.execute(this, shakeCommand());
+    }
+    
+    /**
+     * Simulate touchId event
+     *
+     * @param match Are we simulating a successful fingerprint scan?
+     */
+    default void touchId(boolean match) {
+    	CommandExecutionHelper.execute(this, touchIdCommand(match));
     }
 
 }
--- a/src/main/java/io/appium/java_client/ios/IOSMobileCommandHelper.java
+++ b/src/main/java/io/appium/java_client/ios/IOSMobileCommandHelper.java
@@ -85,7 +85,7 @@ public class IOSMobileCommandHelper extends MobileCommand {
     /**
      * This method forms a {@link java.util.Map} of parameters for the touchId simulator
      * 
-     * @param match Are we simulating a successful fingerprint scan?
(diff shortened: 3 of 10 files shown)


dpgraham:
*Not sure why the README.md was diffed that way. I only added one line under the 'Added Functions' header

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
```diff
@@ -31,8 +34,6 @@
 
 import java.io.File;
 
-import static org.junit.Assert.assertEquals;
-
 public class XCUIAutomationTest {
```
SrinivasanTarget: @dpgraham Can you add some tests here?

src/main/java/io/appium/java_client/ios/IOSMobileCommandHelper.java
```diff
@@ -81,4 +81,15 @@
         return new AbstractMap.SimpleEntry<>(
                 SHAKE, ImmutableMap.<String, Object>of());
     }
+    
+    /**
+     * This method forms a {@link java.util.Map} of parameters for the touchId simulator
+     * 
+     * @param match Are we simulating a successful fingerprint scan?
```
SrinivasanTarget: Why is it a question here?

src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
```diff
@@ -56,5 +57,14 @@ default void hideKeyboard(String strategy, String keyName) {
     default void shake() {
         CommandExecutionHelper.execute(this, shakeCommand());
     }
+    
+    /**
+     * Simulate touchId event
+     *
+     * @param match Are we simulating a successful fingerprint scan?
+     */
+    default void touchId(boolean match) {
```
SrinivasanTarget: Can you add in doc it is Simulator only?

README.md
```diff
@@ -1,311 +1,312 @@
...
+- zoom()
+- isAppInstalled()
+- installApp()
+- removeApp()
+- launchApp()
+- closeApp()
+- endTestCoverage()
+- isLocked()
+- shake()
+- touchId()
```
SrinivasanTarget: Can you revert this change? We will apply this change post release?

src/main/java/io/appium/java_client/ios/IOSMobileCommandHelper.java
dpgraham: It's how I've documented boolean parameters in the past. I'll change it to be more in line with how it's normally documented.

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
dpgraham: I added the test. It doesn't do any assertions because there isn't anything to check, it just calls the methods to verify that there aren't any exceptions.

dpgraham:
Thanks @SrinivasanTarget, I made the changes you requested. If you could review them and they're good let me know so that I can squash the changes and merge them.

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
```diff
@@ -91,4 +92,14 @@
...
         assertEquals(driver.rotation(), landscapeLeftRotation);
     }
+
+    @Test public void testTouchId() {
+        try {
+            driver.touchId(true);
+            driver.touchId(false);
+            assertEquals(true, true);
+        } catch(Exception e){
+            assertEquals(true, false);
```
TikhomirovSergey: @dpgraham I think it is more senseful to throw an exception.
Otherwice we have a risk to produce and invalid build or to not react to server changes.

src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
```diff
@@ -56,5 +57,14 @@ default void hideKeyboard(String strategy, String keyName) {
     default void shake() {
         CommandExecutionHelper.execute(this, shakeCommand());
     }
+
```
TikhomirovSergey: I am against the adding to the `IOSDeviceActionShortcuts`. This interface is going to be deprecated. Please take a look [at this working branch](https://github.com/TikhomirovSergey/java-client/tree/SrinivasanTarget-TouchActionSplitting/src/main/java/io/appium/java_client/ios). [IOSDeviceActionShortcuts](https://github.com/TikhomirovSergey/java-client/blob/SrinivasanTarget-TouchActionSplitting/src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java) is going to be deprecated. 

It is better to add such interface:

```java
package io.appium.java_client.ios;

public interface ChecksTouchId extends ExecutesMethod {

    /**
    * Simulate touchId event
    *
    * @param match If true, simulates a successful fingerprint scan. If false, simulates a failed fingerprint scan
    */     
    default void touchId(boolean match) {
        CommandExecutionHelper.execute(this, touchIdCommand(match));
    }
}
```

and make IOSDriver implement it.

TikhomirovSergey:
@dpgraham
Sorry for the delaying. Usually busy. Please take a look at remarks.

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
```diff
@@ -91,4 +92,14 @@
         driver.rotate(landscapeLeftRotation);
         assertEquals(driver.rotation(), landscapeLeftRotation);
     }
+
+    @Test public void testTouchId() {
+        try {
+            driver.touchId(true);
+            driver.touchId(false);
+            assertEquals(true, true);
```
SrinivasanTarget: Actual Usecase of API ```touchID``` is to unlock or sign in to any application using touchID. Since we dont have our test app built with any registration or sign in feature, we cannot use that scenario.

Otherwise Can we lock the simulator and unlock the simulator using ```driver.touchID(true)```.Ideal case this should be working fine.

src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
```diff
@@ -56,5 +57,14 @@ default void hideKeyboard(String strategy, String keyName) {
     default void shake() {
         CommandExecutionHelper.execute(this, shakeCommand());
     }
+    
+    /**
+     * Simulate touchId event
+     *
+     * @param match If true, simulates a successful fingerprint scan. If false, simulates a failed fingerprint scan
```
SrinivasanTarget: If i'm not wrong, this will fail as per google style each doc line should end with a period.

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
SrinivasanTarget: So test scenario should be something like,

```
   try {
            driver.lockDevice(2);
            driver.touchId(true);
        } catch (Exception e) {
            assertTrue(false);
        }
```

SrinivasanTarget:
@dpgraham I've reviewed again now. Please take a look.

src/main/java/io/appium/java_client/ios/IOSDeviceActionShortcuts.java
dpgraham: Thanks for the input Sergey. I made the change.

dpgraham:
I've updated the code again

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
```diff
@@ -91,4 +92,15 @@
...
         assertEquals(driver.rotation(), landscapeLeftRotation);
     }
+
+    @Test public void testTouchId() {
+        try {
+            driver.performTouchID(true);
+            driver.performTouchID(false);
+            assertEquals(true, true);
+        } catch (Exception e) {
+            assertEquals(true, false);
```
TikhomirovSergey: Test should throw the exception. 

At least you can do that 

```java
try {
//do something
//assert the result
} catch (Exception e) {
    throw e;
} finally {
//attemt to stabilize the simulator
//or something else
}
```

src/main/java/io/appium/java_client/ios/IOSMobileCommandHelper.java
```diff
@@ -81,4 +81,15 @@
         return new AbstractMap.SimpleEntry<>(
                 SHAKE, ImmutableMap.<String, Object>of());
     }
+    
+    /**
+     * This method forms a {@link java.util.Map} of parameters for the touchId simulator
+     * 
+     * @param match If true, simulates a successful fingerprint scan. If false, simulates a failed fingerprint scan
```
SrinivasanTarget: Doc should end with a period everywhere.

dpgraham:
Changes made

TikhomirovSergey:
@dpgraham Everuthing is ok for me. But could please improve the proposed test the way I have asked?
@SrinivasanTarget ?

src/test/java/io/appium/java_client/ios/XCUIAutomationTest.java
SrinivasanTarget: @dpgraham Can you update tests here as suggested please?

PR #525, opened 2016-11-25
Cleanup tests in ServerBuilderTest.

alexander-poulikakos:
## Change list

Refactored unit tests.
 
## Types of changes

- [x] No changes in production code

## Details

Cleanup unit tests.

### DIFF
--- a/src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
+++ b/src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
@@ -80,7 +80,7 @@ public class ServerBuilderTest {
     }
 
     @After
-    public void teardown() throws Exception {
+    public void tearDown() throws Exception {
         if (service != null) {
             service.stop();
         }
--- a/src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
+++ b/src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
@@ -1,14 +1,28 @@
 package io.appium.java_client.localserver;
 
+import static io.appium.java_client.remote.AndroidMobileCapabilityType.APP_ACTIVITY;
+import static io.appium.java_client.remote.AndroidMobileCapabilityType.APP_PACKAGE;
+import static io.appium.java_client.remote.AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE;
+import static io.appium.java_client.remote.MobileCapabilityType.APP;
+import static io.appium.java_client.remote.MobileCapabilityType.FULL_RESET;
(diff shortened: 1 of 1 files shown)


alexander-poulikakos:
In the unit tests that changes the URL (`checkAbilityToStartService*UsingFlag`) I chnaged so the Assertion of URL is dome before starting the service. This makes it fail faster, when failing. No need to wait for client to timeout.

Moved duplicated try/catch clauses to @After method to make code cleaner.

src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
```diff
@@ -1,434 +1,311 @@
...
+            if (testIP != null) {
+                break;
+            }
+        }
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        if (service != null) {
+            service.stop();
```
TikhomirovSergey: @alexander-poulikakos 
Please don't forget to 
```
service = null;
```

src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
```diff
@@ -1,434 +1,311 @@
...
+        }
+        if (stream != null) {
+            stream.close();
```
TikhomirovSergey: Please don't forget to 
```
stream = null;
```

src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
```diff
@@ -1,434 +1,311 @@
...
+        }
+        if (file != null && file.exists()) {
+            file.delete();
```
TikhomirovSergey: Please don't forget to 
```
file = null;
```

alexander-poulikakos: No need to se values to null in @After method. JUnit will make a new instance of ServerBuilderTest for each @Test which will automatically reset all member fields.

src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
alexander-poulikakos: No need to se values to null in @After method. JUnit will make a new instance of ServerBuilderTest for each @Test which will automatically reset all member fields.

src/test/java/io/appium/java_client/localserver/ServerBuilderTest.java
alexander-poulikakos: No need to se values to null in @After method. JUnit will make a new instance of ServerBuilderTest for each @Test which will automatically reset all member fields.

PR #545, opened 2016-12-23
Update Page object tools to XCUIT mode

TikhomirovSergey:
## Change list

- new annotation `io.appium.java_client.pagefactory.iOSXcuitFindBy` was provided 
- supporting tools were actualized
 
## Types of changes

- [ ] No changes in production code.
- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

It was done because users still can need to support old automation (for iOS < 10) and xCuit mode for the same test scope.

PR #546, opened 2016-12-25
Addition to #545

SrinivasanTarget:
## Change list

-  Added Tests for iOS XCUIT mode.
-  Gradle publish tasks are updated

 
## Types of changes

@TikhomirovSergey FYI

### DIFF
--- a/src/test/java/io/appium/java_client/ios/AppXCUITTest.java
+++ b/src/test/java/io/appium/java_client/ios/AppXCUITTest.java
@@ -1,0 +1,39 @@
+package io.appium.java_client.ios;
+
+import io.appium.java_client.remote.AutomationName;
+import io.appium.java_client.remote.IOSMobileCapabilityType;
+import io.appium.java_client.remote.MobileCapabilityType;
+import io.appium.java_client.service.local.AppiumDriverLocalService;
+import io.appium.java_client.service.local.AppiumServerHasNotBeenStartedLocallyException;
+import org.junit.BeforeClass;
+import org.openqa.selenium.remote.DesiredCapabilities;
+
+import java.io.File;
+
+public class AppXCUITTest extends BaseIOSTest {
+
+    /**
+     * initialization.
+     */
+    @BeforeClass public static void beforeClass() throws Exception {
+        service = AppiumDriverLocalService.buildDefaultService();
+        service.start();
+
+        if (service == null || !service.isRunning()) {
+            throw new AppiumServerHasNotBeenStartedLocallyException("An appium server node is not started!");
+        }
+
+        File appDir = new File("src/test/java/io/appium/java_client");
+        File app = new File(appDir, "TestApp.app.zip");
+        DesiredCapabilities capabilities = new DesiredCapabilities();
+        capabilities.setCapability(MobileCapabilityType.BROWSER_NAME, "");
+        capabilities.setCapability(MobileCapabilityType.PLATFORM_VERSION, "10.1");
+        capabilities.setCapability(MobileCapabilityType.DEVICE_NAME, "iPhone 6");
+        //sometimes environment has performance problems
+        capabilities.setCapability(IOSMobileCapabilityType.LAUNCH_TIMEOUT, 500000);
+        capabilities
+            .setCapability(MobileCapabilityType.AUTOMATION_NAME, AutomationName.IOS_XCUI_TEST);
+        capabilities.setCapability(MobileCapabilityType.APP, app.getAbsolutePath());
+        driver = new IOSDriver<>(service.getUrl(), capabilities);
+    }
+}
--- a/src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
+++ b/src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
@@ -31,6 +31,7 @@ import io.appium.java_client.MobileElement;
 import io.appium.java_client.ios.AppXCUITTest;
 import io.appium.java_client.pagefactory.AppiumFieldDecorator;
 import io.appium.java_client.pagefactory.HowToUseLocators;
+import io.appium.java_client.pagefactory.iOSFindBy;
 import io.appium.java_client.pagefactory.iOSXCUITFindBy;
 import org.junit.Before;
 import org.junit.FixMethodOrder;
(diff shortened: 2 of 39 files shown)


SrinivasanTarget:
@TikhomirovSergey I see this is first time javadoc task is running in Travis job :) Its because of javadoc task refractor in build.gradle file :)

TikhomirovSergey:
@SrinivasanTarget The travis build is failed

SrinivasanTarget:
Yeah I have to find a way to ignore signing in Travis.

src/test/java/io/appium/java_client/ios/AppXCUITTest.java
```diff
@@ -0,0 +1,39 @@
...
+import io.appium.java_client.remote.IOSMobileCapabilityType;
+import io.appium.java_client.remote.MobileCapabilityType;
+import io.appium.java_client.service.local.AppiumDriverLocalService;
+import io.appium.java_client.service.local.AppiumServerHasNotBeenStartedLocallyException;
+import org.junit.BeforeClass;
+import org.openqa.selenium.remote.DesiredCapabilities;
+
+import java.io.File;
+
+public class AppXCUITTest extends BaseIOSTest {
```
TikhomirovSergey: 👍

src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
```diff
@@ -0,0 +1,120 @@
...
+    @iOSXCUITFindBy(iOSNsPredicate = "name = 'IntegerB'")
+    @iOSXCUITFindBy(accessibility = "IntegerB")
+    private MobileElement textField2;
+
+    @iOSXCUITFindBy(iOSNsPredicate = "name ENDSWITH 'Gesture'")
+    private MobileElement gesture;
+
+    @iOSXCUITFindBy(className = "XCUIElementTypeSlider")
+    private MobileElement slider;
+
```
TikhomirovSergey: @SrinivasanTarget 
Everything is ok. But I think coverage could be enhanced. There is no case like

```java
@iOSFindBy(id = "someId")
private MobileElement foundLikeIOSId;
```

because I think some locators (like "id") are compatible with both modes. It similar to the [test of the selendroid mode](https://github.com/appium/java-client/blob/master/src/test/java/io/appium/java_client/pagefactory_tests/SelendroidModeTest.java#L67).

TikhomirovSergey:
@SrinivasanTarget Everything is ok but could you provide one more test as it was mentioned in remarks?

SrinivasanTarget:
Sure will do that :)

src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
TikhomirovSergey: @SrinivasanTarget 
Also you could add this test
```java
@iOSFindBy(iOSAutomation = "some searching script")
@iOSXCUITFindBy(iOSNsPredicate = "some predicate") //it should be used for the searching instead of 
//@iOSFindBy
private MobileElement foundLikeIOSId;
```

TikhomirovSergey:
@SrinivasanTarget 
There is one more remark

SrinivasanTarget:
Codacy is GREEN now. Will try to fix travis.

src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
saikrishna321: @SrinivasanTarget @TikhomirovSergey when we initialize IOSDriver and provide the XCUITest as AutomationName only this case will it pick the iOSXCUITFindBy ?

TikhomirovSergey:
@SrinivasanTarget Yes. Everything is ok. Please get build successful

src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
TikhomirovSergey: @saikrishna321 When there is only ```java @iOSFindBy ``` then it will be used. 

When there are 

```java
@iOSFindBy
@iOSXCUITFindBy
```
then ```java @iOSXCUITFindBy``` will be preferred. Selendroid mode works the same way.

SrinivasanTarget:
@TikhomirovSergey Yeah Sure.Its weird. I don't see travis running our command. Command works locally fine.

SrinivasanTarget:
@TikhomirovSergey Travis is GREEN now. Its running our commands now. Moved it to run under container based infra.

src/test/java/io/appium/java_client/pagefactory_tests/XCUITModeTest.java
saikrishna321: @TikhomirovSergey it works that way in Selendroid, coz the automatioName is set to selendroid. It should be the same here as well, if the automatioName is set to XCUITest only then @iOSXCUITFindBy should be preferred. Thoughts ?

SrinivasanTarget: @saikrishna321 Yes you are correct.

TikhomirovSergey: @saikrishna321 I can't get...

It is designed the same way as ```java @AndroidFindBy```/```java @SelendroidFindBy```. It was designed this way because it is possible that old automation for iOS <= 9.3 and the new automation should be supported by one test scope. 

The algorithm:
```java
@iOSFindBy
@iOSXCUITFindBy
```

if you run new mode then ```java @iOSXCUITFindBy``` is used. If you use legacy mode then ```java @iOSFindBy``` is used instead.

```java
@iOSFindBy
```
is used by both modes. 

```java
@iOSXCUITFindBy
```

is used by the new mode. If you launch legacy mode then page object tools create ```java ByIdOrName```

The same works for Android/Selendroid.

So what the concern?

saikrishna321: @TikhomirovSergey no concern, well explained. Cheers

SrinivasanTarget:
Linking gists justs for internal reference: iOS 10 - https://gist.github.com/SrinivasanTarget/a75db7dfcff1e237ae6ad9919ffa8219
iOS 9 - https://gist.github.com/SrinivasanTarget/07046e5208cc431f7d81f19c7d3e663c

PR #560, opened 2017-01-27
new gradle plugins block + fatJar

shauvik:
To build fatJar, apply this PR patch and simply call `./gradlew shadowJar`

## Change list
* Adding support for the new gradle plugins block
* Using the shadow plugin to create a fatJar for appium java client.
 
## Types of changes

## Details

* Usage of plugins block makes buildscript terse
* Allowing building fatJars helps makes life easy for cmd line usage.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,3 +1,8 @@
+plugins {
+  id "org.owasp.dependencycheck" version "1.4.5"
+  id "com.github.johnrengelman.shadow" version "1.2.4"
+}
+
 apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'eclipse'
@@ -15,19 +20,6 @@ repositories {
         url "http://repo.maven.apache.org/maven2"
     }
 }
-
-buildscript {
-    repositories {
-        maven {
-            url "http://repo.maven.apache.org/maven2"
-        }
-    }
-    dependencies {
-        classpath "org.owasp:dependency-check-gradle:1.4.0"
-    }
-}
-
-apply plugin: "org.owasp.dependencycheck"
 
 configurations {
     ecj
--- a/build.gradle
+++ b/build.gradle
@@ -1,8 +1,3 @@
-plugins {
-  id "org.owasp.dependencycheck" version "1.4.5"
-  id "com.github.johnrengelman.shadow" version "1.2.4"
-}
-
 apply plugin: 'java'
 apply plugin: 'maven'
 apply plugin: 'eclipse'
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -1,3 +1,8 @@
+plugins {
+  id "org.owasp.dependencycheck" version "1.4.5"
+  id "com.github.johnrengelman.shadow" version "1.2.4"
```
SrinivasanTarget: I see shadow by default uses ```jfrog``` repo but we are publishing only to maven repo. Where are we specifying class path for owasp dependency check.

shauvik: Specifying it using the plugins block makes it download the dependencies from plugins.gradle.org... Specifically here: https://plugins.gradle.org/plugin/org.owasp.dependencycheck

There is also an option of the longer buildscript{} if you wanted to get dependencies the old way..

SrinivasanTarget: I got it but does it set the classpath as well? Also as i said before we need to explicitly set the repository as maven repo.

shauvik: So, I did some digging. Seems like gradle plugins block uses http://plugins.gradle.com/m2 which simply redirects all requests to jcenter.

```
~ curl -I https://plugins.gradle.org/m2/junit/junit/4.12/junit-4.12.pom
HTTP/1.1 303 See Other
Date: Fri, 27 Jan 2017 10:32:44 GMT
Content-Length: 0
Connection: keep-alive
Set-Cookie: __cfduid=dfb3ae2221aa16499509023673d8240ee1485513164; expires=Sat, 27-Jan-18 10:32:44 GMT; path=/; domain=.gradle.org; HttpOnly
Location: https://jcenter.bintray.com/junit/junit/4.12/junit-4.12.pom
Via: 1.1 vegur
Server: cloudflare-nginx
CF-RAY: 327b8adfa92f0a00-ATL
```

Also, it seems that the gradle plugins website keeps the classpath mapping, as documented in their submission and reclaiming of groupIds docs.
https://plugins.gradle.org/docs/submit
https://plugins.gradle.org/docs/reclaiming

SrinivasanTarget: Then as per your findings, we cannot accept this change since our main source of repository is maven central not jcenter. :(

SrinivasanTarget: Also may i know where does ```shadow``` benefits users? What is your usecase behind it?

shauvik:
Heard a rant about having to download tons of jars while they were trying
to get appium java-client to run.
In their case, they were not using maven or gradle, and probably eclipse or
command line.

Selenium distributes a standalone version and also another version with a
zip tarball of all libs, which makes it super simple to get started.
shadow helps create such a fatJar easily. That's why i wanted to add it in
appium!

SrinivasanTarget:
@shauvik Ok Thanks

build.gradle
SrinivasanTarget: @shauvik Can you make necessary changes to point the repo back to maven central?

shauvik: Done. Had to add a jcenter dependency to buildscript, since shadow is only available there.

PR #562, opened 2017-01-29
add apis to read the performance data

heeseon:
## Change list

add apis to read the resource usage information of the application as like cpu, memory, network traffic, and battery
 
## Types of changes

## Details

The appium already has a way to gather the performance data by log.
but If the resulting data format is set to a log file, the user should conduct another parsing to show the data on the screen.
also I think user wants to decide the moment to read the data and associate the performance data and other information.
so I’d like to add apis to gather the performance data.

please refer below the commit code 
- https://github.com/appium/appium-base-driver/commit/0fd67e4efbed9c8a8352b18557b84b10649baabc
- https://github.com/appium/appium-android-driver/commit/1e6c644a9260ca22726405b44c5e4370b69f9059

### DIFF
--- a/src/main/java/io/appium/java_client/android/AndroidDriver.java
+++ b/src/main/java/io/appium/java_client/android/AndroidDriver.java
@@ -17,6 +17,8 @@
 package io.appium.java_client.android;
 
 import static io.appium.java_client.android.AndroidMobileCommandHelper.endTestCoverageCommand;
+import static io.appium.java_client.android.AndroidMobileCommandHelper.getPerformanceDataCommand;
+import static io.appium.java_client.android.AndroidMobileCommandHelper.getSupportedPerformanceDataTypesCommand;
 import static io.appium.java_client.android.AndroidMobileCommandHelper.openNotificationsCommand;
 import static io.appium.java_client.android.AndroidMobileCommandHelper.toggleLocationServicesCommand;
 
@@ -31,9 +33,11 @@ import io.appium.java_client.service.local.AppiumServiceBuilder;
 import org.openqa.selenium.Capabilities;
 import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.HttpCommandExecutor;
+import org.openqa.selenium.remote.Response;
 import org.openqa.selenium.remote.http.HttpClient;
 
 import java.net.URL;
+import java.util.List;
 
 /**
  * @param <T> the required type of class which implement {@link org.openqa.selenium.WebElement}.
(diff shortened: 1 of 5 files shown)


src/test/java/io/appium/java_client/android/AndroidDriverTest.java
```diff
@@ -140,4 +141,33 @@
...
+        assert(supportedPerformanceDataTypes.size() == 4);
+
+    }
+
+    @Test public void getPerformanceDataTest() {
+        driver.startActivity("io.appium.android.apis", ".ApiDemos");
+
+        List<String> supportedPerformanceDataTypes = driver.getSupportedPerformanceDataTypes();
+
+        for(int i = 0 ; i < supportedPerformanceDataTypes.size() ;  ++ i){
```
mykola-mokhnach: why not to use for .. in here?

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
```diff
@@ -140,4 +141,33 @@
...
+
+            String dataType = supportedPerformanceDataTypes.get(i);
+
+            List<List<Object>> valueTable = driver.getPerformanceData("com.example.android.apis", dataType, 60000);
+
+            int valueTableHeadLength = valueTable.get(0).size();
+
+            for(int j = 1 ; j < valueTable.size() ; ++ j){
```
mykola-mokhnach: same here. Use _subList_ property to get a slice

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
```diff
@@ -140,4 +141,33 @@
         assertNotNull(driver.getDisplayDensity());
         assertNotEquals(0, driver.getSystemBars().size());
     }
+
+    @Test public void getSupportedPerformanceDataTypesTest() {
+        driver.startActivity("io.appium.android.apis", ".ApiDemos");
+
+        List<String> supportedPerformanceDataTypes = driver.getSupportedPerformanceDataTypes();
+        assert(supportedPerformanceDataTypes.size() == 4);
```
mykola-mokhnach: it could be more readable to use assertEquals instead

heeseon: ok

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
heeseon: I am not sure about the difference between List.get(i) and List.subList(i,i).
but I applied it.

and one more, 
in case of this method, 
there is no reason to check the head value of the returned table, because the implementation is dependent for the platform that means the head value could be changed.

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
heeseon: i applied it. 
but I think there is no reason to check the value of the table, because the values could be changed in appium-android-driver. 
I have a plan to add some other values for example, heaphistogram....

src/main/java/io/appium/java_client/android/AndroidMobileCommandHelper.java
```diff
@@ -26,6 +26,7 @@
 import org.openqa.selenium.internal.HasIdentity;
 
 import java.util.AbstractMap;
+import java.util.List;
```
heeseon: applied

src/test/java/io/appium/java_client/android/AndroidDriverTest.java
```diff
@@ -140,4 +141,33 @@
         assertNotNull(driver.getDisplayDensity());
         assertNotEquals(0, driver.getSystemBars().size());
     }
+
+    @Test public void getSupportedPerformanceDataTypesTest() {
```
heeseon: ok I used assertEquals

TikhomirovSergey:
@heeseon 
It looks ok but I would like to make the same the following way

```java
public interface HasSupportedPerformanceDataType {
    default List<String> getSupportedPerformanceDataTypes() {
         return CommandExecutionHelper.execute(this, getSupportedPerformanceDataTypesCommand());
    }

    default List<List<Object>> getPerformanceData(String packageName, String dataType, int dataReadTimeout) {
        return CommandExecutionHelper.execute(this,
                                                getPerformanceDataCommand(packageName, dataType, dataReadTimeout));
    }
}
```

and then you just need to declare 

```java
public class AndroidDriver<T extends WebElement>
    extends AppiumDriver<T>
    implements PressesKeyCode, HasNetworkConnection, PushesFiles, StartsActivity,
        FindsByAndroidUIAutomator<T>, LocksAndroidDevice, HasSettings, HasDeviceDetails, HasSupportedPerformanceDataType {
...
}

```

heeseon:
fixed all

src/main/java/io/appium/java_client/android/AndroidDriver.java
```diff
@@ -153,6 +157,52 @@ public AndroidDriver(Capabilities desiredCapabilities) {
     }
 
     /**
+     * returns the information type of the system state which is supported to read
```
TikhomirovSergey: Hi @heeseon 
Everything is ok but... could you remove the code below and move java doc to the `HasSupportedPerformanceDataType`? It is not necessary to duplicate code of API implemented by default. I think there is no specific details of the AndroidDriver behaviour.

SrinivasanTarget: @heeseon ping

heeseon: Hi, sorry..
I checked this message now. I will apply this tonight

heeseon:
I deleted the duplicated code. Please check it.

src/main/java/io/appium/java_client/android/HasSupportedPerformanceDataType.java
```diff
@@ -0,0 +1,61 @@
...
+import static io.appium.java_client.android.AndroidMobileCommandHelper.getSupportedPerformanceDataTypesCommand;
+
+import io.appium.java_client.CommandExecutionHelper;
+import io.appium.java_client.ExecutesMethod;
+
+import java.util.List;
+
+/**
+ * Created by hwangheeseon on 2017. 2. 2..
+ */
```
SrinivasanTarget: @heeseon Can you remove these statements please?

heeseon:
Ok

TikhomirovSergey:
I have approved these changes. @SrinivasanTarget ?

SrinivasanTarget:
Thanks @heeseon

Issue #567, opened 2017-02-06
5.0.0-BETA2: ClassCastException: WebElement -> IOSElement


amedvedjev:
## Description
Failed to cast WebElement into IOSElement. Worked with 5.0.0-BETA1.

## Environment

* java client build version or git revision if you use some shapshot: 5.0.0-BETA2
* Appium server version or git revision if you use some shapshot: 1.6.4-beta
* Mobile platform/version under test: iOS
* Real device or emulator/simulator: Simulator

## Code To Reproduce Issue [ Good To Have ]
`((IOSElement) driver.findElement(MobileBy.className("XCUIElementTypePickerWheel"))).setValue(countryName);`

## Ecxeption stacktraces
java.lang.ClassCastException: org.openqa.selenium.remote.RemoteWebElement cannot be cast to io.appium.java_client.ios.IOSElement

TikhomirovSergey:
ok.

TikhomirovSergey:
@amedvedjev 
Please provide capabilities which you pass througt the driver constructor.
Also could you show the map that the method 'driver.getSessionDetails()' returns at your case?

amedvedjev:
@TikhomirovSergey 
here is (removed some sensitive names)

**Appium start:**
`start appium as: [appium, --log-level, error, --port, 4727, --bootstrap-port, 5727, --command-timeout, 90, --session-override]`

**Driver capabilities:** (i know some deprecated...)
`Capabilities [{app=/Users/jenkinss/Downloads/test_clients/xxxxx.app, preventWDAAttachments=false, noReset=false, clearSystemFiles=true, scaleFactor=0.5, version=, deviceName=iPhone 6s, iosInstallPause=200, fullReset=false, platform=MAC, wdaLaunchTimeout=60000, newCommandTimeout=120, platformVersion=10.2, automationName=XCuiTest, useNewWDA=true, browserName=, waitForAppScript=$.delay(3000); $.acceptAlert(); true;, wdaConnectionTimeout=60000}]`

**getSessionDetails:**
[...]

PR #568, opened 2017-02-10
#567 fix

TikhomirovSergey:
## Change list

#567 fix
 
## Types of changes

## Details

- 'isBrowserMethod' is not implemented by default;
- element generaration was reverted to the old rule which was used before 5.0.0-BETA1


TikhomirovSergey:
@amedvedjev @derunewige
You are invided to review/try this PR too.
If everything is ok then it will be published immideately at the BETA3.

PS. Appium field decorator was changed a little bit. It is not the breaking change... bit it is starting of the refactoring. It is related to the feature request opened by @amedvedjev. It is supposed to be finished at the BETA4.

amedvedjev:
@TikhomirovSergey can you give advice how i can try this pull with mine maven project easily?

PR #578, opened 2017-02-22
Simplified the StartsActivity by reducing the number of parameters through POJO class

email2vimalraj:
Simplified the StartsActivity by reducing the number of parameters through POJO class

## Change list

Currently the number of `startActivity` methods are unnecessary which can be solved by one single method with single parameter. So introduced the simple POJO class which in future can be expanded with any number of parameters.
 
## Types of changes

## Details

The sample usage of new implementation is as follows:

```java
Activity activity = new Activity();
activity.setAppPackage("com.foo");
activity.setAppActivity(".bar");
driver.startActivity(activity);
```

### DIFF
--- a/docs/The-starting-of-an-Android-app.md
+++ b/docs/The-starting-of-an-Android-app.md
@@ -103,6 +103,7 @@ use Android-specific API eventually.
  
  ```java
  import io.appium.java_client.android.StartsActivity;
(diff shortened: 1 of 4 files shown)


email2vimalraj:
Closing my PR since I have some more failures. I'll raise again once I fixed everything.

PR #595, opened 2017-03-09
Add possibility to calculate screenshots overlap score + several helpers

mykola-mokhnach:
## Change list

This feature has been asked by several users, so I've just nicely decorated the code we use in our own automation framework and added it to the client.
 
## Types of changes

- [ ] No changes in production code.
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

Screenshots comparison algorithm itself is based on standard opencv calls. That is why the library has been added to Gradle config as a dependency.

PR #611, opened 2017-03-30
Use java.time.Duration to represent time deltas instead of elementary types

mykola-mokhnach:
## Change list

Replace timeout/duration arguments where possible with _java.time.Duration_ type, which has been introduced in Java 8.. This gives us better control over time conversion and helps to avoid time unit conversion issues in general.

All the existing methods, that have replacements, were marked as deprecated.
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/InteractsWithApps.java
+++ b/src/main/java/io/appium/java_client/InteractsWithApps.java
@@ -27,6 +27,7 @@ import static io.appium.java_client.MobileCommand.prepareArguments;
 
 import com.google.common.collect.ImmutableMap;
 
+import java.time.Duration;
 import java.util.AbstractMap;
 
 public interface InteractsWithApps extends ExecutesMethod {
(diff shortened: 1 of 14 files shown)


src/main/java/io/appium/java_client/internal/DurationUtils.java
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client.internal;
+
+import java.time.Duration;
+
+public class DurationUtils {
+    public static int durationToMillis(Duration duration) {
```
TikhomirovSergey: @mykola-mokhnach 
Maybe it is more senseful to use this method instead of the util

```java
Duration duration;
....
duration.toMillis();
```

TikhomirovSergey:
@mykola-mokhnach Hi. Sorry for the delaying. I am usually busy now. 
These changes are cool 👍  But there is the question above. Cold you answer or make some improvements?

@SrinivasanTarget What is your opinion?

src/main/java/io/appium/java_client/internal/DurationUtils.java
mykola-mokhnach: fixed. I dunno why I skipped the fact, that Duration class already has such method %)

mykola-mokhnach:
@TikhomirovSergey np, thanks for the review. I'd be happy if someone could also add comments to the other one at https://github.com/appium/java-client/pull/595

TikhomirovSergey:
@mykola-mokhnach Thank you very much!

PR #652, opened 2017-06-03
Deprecated youiengine

TikhomirovSergey:
## Change list

- all the content of the `io.appium.java_client.youiengine` package was marked as `Deprecated`
- update of the `com.google.code.gson:gson` to v2.8.1
 
## Types of changes

## Details

This package was added at #429. It was accepted due to previous implementation was not flexible and it did not allowed user to use their own commands/JSONWP protocol extensions. Now `AppiumDriver` is not abstract and it can be used across supported platfoms. 

Please take a look at the following sample: 

```java
...
import io.appium.java_client.AppiumDriver;
import io.appium.java_client.android.Activity;
import io.appium.java_client.android.AndroidElement;
...

    private static AppiumDriver<AndroidElement> driver;
    private StartsActivity startsActivity;

        ...
        DesiredCapabilities capabilities = new DesiredCapabilities();
        ...
        driver = new AppiumDriver<>(...);

        ...
        startsActivity = new StartsActivity() {
            @Override
            public Response execute(String driverCommand, Map<String, ?> parameters) {
                return driver.execute(driverCommand, parameters);
            }

            @Override
            public Response execute(String driverCommand) {
                return driver.execute(driverCommand);
            }
        };
        Activity activity = new Activity("io.appium.android.apis", ".ApiDemos");
        startsActivity.startActivity(activity);
```

Also now it allows end user to extend command repository the following way:

```java
MobileCommand.commandRepository.put("someCommand", new AppiumCommandInfo("someUrl", HttpMethod.POST));
```
Also I was motivated to remove this package because there is no activity/development.

### DIFF
--- a/src/main/java/io/appium/java_client/youiengine/YouiEngineDriver.java
+++ b/src/main/java/io/appium/java_client/youiengine/YouiEngineDriver.java
@@ -23,11 +23,11 @@ import org.openqa.selenium.WebElement;
 
 import java.net.URL;
 
-@Deprecated
 /**
  * This class is going to be removed.
  * Please use {@link AppiumDriver} instead
  */
+@Deprecated
 public class YouiEngineDriver<T extends WebElement> extends AppiumDriver<T> {
 
     /** Constructor takes in the Appium Server URL and the capabilities you want to use for this
--- a/src/main/java/io/appium/java_client/youiengine/YouiEngineDriver.java
+++ b/src/main/java/io/appium/java_client/youiengine/YouiEngineDriver.java
@@ -23,6 +23,11 @@ import org.openqa.selenium.WebElement;
 
 import java.net.URL;
 
+@Deprecated
+/**
+ * This class is going to be removed.
+ * Please use {@link AppiumDriver} instead
+ */
 public class YouiEngineDriver<T extends WebElement> extends AppiumDriver<T> {
 
     /** Constructor takes in the Appium Server URL and the capabilities you want to use for this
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
@SrinivasanTarget @mykola-mokhnach Could you review it.

FYI @pfoster-youitv @philgreenyouilabs @jlipps @simongranger

src/main/java/io/appium/java_client/youiengine/YouiEngineDriver.java
```diff
@@ -23,6 +23,11 @@
 
 import java.net.URL;
 
+@Deprecated
```
mykola-mokhnach: I suppose the common practice is to put a docsting before an annotation, but not after

TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget I have updated it a little bit and now I am trying to notify youiengine guys.

SrinivasanTarget:
LGTM @TikhomirovSergey

mykola-mokhnach:
@TikhomirovSergey I think the best place to find them is **general** channel of Appium's slack. We could add your account if you are still not there.

pfoster-youitv:
> Now AppiumDriver is not abstract and it can be used across supported platfoms.

That was the main thing we needed and not be strictly tied to either iOS or Android elements/commands since we do not use Android or iOS native elements. 👍

TikhomirovSergey:
@pfoster-youitv 
Ok. So we are waiting for the `youiengine ` decision.

PR #658, opened 2017-06-16
enable nativeWebTap setting for iOS

jlipps:
## Change list

add a setting for iOS to toggle `nativeWebTap`
 
## Types of changes

## Details

Server-side change: https://github.com/appium/appium-xcuitest-driver/pull/455

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumSetting.java
+++ b/src/main/java/io/appium/java_client/AppiumSetting.java
@@ -19,7 +19,7 @@ package io.appium.java_client;
 
 /**
  * This enum is deprecated. Was moved to
- * {@link io.appium.java_client.android.Setting}.
+ * {@link io.appium.java_client.Setting}.
  */
 @Deprecated
 public enum AppiumSetting {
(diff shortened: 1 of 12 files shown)


SrinivasanTarget:
@jlipps Looks good overall. Can you add some tests please?

jlipps:
@SrinivasanTarget added tests, and ran it locally with success. note that this will probably fail CI since it is set to iOS 10.2 and requires appium 1.6.6-beta.2 with a fresh download. not sure what the appium local service runs in the java client... but anyway it works!

jlipps:
found the javadoc error that was failing travis. hopefully it will pass now.

SrinivasanTarget:
> probably fail CI since it is set to iOS 10.2 and requires appium 1.6.6-beta.2 with a fresh download

We have excluded running tests in CI, so it wouldn't be a problem now. Also this is in pipeline for us.

>not sure what the appium local service runs in the java client

Appium local service runs appium as a service automatically at stated port,etc  without being started manually.

jlipps:
Great! Now how do we publish a beta so someone can use it? :-)

PR #663, opened 2017-06-26
Clear existing actions/parameters after perform is invoked

mykola-mokhnach:
## Change list

Clear existing actions/parameters after _perform_ method of the corresponding actions chain is invoked
 
## Types of changes

## Details

I find the current behaviour confusing, since it may cause issues like this one: https://github.com/appium/appium/issues/8734

The second possible solution might be to make a chain immutable after _perform_ is called, so it will reject all the further parameter/action adding operations after _perform_ method has been invoked once on a chain instance.


TikhomirovSergey:
@mykola-mokhnach Hi! I will review it ASAP. Thank you :)

Issue #655, opened 2017-06-12
Java-client will throw exceptions - GridException: Cannot extract a capabilities xxx with webdriver 3.x


fionazj:
## Description

Before the session of the simulator is started, some errors will show up in hub log as follow.
## Environment

* WebDriver version build version : webdriver 3.3.1 (Also tried with webdriver 3.3.4)
* java client build version: java-client-5.0.0-BETA7
* Appium : 1.6.5
* Desktop OS/version used to run Appium if necessary: Mac OS 10.12.5
* Node.js version : 4.2.0
* Mobile platform/version under test:  IPhone  IOS 9.3
* Real device or emulator/simulator: Simulator

## Details

Since webdriver3.x is released, so we need to do a trail run on that, so I updated the webdriver in both our project and hub, also java-client in the project, and start the hub to execute cases. But when a session start, exception occurs, and seemed like the Appium is fine, but java-client 5.x with webdriver3.x is not, did i miss something?

And it was totally fine when I change back webdriver to 2.53.0

## Code To Reproduce Issue [ Good To Have ]

Hub startup command line:

> /usr/bin/java -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=5005 -jar -verbose /root/selenium-grid/selenium-server-standalone-3.3.1.jar -role hub -port 44442 >> bjdev3.log &

Node configuration of Appium
> admins-Mac-mini:Desktop uia$ cat appium_node_config_ios.json
> {
>   "capabilities":
>       [
>         {
>           "browserName": "Safari",
>           "platformVersion":"9.3",
>           "maxInstances": 1,
>           "platformName":"iOS",
>           "deviceName":"iPhone 6s",
>           "seleniumProtocol":"WebDriver"
>         }
>       ],
> 
> 
>   "configuration":
>   {
>     "cleanUpCycle":2000,
>     "timeout":30000,
>     "proxy": "org.openqa.grid.selenium.proxy.DefaultRemoteProxy",
>     "url":"http://10.37.2.203:4723/wd/hub",
>     "maxSession": 1,
>     "port": 4723,
>     "host": "10.37.2.203",
>     "register": true,
>     "registerCycle": 5000,
>     "hubPort": 44442,
>     "hubHost": "10.37.10.37"
>   }
> }

## Ecxeption stacktraces

[...]

PR #597, opened 2017-03-10
#594 FIX

TikhomirovSergey:
## Change list

#594 FIX:
- update to selenium 3.3.1
- AppiumCommandExecutor was reimplemented. Now it implements CommandExecutor and doesn't extends HttpCommandExecutor. However the most part of implementation was taken from Selenium.
- constructor changes.
 
## Types of changes
_Put an `x` in the boxes that apply_

- [ ] No changes in production code.
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details
Some changes may be considered as breaking changes.

Issue #594, opened 2017-03-08
5.0.0-BETA5


TikhomirovSergey:
## Description
The failing on the session creation.

## Environment

* java client build version or git revision if you use some shapshot: 5.0.0-BETA5
* Appium server version or git revision if you use some shapshot: 1.6.3
* Desktop OS/version used to run Appium if necessary: Windows 8
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: 7.0.0
* Mobile platform/version under test: Android
* Real device or emulator/simulator: Emulator

## Details

When I published the 5.0.0-BETA5 I had run tests and everything was ok. Yesterday I cleared caches and tried to check something. Instead I caught the exception. It reproduces with Selenium 3.2.0 and 3.3.0.

## Code To Reproduce Issue [ Good To Have ]

```java
        service = AppiumDriverLocalService.buildDefaultService();
        service.start();

        if (service == null || !service.isRunning()) {
            throw new AppiumServerHasNotBeenStartedLocallyException(
                "An appium server node is not started!");
        }

        File appDir = new File("src/test/java/io/appium/java_client");
        File app = new File(appDir, "ApiDemos-debug.apk");
        DesiredCapabilities capabilities = new DesiredCapabilities();
        capabilities.setCapability(MobileCapabilityType.DEVICE_NAME, "Android Emulator");
        capabilities.setCapability(MobileCapabilityType.APP, app.getAbsolutePath());
        driver = new AndroidDriver<>(service.getUrl(), capabilities);
```

## Ecxeption stacktraces

```
Unable to create new remote session. desired capabilities = Capabilities [{app=C:\appium\java-client\src\test\java\io\appium\java_client\ApiDemos-debug.apk, platformName=Android, deviceName=Android Emulator}], required capabilities = Capabilities [{}]
Build info: version: '3.3.0', revision: 'b526bd5b41', time: '2017-03-07 19:26:04 +0000'
System info: host: 'HP', ip: '192.168.76.1', os.name: 'Windows 8.1', os.arch: 'amd64', os.version: '6.3', java.version: '1.8.0_73'
Driver info: driver.version: AndroidDriver
org.openqa.selenium.SessionNotCreatedException: Unable to create new remote session. desired capabilities = Capabilities [{app=xxx\ApiDemos-debug.apk, platformName=Android, deviceName=Android Emulator}], required capabilities = Capabilities [{}]
[...]

PR #682, opened 2017-07-31
#655 FIX

TikhomirovSergey:
## Change list

#655 FIX
 
## Types of changes

## Details

`io.appium.java_client.remote.AppiumCommandExecutor` was rolled back.

It is breaking change because it is compatible with appium server > v1.6.5


TikhomirovSergey:
#594 #597

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -47,124 +35,44 @@
...
-    public AppiumCommandExecutor(Map<String, AppiumCommandInfo> additionalCommands,
-        URL addressOfRemoteServer, HttpClient.Factory httpClientFactory) {
-        checkNotNull(addressOfRemoteServer);
-        remoteServer = addressOfRemoteServer;
-        this.additionalCommands = additionalCommands;
-        this.client = httpClientFactory.createClient(remoteServer);
+    public AppiumCommandExecutor(Map<String, CommandInfo> additionalCommands,
+                                 URL addressOfRemoteServer, HttpClient.Factory httpClientFactory) {
+        super(additionalCommands, addressOfRemoteServer, httpClientFactory);
+        service = null;
```
mykola-mokhnach: can this be a reason of unexpected NPE?

mykola-mokhnach:
How about having some unit test to verify resulting capabilities structure?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
TikhomirovSergey: No. It is checked inside this method
https://github.com/TikhomirovSergey/java-client/blob/12cb9d9b6e117274340e8c73f40b0fa7a11396ba/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java#L64

mykola-mokhnach: then it might be useful to check the property as nullable

mykola-mokhnach: or make it optional (I'd prefer this option)

TikhomirovSergey:
@mykola-mokhnach 
>How about having some unit test to verify resulting capabilities structure?

I think we don't need for these tests:
1) as you can see `public class AppiumCommandExecutor extends HttpCommandExecutor` and

```java
 @Override public Response execute(Command command) throws IOException, WebDriverException {
        ...
      return super.execute(command);
       ...
   }
```

2) The capability forming is fully inherited from Selenium
[HttpCommandExecutor](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/HttpCommandExecutor.java#L124) [ProtocolHandshake](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/remote/ProtocolHandshake.java)

Our tests which run some AppiumDriver are successful.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
TikhomirovSergey: @mykola-mokhnach ok. I will improve it soon.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -47,131 +37,56 @@
...
-
-            Response response = responseCodec.decode(httpResponse);
-            if (response.getSessionId() == null) {
-                if (httpResponse.getTargetHost() != null) {
-                    response.setSessionId(HttpSessionId.getSessionId(httpResponse.getTargetHost()));
-                } else {
-                    response.setSessionId(command.getSessionId().toString());
+    @Override public Response execute(Command command) throws WebDriverException {
+        if (DriverCommand.NEW_SESSION.equals(command.getName())) {
+            ofNullable(service).ifPresent(driverService -> {
```
mykola-mokhnach: I assume the code will be simpler if we make the service variable itself of type Optional

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -47,131 +37,56 @@
...
-                    throw new WebDriverException("The appium server has accidentally died!", t);
-                }
+                    && rootCause.getMessage().contains("Connection refused")) {
+                throw ofNullable(service).map((Function<DriverService, WebDriverException>) service -> {
+                    if (service.isRunning()) {
+                        return new WebDriverException("The session is closed!", rootCause);
+                    }
+
+                    return new WebDriverException("The appium server has accidentally died!", rootCause);
+                }).orElse(new WebDriverException(rootCause.getMessage(), rootCause));
```
mykola-mokhnach: probably, orElseGet ?

TikhomirovSergey:
@mykola-mokhnach There is another one improvement.

mykola-mokhnach:
@TikhomirovSergey 👍

PR #698, opened 2017-08-19
Address compilation error with selenium 3.5.1

mykola-mokhnach:
## Change list

Fix compilation error with selenium 3.5.1 module (see https://travis-ci.org/appium/java-client/builds/266018745)
 
## Types of changes

- [ ] No changes in production code.
- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

Run several functional tests and the stuff looks good so far. However, it would be good to run more extended suite.

Issue #696, opened 2017-08-17
Warning: Selenium 3.5.1 breaks java client 5.0.0-BETA9


welnanick:
I am using gradle to handle dependencies for my project. In my `build.gradle` I have `testCompile "io.appium:java-client:5.0.0-BETA9"` to compile the appium java client. It seems Selenium 3.5.1 was just released (their website has not been updated as I am writing this), and for some reason gradle is grabbing version 3.5.1 automatically for selenium-api, selenium-remote-driver, and selenium-support. I don't know why this is happening, but it is and it breaks the java client. I had to add
```
compile ('org.seleniumhq.selenium:selenium-api:3.4.0'){force = true}
compile ('org.seleniumhq.selenium:selenium-remote-driver:3.4.0'){force = true}
compile ('org.seleniumhq.selenium:selenium-support:3.4.0'){force = true}
```
to fix my project.

So, if your project just suddenly stopped working for no reason whatsoever, check if these packages were updated, and if so make sure to force version 3.4.0 for them.

jason-edstrom:
I just solved this exact issue on my setup. Same resolution approach, forcing 3.4.0.

QAutomatron:
Faced the same problem. Thank you for a solution. 

I tried a different one, it also helps:
```
configurations.all {
    resolutionStrategy {
        force 'org.seleniumhq.selenium:selenium-support:3.4.0',
                'org.seleniumhq.selenium:selenium-api:3.4.0'
    }
}
```

SrinivasanTarget:
@QAutomatron @welnanick @jason-edstrom Whats the actual problem faced here? What breaks with 3.5.1? Gonna hands-on anyways tomorrow.

SrinivasanTarget:
Changes are here: https://github.com/SeleniumHQ/selenium/commit/323132d747fc5a64e57efd34f392f752073da503

welnanick:
It seems annotations (or possible all FindBy operations given the changes in selenium) break. When testing android, elements populated using `@AndroidFindBy` stay null, and when testing iOS elements populated using `@iOSFindBy` cause `org.openqa.selenium.WebDriverException`

Unfortunately I am not in a position to investigate further, nor confirm whether it is just presenting itself with gradle or not.

SrinivasanTarget:
Cool thanks. will investigate further.

SrinivasanTarget:
@TikhomirovSergey ping

Brian-Watson:
[...]

PR #699, opened 2017-08-20
	#696 FIX

TikhomirovSergey:
## Change list

- changes from the #698
 - selenium-java was updated to 3.5.x
 - org.apache.commons-lang3 was updated to 3.6
 - org.springframework.spring-context was updated to 4.3.10.RELEASE
 - the element genering was improved. #696 FIX
 - some improvements of AppiumDriver#context
 
## Types of changes

## Details

I suppose that it is the last change and we should publish 5.0.0. All the pended pull-requests will be merged and published at 5.1.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'java'
+apply plugin: 'idea'
 apply plugin: 'maven'
 apply plugin: 'eclipse'
 apply plugin: 'jacoco'
--- a/build.gradle
+++ b/build.gradle
@@ -55,7 +55,7 @@ compileJava {
 }
 
 dependencies {
-    compile('org.seleniumhq.selenium:selenium-java:3.4.0'){
+    compile('org.seleniumhq.selenium:selenium-java:3.5.1'){
         exclude module: 'cglib'
         exclude group: 'com.google.code.gson'
     }
@@ -63,9 +63,9 @@ dependencies {
     compile 'org.apache.httpcomponents:httpclient:4.5.3'
     compile 'cglib:cglib:3.2.5'
     compile 'commons-validator:commons-validator:1.6'
-    compile 'org.apache.commons:commons-lang3:3.5'
+    compile 'org.apache.commons:commons-lang3:3.6'
     compile 'commons-io:commons-io:2.5'
-    compile 'org.springframework:spring-context:4.3.8.RELEASE'
+    compile 'org.springframework:spring-context:4.3.10.RELEASE'
     compile 'org.aspectj:aspectjweaver:1.8.10'
     compile 'org.openpnp:opencv:3.2.0-1'
 
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'signing'
 apply plugin: 'maven-publish'
 
 group 'io.appium'
-version '5.0.0-BETA9'
+version '5.0.0'
 
 repositories {
     jcenter()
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -54,17 +55,17 @@ compileJava {
 }
 
 dependencies {
-    compile('org.seleniumhq.selenium:selenium-java:3.4.0'){
+    compile('org.seleniumhq.selenium:selenium-java:3.5.1'){
```
mykola-mokhnach: it would be nice to know whether it is possible to limit the update of major component version, so it would be possible to avoid similar issues in the future. I'm thinking about the same logic, like in package.json for Node JS modules, where one can say "bla: ^1.0" and this will block the component "bla" from major version update (however, minor version updates are allowed)

mykola-mokhnach:
> [ant:checkstyle] [ERROR] /home/travis/build/appium/java-client/src/main/java/io/appium/java_client/AppiumDriver.java:215:9: '}' at column 9 should be on the same line as the next part of a multi-block statement (one that directly contains multiple blocks: if/else-if/else, do/while or try/catch/finally). [RightCurly]
:checkstyleMain FAILED

build.gradle
TikhomirovSergey: @mykola-mokhnach Ok I will google how the similar problems are resolved.

TikhomirovSergey:
@mykola-mokhnach 
Yes I know about the check style issue and I will get it fixed ASAP

build.gradle
mykola-mokhnach: Seems like this might be a good match to our purposes: https://stackoverflow.com/questions/39664589/use-semantic-versioning-for-a-dependency-in-gradle

SrinivasanTarget: @TikhomirovSergey @mykola-mokhnach i think it only restricts `selenium-java` dependency and not its sub modules or transitive dependency.

SrinivasanTarget: @TikhomirovSergey We are good to update to `3.5.2` as `3.5.+`. I don't see any issues on yesterday's selenium release.

mykola-mokhnach: Probably, we could also define selenium dependency with `forced` flag, to make sure it is always used for our lib, even if Selenium component version is updated for the main project?

https://docs.gradle.org/3.3/dsl/org.gradle.api.artifacts.dsl.DependencyHandler.html

SrinivasanTarget: yah makes sense. Will try few things locally and update here.

mykola-mokhnach: Yeah, something like this:

```
    compile('org.seleniumhq.selenium:selenium-java:3.5.+'){
        force = true

        exclude module: 'cglib'
        exclude group: 'com.google.code.gson'
    }
```

SrinivasanTarget: Yeah just tried this like below but it still downloads latest sub modules of it.
```
compile('org.seleniumhq.selenium:selenium-java:3.4.+'){
        exclude module: 'cglib'
        exclude group: 'com.google.code.gson'
        force = true
    }
```
It still downloads latest `selenium-api` i.e 3.5.2

SrinivasanTarget: How about this @mykola-mokhnach ?
```
compile('org.seleniumhq.selenium:selenium-java:3.4.+'){
        exclude module: 'cglib'
        exclude group: 'com.google.code.gson'
        force = true
        transitive = false
}
compile('org.seleniumhq.selenium:selenium-api:3.4.+')
compile('org.seleniumhq.selenium:selenium-support:3.4.+')
compile('org.seleniumhq.selenium:selenium-remote-driver:3.4.+')
```

mykola-mokhnach: How about this solution?
seleenium

mykola-mokhnach: I've almost created the same. Although having Selenium version assigned to a single variable looks more elegant to me ;)

mykola-mokhnach: or even shorter:

```
ext.seleniumVersion = '3.4.+'

dependencies {
    compile ("org.seleniumhq.selenium:selenium-java:${seleniumVersion}") {
        force = true

        exclude module: 'cglib'
        exclude group: 'com.google.code.gson'
    }
    compile ("org.seleniumhq.selenium:selenium-support:${seleniumVersion}") {
        force = true
    }
    compile ("org.seleniumhq.selenium:selenium-api:${seleniumVersion}") {
        force = true
    }
```

SrinivasanTarget: Awesome Looks good 👍

TikhomirovSergey: @mykola-mokhnach @SrinivasanTarget Hi guys. Thank you for the discussion and hints. Today I am going to update the PR

SrinivasanTarget: Sure @TikhomirovSergey

TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget I have updated this PR

mykola-mokhnach:
Yep, now the PR looks much better. @TikhomirovSergey please merge it

Issue #714, opened 2017-08-30
The old Selenium dependency makes user's code throw java.lang.NoSuchMethodError


TikhomirovSergey:
## Description

There are many issue reports where users complain about java.lang.NoSuchMethodError and related dependency conflicts.

## Environment

Any

## Details
https://gist.github.com/TikhomirovSergey/f008a577cac9b6c92fb1def9a2aa5406

It seems I found the root cause. I is probably because of html-unit driver which is supplied by Selenium. 
The possible solution:
- exclude this artefact

SrinivasanTarget:
@TikhomirovSergey can we exclude all the browser specific drivers?

TikhomirovSergey:
@SrinivasanTarget I think only htmlUnit. Also could you check dependencies of phantomJS.

SrinivasanTarget:
Sure @TikhomirovSergey

EitanBe:
Can you please tell when you going to merge the fix, or if there is a workaround until the fix would be published. (Which version)
Thanks a lot guys!

TikhomirovSergey:
@EitanBe We are going to merge it and publish 5.0.2 on this weekend (02.09.2017-03.09.2017).  As a workaround you can add exclusions of htmlunit-driver and phantomjsDriver

SrinivasanTarget:
@TikhomirovSergey I'm not able to reproduce this locally. I don't see any issues including html-unit driver. Everything works fine for me.

SrinivasanTarget:
@EitanBe Can you share me your pom.xml or build.gradle inform of gist here?

TikhomirovSergey:
@SrinivasanTarget It is difficult to reproduce it inside in sources. It may be reproduced when java_client is used as a dependency. Please take a look at this issue report #709. Please read the dependency tree attached to the report and take a look at dependencies.

SrinivasanTarget:
Yeah @TikhomirovSergey I was trying to reproduce it as external dependency but still not able too.

EitanBe:
Hi here list of my dependencies:
<dependencies>
        

        <dependency>
            <groupId>info.cukes</groupId>
            <artifactId>cucumber-java</artifactId>
            <version>1.2.5</version>
        </dependency>
        <dependency>
            <groupId>info.cukes</groupId>
            <artifactId>cucumber-java8</artifactId>
            <version>1.2.5</version>
        </dependency>
        <dependency>
            <groupId>org.springframework</groupId>
[...]

PR #717, opened 2017-09-01
#714 FIX

TikhomirovSergey:
## Change list

- Update to Selenium 3.5.3
- phantomJSdriver and htmlunit-drive were excluded
- eclipse compiler was updated to 4.6.1
- also AppiumElementLocator was improved

 
## Types of changes

## Details

Dependencies of the html unit driver
https://mvnrepository.com/artifact/org.seleniumhq.selenium/htmlunit-driver/2.27

Dependencies of the phantom js driver
https://mvnrepository.com/artifact/com.codeborne/phantomjsdriver/1.4.3

Also something weird was foaund in AppiumElementLocator. I wonder why we couldn't find it before.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'signing'
 apply plugin: 'maven-publish'
 
 group 'io.appium'
-version '5.0.1'
+version '5.0.2'
 
 repositories {
     jcenter()
@@ -38,7 +38,7 @@ configurations {
 }
 
 dependencies {
-    ecj 'org.eclipse.jdt.core.compiler:ecj:4.5.1'
+    ecj 'org.eclipse.jdt.core.compiler:ecj:4.6.1'
 }
 
 compileJava {
@@ -54,7 +54,7 @@ compileJava {
     ]
 }
 
-ext.seleniumVersion = '[3.5.2,3.5.2]'
+ext.seleniumVersion = '[3.5.3,3.5.3]'
 
 dependencies {
     compile ("org.seleniumhq.selenium:selenium-java:${seleniumVersion}") {
@@ -62,6 +62,10 @@ dependencies {
 
         exclude module: 'cglib'
         exclude group: 'com.google.code.gson'
+        exclude module: 'phantomjsdriver'
+        exclude module: 'htmlunit-driver'
+        exclude group: 'net.sourceforge.htmlunit'
+
     }
     compile ("org.seleniumhq.selenium:selenium-support:${seleniumVersion}") {
         force = true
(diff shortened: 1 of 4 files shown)


src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -21,8 +21,6 @@
 import static io.appium.java_client.pagefactory.ThrowableUtil.isStaleElementReferenceException;
 
 
-import com.google.common.base.Function;
```
mykola-mokhnach: I also see this import in several other sources - are they all expected?

build.gradle
```diff
@@ -54,14 +54,18 @@ compileJava {
     ]
 }
 
-ext.seleniumVersion = '[3.5.2,3.5.2]'
+ext.seleniumVersion = '[3.5.3,3.5.3]'
```
mykola-mokhnach: what is the point to set a range, which includes only one item? Isn't it simply equal to '3.5.3'?

mykola-mokhnach:
I was also thinking about enforcing the version of guava library, since our module cannot work if it is older than 20

build.gradle
TikhomirovSergey: @mykola-mokhnach The purpose is to make a valid pom.xml file. The version like 3.5+ is malformed for maven.

SrinivasanTarget:
@TikhomirovSergey As @mykola-mokhnach suggested i think we should force `com.google.guava:guava:23.0` dependency.

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
TikhomirovSergey: @mykola-mokhnach No
Except the `AppiumFunction` class.

TikhomirovSergey:
@SrinivasanTarget I will try it with the `com.google.guava:guava:23.0` dependency.

src/test/java/io/appium/java_client/android/OpenNotificationsTest.java
```diff
@@ -17,18 +16,15 @@ public void openNotification() throws Exception {
...
+        assertNotEquals(0, wait.until(input -> {
+            List<AndroidElement> result = input
+                    .findElements(id("com.android.systemui:id/carrier_label"));
 
-                if (result.size() == 0) {
-                    return null;
-                }
-
-                return result;
+            if (result.size() == 0) {
```
mykola-mokhnach: return result.isEmpty() ? null : result;

TikhomirovSergey:
@SrinivasanTarget I think it is not necessary to force the `com.google.guava:guava:23.0` because Selenium already depends on this version.

https://mvnrepository.com/artifact/org.seleniumhq.selenium/selenium-remote-driver/3.5.3

SrinivasanTarget:
> I think it is not necessary to force the com.google.guava:guava:23.0 because Selenium already depends on this version.

Ok

Issue #720, opened 2017-09-07
Appium 1.7.0 will support pushFile for iOS Simulators


imurchie:
## Description

Appium 1.7.0 will support pushing individual files to iOS simulators through `simctl addmedia`. The interface is the same as Android: a filename (for iOS, this is the name of a temporary local file that will be pushed) and the base 64 data of the contents. 

It will only work for pushing photos.

The implementation on the server side: https://github.com/appium/appium-xcuitest-driver/commit/430005e7ffec00dc03223efc22575e3984356f9b

mykola-mokhnach:
I'll take care of it

PR #721, opened 2017-09-07
Add pushFile support to IOSDriver

mykola-mokhnach:
## Change list

The PR adds support of pushFile endpoint to IOSDriver. For now it is only possible to upload media file to Simulator.
 
## Types of changes

## Details

Addresses https://github.com/appium/java-client/issues/720


mykola-mokhnach:
For iOS the documentation differs, wince the call works for Simulator only, but the body is the same. Frankly speaking, I don't see a need to extract these methods into a common lib - they anyway contain 1 to 3 lines of code. Even the documentation takes more space

TikhomirovSergey:
@mykola-mokhnach ok. So lets wait the result from CI.

mykola-mokhnach:
@KazuCocoa FYI

Issue #718, opened 2017-09-03
Selenium dependency boundaries not working with Maven


Hi-Fi:
## Description

While trying to create jar with dependencies with Selenium2library-java (https://github.com/Hi-Fi/robotframework-selenium2library-java/tree/Selenium3.5.2) with maven-assembly-plugin:3.1.0, error is thrown because of identical boundaries in Selenium version.

## Environment

* io.appium:java-client:jar:5.0.1
* Microsoft Windows [Version 10.0.15063]
* Maven version: 3.5.0
* Java version: 1.8.0_144, vendor: Oracle Corporation

## Details

Error message from Maven:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.1.0:single (make-assembly) on project robotframework-selenium2library-java: Failed to create assembly: Unable to resolve dependencies for assembly 'jar-with-dependencies': Failed to resolve dependencies for assembly: Unable to get dependency information for io.appium:java-client:jar:5.0.1: Invalid version for dependency org.seleniumhq.selenium:selenium-java:jar: Range cannot have identical boundaries: [3.5.2,3.5.2]
[ERROR]   io.appium:java-client:jar:5.0.1
```

## Way To Reproduce Issue

Can be seen by executing _mvn clean verify_ with code from https://github.com/Hi-Fi/robotframework-selenium2library-java/tree/Selenium3.5.2

mykola-mokhnach:
@TikhomirovSergey FYI

TikhomirovSergey:
@Hi-Fi @mykola-mokhnach Yes. I am taking this issue. As we found the rootcause of many issues (#714) I will return the version number to its normal form (3.5.3 or another higher number) in 5.0.3

mykola-mokhnach:
@TikhomirovSergey 

This is what I have found on this topic so far:
http://gradle.1045684.n5.nabble.com/Translating-Ivy-dep-version-syntax-to-Maven-in-POM-generation-td5712755.html

Probably, we can use https://github.com/nebula-plugins/nebula-publishing-plugin, which can properly publish version ranges into Maven pom files.

TikhomirovSergey:
@mykola-mokhnach Ok. Will try the nebula plugin. Thank you.

realnikunj:
I am facing this same issue with java-client version 5.0.2

[...]

PR #722, opened 2017-09-07
#718 FIX

TikhomirovSergey:
## Change list

#718 FIX
 
## Types of changes

## Details

There was the dependency conflict inside java_clint. As it was resolved by exclusion of malformed artefacts, I think we could return selenium dependency version to its normal form with force compilation.  I think that it is the case when the simple solution is enough.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ apply plugin: 'signing'
 apply plugin: 'maven-publish'
 
 group 'io.appium'
-version '5.0.2'
+version '5.0.3'
 
 repositories {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -54,7 +54,7 @@ compileJava {
     ]
 }
 
-ext.seleniumVersion = '[3.5.3,3.5.3]'
+ext.seleniumVersion = '3.5.3'
```
mykola-mokhnach: nit: add a comment here about version number format restriction

Issue #494, opened 2016-10-21
'gradlew test' target is broken on master


artkoshelev:
I've cloned the latest version from github and run `gradlew test`.  59 of 120 tests fail, here is output https://gist.github.com/artkoshelev/45d387c080580c7d13b336745c492450

TikhomirovSergey:
Hi @artkoshelev 
It is not the bug. This is probably the issue of your environment. 

Why does it happen? 
It cann't find the `main.js` file. 
Ways to resolve the problem:

1. 
Execute the command line `npm root -g`. It will show you the path to the directory where packages are installed by default. Then please check it. There should be the directory `/appium/lib` which contains `main.js`. If there is no such directory or file then appium should be installed globally via npm or you can place required files and folders to the directory.

2. Another way is described here: 
https://github.com/appium/java-client/blob/master/docs/The-starting-of-an-app-using-Appium-node-server-started-programmatically.md#fyi. You can add the **APPIUM_BINARY_PATH** and define the full path to the `main.js`.

artkoshelev:
It's quite weird to depend on some js code to exist in java library

artkoshelev:
anyway, after installing appium with `npm -g appium` i still got 59 tests failed https://gist.github.com/artkoshelev/7e406524f9ad8aba5c3d714ee5ebf3e7

TikhomirovSergey:
Strange. 
However there is another exception. 

Previous:
>io.appium.java_client.service.local.InvalidServerInstanceException
        Caused by: java.io.IOException

Now
> org.openqa.selenium.WebDriverException

@artkoshelev Can you provide java stacktrace?

artkoshelev:
I've figured out i need an android device connected as well, and now i need to install some app's inside. Is there any description of all the ENV i need to setup manually before running tests?

TikhomirovSergey:
@artkoshelev 
ENV settings are described [here](https://github.com/appium/appium/blob/master/docs/en/appium-setup/android-setup.md), [here](https://github.com/appium/appium/blob/master/docs/en/appium-setup/running-on-windows.md) and [here](https://github.com/appium/appium/blob/master/docs/en/appium-setup/android-hax-emulator.md).

All the appium setup instructions you can find [in this directory](https://github.com/appium/appium/tree/master/docs/en/appium-setup).

>and now i need to install some app's inside

I can't understand it :) If an app is defined at capabilities then it should be installed by appium automatically. Please clarify it.

[...]

Issue #325, opened 2016-02-25
tests should be redesigned 


TikhomirovSergey:
They should:

- run rapidly

- monatomic (an one function should be checked per one test)

- allow to use CI environment or settings which preferred by a contributor

TikhomirovSergey:
https://github.com/appium/java-client/issues/269

AlessandroMiccoli:
Is there any tests I can pick up for improvement?

mykola-mokhnach:
@AlessandroMiccoli There are definitely many tests that need to be fixed. Check, for example, https://github.com/appium/java-client/blob/6b6de17409a26c080597e95ffee15c2a430b101a/build.gradle#L269

PR #728, opened 2017-09-16
Tests of local appium DriverService were re-designed.

TikhomirovSergey:
## Change list

Test of the appium-specific DriverService implementation were improved/redesigned.
 
## Types of changes

## Details

We are going to improve tests and make them runnable on many environments.  
#325 #494


src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+import java.io.FileOutputStream;
+import java.io.OutputStream;
+import java.net.InetAddress;
+import java.net.NetworkInterface;
+import java.util.Enumeration;
+import java.util.List;
+
+public class ServerBuilderTest {
+
+    private static String pathToAppiumNodeInProperties;
```
mykola-mokhnach: probably, we can mark properties, which are not going to be changed, as final

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+                if (!inetAddress.isLoopbackAddress()) {
+                    InetAddressValidator validator = InetAddressValidator.getInstance();
+                    testIP = inetAddress.getHostAddress().toString();
+                    if (validator.isValid(testIP)) {
+                        break;
+                    }
+                    testIP = null;
+                }
+            }
+            if (testIP != null) {
```
mykola-mokhnach: I assume having this in a separate private method would save us several lines of code (break -> return) and will improve readability.

mykola-mokhnach: the returned result can be of type Optional

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+    @After
+    public void tearDown() throws Exception {
+        ofNullable(service).ifPresent(AppiumDriverLocalService::stop);
+
+        if (stream != null) {
+            stream.close();
+        }
+
+        ofNullable(file).ifPresent(fileArg -> {
+            if (file.exists()) {
```
mykola-mokhnach: file or fileArg?

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+    }
+
+    @Test public void checkAbilityToFindNodeDefinedInProperties() {
+        String definedNode = new File("src/test/java/io/appium/java_client/service/local/appium.js").getAbsolutePath();
+        setProperty(APPIUM_PATH, definedNode);
+        assertThat(new File(new AppiumServiceBuilder().createArgs().get(0)), is(new File(definedNode)));
+    }
+
+    @Test public void checkAbilityToUseNodeDefinedExplicitly() {
+        File node = new File("src/test/java/io/appium/java_client/service/local/appium.js");
```
mykola-mokhnach: the magic string can be moved to constants

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+            .withArgument(CALLBACK_ADDRESS, testIP)
+            .withArgument(SESSION_OVERRIDE)
+            .withArgument(PRE_LAUNCH)
+            .build();
+        service.start();
+        assertTrue(service.isRunning());
+    }
+
+    @Test public void checkAbilityToStartServiceUsingCapabilities() throws Exception {
+        File appDir = new File("src/test/java/io/appium/java_client");
```
mykola-mokhnach: there are many common root paths. The common part can be represented as a constant, so then one can use new File(parent, basename) constructor, where parent is a constant

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+    }
+
+    @Test public void checkAbilityToChangeOutputStream() throws Exception {
+        file = new File("test");
+        file.createNewFile();
+        stream = new FileOutputStream(file);
+        service = buildDefaultService();
+        service.addOutPutStream(stream);
+        service.start();
+        assertTrue(file.length() > 0);
```
mykola-mokhnach: -> assertThat(file.length(), greaterThan(0L))

mykola-mokhnach: this is a bit longer, but produces better error message in comparison to the generic True assertion

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+        assertFalse(service.isRunning());
+    }
+
+    @Test public void checkAbilityToStartAndShutDownFewServices() throws Exception {
+        List<AppiumDriverLocalService> services = asList(
+                new AppiumServiceBuilder().usingAnyFreePort().build(),
+                new AppiumServiceBuilder().usingAnyFreePort().build(),
+                new AppiumServiceBuilder().usingAnyFreePort().build(),
+                new AppiumServiceBuilder().usingAnyFreePort().build());
+        services.parallelStream().forEach(AppiumDriverLocalService::start);
```
mykola-mokhnach: ❤️

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,291 @@
...
+        assertTrue(services.stream().noneMatch(AppiumDriverLocalService::isRunning));
+    }
+
+    @Test public void checkAbilityToStartServiceWithLogFile() throws Exception {
+        file = new File("Log.txt");
+        file.createNewFile();
+        service = new AppiumServiceBuilder().withLogFile(file).build();
+        service.start();
+        assertTrue(file.exists());
+        assertTrue(file.length() > 0);
```
mykola-mokhnach: same as above

mykola-mokhnach:
Shall we add a description somewhere about src/test/java/io/appium/java_client/service/local/appium.js is just a stub?

TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget 
Code was improved

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -58,16 +68,16 @@
...
-                        break;
-                    }
-                    testIP = null;
+
+                    testIP = ofNullable(testIP).orElseGet(() -> {
+                        String result = inetAddress.getHostAddress();
+                        if (validator.isValid(result)) {
+                            return result;
+                        }
+                        return null;
```
SrinivasanTarget: return validator.isValid(result) ? result : null;

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,301 @@
...
+        for (Enumeration<NetworkInterface> en = NetworkInterface.getNetworkInterfaces(); en
+            .hasMoreElements(); ) {
+            NetworkInterface intf = en.nextElement();
+            for (Enumeration<InetAddress> enumIpAddr = intf.getInetAddresses(); enumIpAddr
+                .hasMoreElements(); ) {
+                InetAddress inetAddress = enumIpAddr.nextElement();
+                if (!inetAddress.isLoopbackAddress()) {
+                    InetAddressValidator validator = InetAddressValidator.getInstance();
+
+                    testIP = ofNullable(testIP).orElseGet(() -> {
```
mykola-mokhnach: this won't work as expected, because the iteration does not stop here even if _testIP_ set. Please extract the loop into a separate method `private static Optional<String> getTestIP()`, so then beforeClass's body should look like `testIP = getTestIP().orElseThrow(() -> new IllegalStateException("Cannot initialize the IP address for tests"))`

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,301 @@
...
+public class ServerBuilderTest {
+
+    /**
+     * It may be impossible to find the path to the instance of appium server due to different circumstance.
+     * So user may use environment variables/system properties to define the full path to the server
+     * appium.js that is supposed to be default.
+     */
+    private static final String PATH_TO_APPIUM_NODE_IN_PROPERTIES = getProperty(APPIUM_PATH);
+
+    private static final String ROOT_TEST_PATH = "src/test/java/io/appium/java_client/";
```
mykola-mokhnach: It will be more handy to keep these as File objects instead:
`TEST_PATH = new File("src/test/java/io/appium/java_client")`
`TEST_APPIUM_JS = new File(TEST_PATH.getPath(), "service/local/appium.js")`

mykola-mokhnach: trailing slash is not needed if path concatenation is done in File constructor

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,301 @@
...
+
+    /**
+     * This is the path to the stub appium.js file
+     */
+    private static final String PATH_T0_TEST_APPIUM_JS = ROOT_TEST_PATH + "service/local/appium.js";
```
mykola-mokhnach: It will be more handy to keep it as File object:
`TEST_APPIUM_JS = Path.get(ROOT_TEST_PATH).resolve("service").resolve("local").resolve("appium.js").toFile()`

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
mykola-mokhnach: also, will these tests work properly in Windows if one hardcodes path separators?

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,301 @@
...
+        }
+
+        ofNullable(testLogFile).ifPresent(savedTestLogFile -> {
+            if (savedTestLogFile.exists()) {
+                savedTestLogFile.delete();
+            }
+        });
+
+        System.clearProperty(APPIUM_PATH);
+        ofNullable(PATH_TO_APPIUM_NODE_IN_PROPERTIES).ifPresent(s -> setProperty(APPIUM_PATH, s));
```
mykola-mokhnach: the constant cannot be null

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,301 @@
...
+        caps.setCapability(CHROMEDRIVER_EXECUTABLE, chrome.getAbsolutePath());
+
+        service = new AppiumServiceBuilder().withCapabilities(caps).build();
+        service.start();
+        assertTrue(service.isRunning());
+    }
+
+    @Test public void checkAbilityToStartServiceUsingCapabilitiesAndFlags() throws Exception {
+        File app = new File(ROOT_TEST_PATH, "ApiDemos-debug.apk");
+        File chrome = new File(new File(ROOT_TEST_PATH, "pagefactory_tests"), "chromedriver.exe");
```
mykola-mokhnach: Use Path.resolve method to construct longer paths

TikhomirovSergey:
Ok. Will improve it ASAP

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
SrinivasanTarget: @TikhomirovSergey Why don't we name stub as `main.js` in sync with appium server's `main.js`?

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
TikhomirovSergey: @mykola-mokhnach  It may be when the property was not defined

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
TikhomirovSergey: @SrinivasanTarget 
Because it searches for appium.js when it tries to launch the node locally

saikrishna321: @TikhomirovSergey with the latest appium its main.js

TikhomirovSergey: https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java#L70

@SrinivasanTarget @saikrishna321 Oops! My fault. yes. Guys you are right :)

TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget 
Could you take a look one more time. It seems it should be finally resolved.

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+     */
+    private static final Path PATH_T0_TEST_MAIN_JS = ROOT_TEST_PATH
+            .resolve("service").resolve("local").resolve("main.js");
+
+    private static String testIP;
+    private AppiumDriverLocalService service;
+    private File testLogFile;
+    private OutputStream stream;
+
+    private static String getLocalIP(NetworkInterface intf) {
```
mykola-mokhnach: please add Nullable annotation

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+        String result = null;
+        for (Enumeration<InetAddress> enumIpAddr = intf.getInetAddresses(); enumIpAddr
+                .hasMoreElements(); ) {
+            String calculated;
+            InetAddress inetAddress = enumIpAddr.nextElement();
+            if (!inetAddress.isLoopbackAddress()) {
+                InetAddressValidator validator = InetAddressValidator.getInstance();
+                calculated = inetAddress.getHostAddress().toString();
+                if (validator.isValid(calculated)) {
+                    result = calculated;
```
mykola-mokhnach: return calculated

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+                    break;
+                }
+            }
+        }
+        return result;
```
mykola-mokhnach: `return null`, _result_ variable is not needed.

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+    private static String testIP;
+    private AppiumDriverLocalService service;
+    private File testLogFile;
+    private OutputStream stream;
+
+    private static String getLocalIP(NetworkInterface intf) {
+        String result = null;
+        for (Enumeration<InetAddress> enumIpAddr = intf.getInetAddresses(); enumIpAddr
+                .hasMoreElements(); ) {
+            String calculated;
```
mykola-mokhnach: what's the point to define _calculated_ before if condition?

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+    }
+
+    @Test public void checkAbilityToStartDefaultService() {
+        service = buildDefaultService();
+        service.start();
+        assertEquals(true, service.isRunning());
+    }
+
+    @Test public void checkAbilityToFindNodeDefinedInProperties() {
+        File definedNode = PATH_T0_TEST_MAIN_JS.toFile();
```
mykola-mokhnach: ❤️

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+        service = new AppiumServiceBuilder()
+                .withArgument(() -> "-a", testIP)
+                .build();
+        String actualUrl = service.getUrl().toString();
+        assertEquals(expectedUrl, actualUrl);
+        service.start();
+    }
+
+    @Test public void checkAbilityToStartServiceWithLogFileUsingFlag() throws Exception {
+        testLogFile = new File("Log2.txt");
```
mykola-mokhnach: shouldn't we delete these temp files after the test is finished?

mykola-mokhnach: the same comment is applicable for other tests where logs are checked

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -1,311 +1,312 @@
...
+                break;
+            }
+        }
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        ofNullable(service).ifPresent(AppiumDriverLocalService::stop);
+
+        if (stream != null) {
```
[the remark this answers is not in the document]
TikhomirovSergey: It throws the checked IO exception that needs to be handled by catch. So I decided to don't change code there because lambda with the catch clause looks not so cool.

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
TikhomirovSergey: Will try it but codecy is sensetive to things like that

Issue #735, opened 2017-09-25
AppiumElementLocator & Implicit Waits


titusfortner:
Am I reading [this code](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java) correctly, that every single element location in the Java bindings makes multiple implicit wait wire calls? 

When running a test on a remote platform this causes a huge amount of overhead, exacerbated by any increased latency from a proxy, etc during any given run. Implicit waits are heavily discouraged by the Selenium developers (including the guy who created them in the first place). I think it should be relatively easy to change this code to be implemented by an explicit wait strategy, which would only make additional wire calls if they are actually needed.

Are there any other reasons that I'm not aware of that make implicit waits necessary, or can someone pair with me on a PR for this?

jlipps:
Yeah, seems weird for a client to make so many implicit wait calls. Could lead to a performance degradation, especially when there's latency on the wire.

TikhomirovSergey:
@titusfortner @jlipps the new build is going to be published today

PR #738, opened 2017-09-27
Explicit

titusfortner:
## Change list
* Changes name of constant since it actually isn't being used implicitly
* Removes the resetting of implicit waits
 
## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

Implicit waits are currently entirely overridden. Setting implicit wait to 0.5 or 300 won't make any difference when locating elements.

I don't think we should remove the user's ability to set an implicit wait, even if it can have negative consequences for them.

This is a "breaking" change because implicit waits will now be respected, which can allow users to combine implicit and explicit in ways that are not ideal, or cause longer test runs if users are testing for the absence of an element. But I think them not being respected before was technically the unexpected behavior, and this is essentially a bug fix for that.

I think this should be released with the recognition that people might have timing issues and that they should be told not to set implicit waits if they want to keep previous behavior.

I'm trying to think about how overriding the implicit wait setting in the appium driver, storing the value and doing fancy things with it during location method could work... but all I come up with would remove options from the user that I think they should have.

Anyway, let's discuss. Feel free to use/modify this code as necessary in anything you end up doing..

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
@@ -62,7 +62,7 @@ public class AppiumFieldDecorator implements FieldDecorator {
     private static final List<Class<? extends WebElement>> availableElementClasses = ImmutableList.of(WebElement.class,
             RemoteWebElement.class, MobileElement.class, AndroidElement.class,
             IOSElement.class, WindowsElement.class);
-    public static long DEFAULT_IMPLICITLY_WAIT_TIMEOUT = 1;
+    public static long DEFAULT_TIMEOUT = 1;
     public static TimeUnit DEFAULT_TIMEUNIT = TimeUnit.SECONDS;
     private final WebDriver originalDriver;
     private final DefaultFieldDecorator defaultElementFieldDecoracor;
@@ -73,9 +73,9 @@ public class AppiumFieldDecorator implements FieldDecorator {
     private final HasSessionDetails hasSessionDetails;
 
 
-    public AppiumFieldDecorator(SearchContext context, long implicitlyWaitTimeOut,
+    public AppiumFieldDecorator(SearchContext context, long timeOut,
         TimeUnit timeUnit) {
-        this(context, new TimeOutDuration(implicitlyWaitTimeOut, timeUnit));
+        this(context, new TimeOutDuration(timeOut, timeUnit));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


titusfortner:
Oh, forgot to mention this is for #735

TikhomirovSergey:
@titusfortner Thank you very much. I will review it soon.

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
```diff
@@ -73,9 +73,9 @@
     private final HasSessionDetails hasSessionDetails;
 
 
-    public AppiumFieldDecorator(SearchContext context, long implicitlyWaitTimeOut,
+    public AppiumFieldDecorator(SearchContext context, long timeOut,
```
mykola-mokhnach: -> timeout

titusfortner: There are 32 places that use `timeOut` vs none with `timeout`. This PR should keep it consistent, though a later commit to change all usages is probably appropriate.

titusfortner:
> I think we should warn users to not use `driver.manage().timeouts().implicitlyWait()` if they use page/screen objects.

I'm not sure what difference it makes page objects vs non. Isn't this locator code what is used regardless?

That being said, I'm all about warning users against using implicit waits, is there a mechanism to do so, or are we just talking about in the responses we give in the likely github issues? :)

@TikhomirovSergey - I removed `changeImplicitlyWaitTimeOut`, so I think this should be good.

SrinivasanTarget:
> That being said, I'm all about warning users against using implicit waits, is there a mechanism to do so, or are we just talking about in the responses we give in the likely github issues? :)

@titusfortner @TikhomirovSergey AFAIK Selenium is planning to deprecate `implicit waits` soon. Simon had spoke about deprecating it on last SeleniumConf. May be we can wait for it.

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
SrinivasanTarget: Why can't we change all the instances?

TikhomirovSergey:
@SrinivasanTarget ok. So there is nothing to worry about.

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
titusfortner: We can/should, just didn't want to deal with those changes in this PR/release. :-)

titusfortner:
@SrinivasanTarget Simon said not to use them, but he still put them in the w3c spec, so I don't think they'll be actually deprecated any time soon.

SrinivasanTarget:
> Simon said not to use them, but he still put them in the w3c spec, so I don't think they'll be actually deprecated any time soon.

Ok

PR #741, opened 2017-09-29
The addition to #738

TikhomirovSergey:
## Change list

The addition to #738
 
## Types of changes

## Details

The addition to #738  …
- following dependencies were updated:
  `org.seleniumhq.selenium:selenium-java` to 3.6.0
  `com.google.code.gson:gson` to 2.8.2
  `org.springframework:spring-context` to 5.0.0.RELEASE
  `org.aspectj:aspectjweaver` to 1.8.11

 - unused parameters and fields were removed from AppiumElementLocator,
 AppiumElementLocatorFactory

 - some test improving


src/test/java/io/appium/java_client/pagefactory_tests/MobileBrowserCompatibilityTest.java
```diff
@@ -44,7 +44,7 @@
     private AppiumDriverLocalService service;
 
     @AndroidFindBy(className = "someClass") @AndroidFindBy(xpath = "//someTag")
-    private RemoteWebElement btnG; //this element should be found by id = 'btnG' or name = 'btnG'
+    private RemoteWebElement btnK; //this element should be found by id = 'btnG' or name = 'btnG'
```
titusfortner: If this is changed, then the comment should also be changed?

TikhomirovSergey: ok. I will improve it a bit later.

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -61,29 +59,21 @@
      *                          are found once should be cached
      * @param duration          is a POJO which contains timeout parameters for the element to be searched
      * @param originalDuration  is a POJO which contains timeout parameters from page object which contains the element
-     * @param originalWebDriver is an instance of WebDriver that is going to be
-     *                          used by a proxied element
      */
 
     public AppiumElementLocator(SearchContext searchContext, By by, boolean shouldCache,
-        TimeOutDuration duration, TimeOutDuration originalDuration, WebDriver originalWebDriver) {
+        TimeOutDuration duration, TimeOutDuration originalDuration) {
```
mykola-mokhnach: would it be possible to also replace the custom TimeOutDuration class with the native Duration one, since we anyway change method signature?

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocatorFactory.java
```diff
@@ -21,30 +21,26 @@
...
     /**
      * Creates a new mobile element locator factory.
      *
      * @param searchContext     The context to use when finding the element
      * @param timeOutDuration   is a POJO which contains timeout parameters for the element to be searched
-     * @param originalWebDriver is an instance of WebDriver that is going to be used by a proxied element
      * @param builder           is handler of Appium-specific page object annotations
      */
 
     public AppiumElementLocatorFactory(SearchContext searchContext, TimeOutDuration timeOutDuration,
```
mykola-mokhnach: same comment as above

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
```diff
@@ -62,7 +62,7 @@
     private static final List<Class<? extends WebElement>> availableElementClasses = ImmutableList.of(WebElement.class,
             RemoteWebElement.class, MobileElement.class, AndroidElement.class,
             IOSElement.class, WindowsElement.class);
-    public static long DEFAULT_IMPLICITLY_WAIT_TIMEOUT = 1;
+    public static long DEFAULT_TIMEOUT = 1;
```
mykola-mokhnach: we can deprecate the constant of TimeUnit type if we use Duration type for DEFAULT_TIMEOUT

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
```diff
@@ -73,9 +73,9 @@
     private final HasSessionDetails hasSessionDetails;
 
 
-    public AppiumFieldDecorator(SearchContext context, long implicitlyWaitTimeOut,
+    public AppiumFieldDecorator(SearchContext context, long timeOut,
```
mykola-mokhnach: timeout

src/test/java/io/appium/java_client/ChromeDriverPathUtil.java
```diff
@@ -0,0 +1,34 @@
...
+import static org.openqa.selenium.Platform.WINDOWS;
+import static org.openqa.selenium.Platform.getCurrent;
+
+import org.openqa.selenium.Platform;
+
+import java.io.File;
+import java.nio.file.Path;
+
+public final class ChromeDriverPathUtil {
+    private static final Path ROOT_TEST_PATH = getDefault().getPath("src")
```
mykola-mokhnach: ❤️

src/test/java/io/appium/java_client/ChromeDriverPathUtil.java
```diff
@@ -0,0 +1,34 @@
...
+            .resolve("test").resolve("java").resolve("io").resolve("appium").resolve("java_client");
+
+    /**
+     * @return the choromedriver file which depends on platform.
+     */
+    public static File getChromeDriver() {
+        Path resultPath;
+        Platform current = getCurrent();
+
+        if (current.is(WINDOWS)) {
```
mykola-mokhnach: why cannot we use `switch () case ...: return` here? this will allow to get rid of many temporary variables

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.openqa.selenium.chrome.ChromeDriver;
+import org.openqa.selenium.chrome.ChromeDriverService;
+import org.openqa.selenium.support.FindAll;
+import org.openqa.selenium.support.FindBy;
+
+import java.util.Calendar;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class TimeOutTest {
```
mykola-mokhnach: TimeoutTest

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final long ACCEPTABLE_DELTA_MILLS = 1500;
```
mykola-mokhnach: ->Duration

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
```diff
@@ -0,0 +1,132 @@
...
+    private List<WebElement> stubElements;
+
+    @WithTimeout(time = 5, unit = SECONDS)
+    @FindAll({@FindBy(className = "ClassWhichDoesNotExist"),
+            @FindBy(className = "OneAnotherClassWhichDoesNotExist")})
+    private List<WebElement> stubElements2;
+
+    private TimeOutDuration timeOutDuration;
+
+    private static long getBenchMark(List<WebElement> stubElements) {
```
mykola-mokhnach: getBenchmark

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
```diff
@@ -0,0 +1,132 @@
...
+        long startMark = Calendar.getInstance().getTimeInMillis();
```
mykola-mokhnach: I usually use `System.nanoTime()` for such purpose or `System.currentTimeMillis()` if we don't need nano-preciseness. _Calendar_ class has different purpose.

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
mykola-mokhnach: One can make this method more generic by providing a lambda function to benchmark:

```java
private static long getBenchmark(Runnable dstFunc) {
final long startTime = System.nanoTime();
dstFunc.run();
return System.nanoTime() - startTime;
}
```

src/test/java/io/appium/java_client/ChromeDriverPathUtil.java
TikhomirovSergey: @mykola-mokhnach 

Because there may be WINDOWS8, WINDOWS10, WINDOWS_XP etc. I think that it is easier to compare current platforn with WINDOWS (the result for WINDOWS8, WINDOWS10, WINDOWS_XP is TRUE)

src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java
TikhomirovSergey: yes :) it was taken from the old code of the test. will improve it.

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
TikhomirovSergey: @mykola-mokhnach 
I think these parameters are needed still.

TimeOutDuration is just the container of duration values. 
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/pagefactory/TimeOutDuration.java

It was designed for the purposes like 
https://github.com/TikhomirovSergey/java-client/blob/73077776d513e4521d7d5e8304469a4b85232e05/src/test/java/io/appium/java_client/pagefactory_tests/TimeOutTest.java#L122

I think it is more useful to completely change value than do 'plus' or 'minus'. 
However, I think I could redesign `TimeOutDuration` and make it work with `java.time.Duration`. I think it should be done step by step. The first is to mark some things `@Deprecated` and to add new good things instead, the second is removal. 

I just think that the change of TimeOutDuration may impact many users.

mykola-mokhnach: yep, I agree about we can do a separate PR for it

src/test/java/io/appium/java_client/ChromeDriverPathUtil.java
mykola-mokhnach: np if you prefer this. However it still can be simplified by retuning the value immediately inside if block

TikhomirovSergey:
@mykola-mokhnach This PR was updated. 
Also I have opened the #742. It is going to be closed in the next release.

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocatorFactory.java
```diff
@@ -59,16 +57,15 @@ public CacheableLocator createLocator(Field field) {
...
-        if (by != null) {
-            return new AppiumElementLocator(searchContext, by, builder.isLookupCached(),
-                    customDuration, timeOutDuration, originalWebDriver);
-        }
-        return null;
+        By byResult = builder.buildBy();
+
+        return ofNullable(byResult)
+                .map(by -> new AppiumElementLocator(searchContext, by, builder.isLookupCached(), customDuration))
+                .orElse(null);
```
mykola-mokhnach: The method can be marked as Nullable

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
```diff
@@ -62,20 +62,20 @@
...
     private final AppiumElementLocatorFactory widgetLocatorFactory;
     private final String platform;
     private final String automation;
-    private final TimeOutDuration timeOutDuration;
+    private final TimeOutDuration duration;
     private final HasSessionDetails hasSessionDetails;
 
 
-    public AppiumFieldDecorator(SearchContext context, long implicitlyWaitTimeOut,
+    public AppiumFieldDecorator(SearchContext context, long timeout,
```
mykola-mokhnach: I'd rather rename it to timeoutMs, since we don't mention the measurement unit anywhere explictly

src/test/java/io/appium/java_client/ChromeDriverPathUtil.java
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * @return the choromedriver file which depends on platform.
+     */
+    public static File getChromeDriver() {
+        Platform current = getCurrent();
+        if (current.is(WINDOWS)) {
+            return ROOT_TEST_PATH.resolve("chromedriver.exe").toFile();
+        } else if (current.is(MAC)) {
+            return ROOT_TEST_PATH.resolve("chromedriver_mac").toFile();
+        } else {
```
mykola-mokhnach: this `else` is not needed

src/test/java/io/appium/java_client/pagefactory_tests/TimeoutTest.java
```diff
@@ -0,0 +1,132 @@
...
+    /**
+     * finishing.
+     */
+    @After public void tearDown() throws Exception {
+        driver.quit();
+    }
+
+    @Test public void defaultTimeOutTest() {
+        assertThat(format(MESSAGE, DEFAULT_TIMEOUT, DEFAULT_TIMEUNIT),
+                abs(getExpectedMillis(DEFAULT_TIMEOUT, DEFAULT_TIMEUNIT) - getBenchmark(() -> stubElements.size())),
```
mykola-mokhnach: I assume the whole formula can be extracted into a separate method, for example:
```java
private static long getPerformanceDiff(long expectedMs, Runnable m) {
   returns abs(expectedMs  - getBenchmark(m1))
}
```

src/test/java/io/appium/java_client/pagefactory_tests/TimeoutTest.java
```diff
@@ -0,0 +1,132 @@
...
+import org.openqa.selenium.chrome.ChromeDriverService;
+import org.openqa.selenium.support.FindAll;
+import org.openqa.selenium.support.FindBy;
+
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class TimeoutTest {
+
+    private static final long ACCEPTABLE_TIME_DIFF = 1500;
```
mykola-mokhnach: It is a good practice to explicitly mention measurement unit in var name in case it is of primitive data type (e.g. _MS, _SEC)

mykola-mokhnach:
Replacing DIFF with DELTA might be easier to read. Otherwise LGTM

PR #756, opened 2017-10-15
Refactor setting of touch actions options

mykola-mokhnach:
## Change list

The idea of this PR is to unify touch action options, so it is easier to extend them. 
The main reason for this change was my intension to add _pressure_ argument to press/longPress actions for iOS. With the current implementation the count of overriden methods in TouchAction class would have grown up exponentially.
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/TouchAction.java
+++ b/src/main/java/io/appium/java_client/TouchAction.java
@@ -19,10 +19,17 @@ package io.appium.java_client;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
+import io.appium.java_client.touch.ActionOptions;
+import io.appium.java_client.touch.LongPressOptions;
+import io.appium.java_client.touch.MoveToOptions;
+import io.appium.java_client.touch.PressOptions;
+import io.appium.java_client.touch.TapOptions;
+import io.appium.java_client.touch.WaitOptions;
 import org.openqa.selenium.WebElement;
-import org.openqa.selenium.internal.HasIdentity;
 
 import java.time.Duration;
+
+import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * Used for Webdriver 3 touch actions
@@ -34,7 +41,7 @@ import java.time.Duration;
  * Calling perform() sends the action command to the Mobile Driver. Otherwise,
  * more and more actions can be chained.
  */
-public class TouchAction implements PerformsActions<TouchAction> {
+public class TouchAction<T extends TouchAction<T>> implements PerformsActions<T> {
 
     protected ImmutableList.Builder<ActionParameter> parameterBuilder;
     private PerformsTouchActions performsTouchActions;
@@ -45,15 +52,32 @@ public class TouchAction implements PerformsActions<TouchAction> {
     }
 
     /**
+     * Press on an element.
+     *
+     * @param pressOptions see {@link PressOptions}.
+     * @return this TouchAction, for chaining.
+     */
+    public T press(ActionOptions pressOptions) {
+        parameterBuilder.add(new ActionParameter("press", pressOptions));
+        //noinspection unchecked
+        return (T) this;
+    }
+
+    /**
      * Press on the center of an element.
      *
      * @param el element to press on.
      * @return this TouchAction, for chaining.
+     * @deprecated use {@link #press(ActionOptions)} instead
      */
-    public TouchAction press(WebElement el) {
-        ActionParameter action = new ActionParameter("press", (HasIdentity) el);
+    @Deprecated
+    public T press(WebElement el) {
+        ActionParameter action = new ActionParameter("press",
+                new PressOptions()
+                        .withElement(el));
         parameterBuilder.add(action);
-        return this;
+        //noinspection unchecked
+        return (T) this;
     }
 
     /**
@@ -62,13 +86,16 @@ public class TouchAction implements PerformsActions<TouchAction> {
      * @param x x coordinate.
      * @param y y coordinate.
      * @return this TouchAction, for chaining.
+     * @deprecated use {@link #press(ActionOptions)} instead
      */
-    public TouchAction press(int x, int y) {
-        ActionParameter action = new ActionParameter("press");
-        action.addParameter("x", x);
-        action.addParameter("y", y);
+    @Deprecated
+    public T press(int x, int y) {
+        ActionParameter action = new ActionParameter("press",
+                new PressOptions()
+                        .withAbsoluteOffset(x, y));
         parameterBuilder.add(action);
-        return this;
+        //noinspection unchecked
+        return (T) this;
     }
 
     /**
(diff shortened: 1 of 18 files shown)


mykola-mokhnach:
@KazuCocoa I would be happy to see your feedback on this change. Would it be useful to set action options this way? If yes, then would it be necessary to have something like this in Ruby/Python libs?

KazuCocoa:
looks good for me. 🆗 
For Ruby/Python client, it isn't probably necessary for them, but I'd like to consider a bit.

src/main/java/io/appium/java_client/TouchAction.java
```diff
@@ -118,13 +168,16 @@ public TouchAction moveTo(WebElement el) {
...
      */
-    public TouchAction moveTo(int x, int y) {
-        ActionParameter action = new ActionParameter("moveTo");
-        action.addParameter("x", x);
-        action.addParameter("y", y);
+    @Deprecated
+    public T moveTo(int x, int y) {
+        ActionParameter action = new ActionParameter("moveTo",
+                new MoveToOptions()
+                        .withRelativeOffset(x, y));
```
SrinivasanTarget: @mykola-mokhnach Isn't it absoluteOffset?

mykola-mokhnach: > * @param x change in x coordinate to move through.
> * @param y change in y coordinate to move through.

That is why I find the current approach useful, because it helps to avoid such misunderstandings while settings options ;)

SrinivasanTarget: @mykola-mokhnach haha makes sense.thanks :)

mykola-mokhnach:
@TikhomirovSergey when do you have time to take a look at the PR?

src/main/java/io/appium/java_client/ios/IOSTouchAction.java
```diff
@@ -1,12 +1,28 @@
...
 
 import io.appium.java_client.PerformsTouchActions;
 import io.appium.java_client.TouchAction;
+import io.appium.java_client.ios.touch.DoubleTapOptions;
 import org.openqa.selenium.WebElement;
-import org.openqa.selenium.internal.HasIdentity;
 
 
-public class IOSTouchAction extends TouchAction {
+public class IOSTouchAction extends TouchAction<IOSTouchAction> {
```
TikhomirovSergey: @mykola-mokhnach Why this class is not fully deprecated?

src/main/java/io/appium/java_client/android/AndroidTouchAction.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package io.appium.java_client.android;
+
+import io.appium.java_client.PerformsTouchActions;
+import io.appium.java_client.TouchAction;
+
+
+public class AndroidTouchAction extends TouchAction<AndroidTouchAction> {
```
TikhomirovSergey: What the purpose of this class?

src/main/java/io/appium/java_client/ios/IOSTouchAction.java
```diff
@@ -18,12 +34,15 @@ public IOSTouchAction(PerformsTouchActions performsTouchActions) {
...
-     * @return this TouchAction, for chaining.
+     * @return this IOSTouchAction, for chaining.
+     * @deprecated use {@link #tap(ActionOptions)} with count=2 instead.
      */
+    @Deprecated
     public IOSTouchAction doubleTap(WebElement el, int x, int y) {
-        ActionParameter action = new ActionParameter("doubleTap", (HasIdentity) el);
-        action.addParameter("x", x);
-        action.addParameter("y", y);
+        ActionParameter action = new ActionParameter("doubleTap",
```
TikhomirovSergey: Also
Won't the `doubleTap`  be supported anymore?

src/main/java/io/appium/java_client/ios/touch/DoubleTapOptions.java
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client.ios.touch;
+
+import io.appium.java_client.touch.OptionsWithAbsolutePositioning;
```
TikhomirovSergey: If this class was added only for bacward compatibility so I think it might be better to make it nested in `IOSTouchAction`

src/main/java/io/appium/java_client/android/AndroidTouchAction.java
mykola-mokhnach: This class will be needed in the future to add Android-specific touch actions (as well as the IOS one). 

Also, there is second very Java-specific reason. This class is a specific implementation of Builder pattern with inheritance. If you check the ancestor classes they all have quite specific generic signatures and this all is done to make builder methods return appropriate class without casting, for example:

```java
new IOSTouchAction(driver).press(...).doubleTap(...)
```

won't work in the previous implementation, because _press_ returns the instance of TouchAction class instead of IOSTouchAction, where this method is implemented, so it is necessary to perform class cast:

```java
((IOSTouchAction)new IOSTouchAction(driver).press(...)).doubleTap(...)
```

This approach fixes it, but now Java prevents one from passing TouchAction as a generic parameter, so Supplier<TouchAction> won't work, but Supplier<IOSTouchAction> works as expected. https://stackoverflow.com/questions/21086417/builder-pattern-and-inheritance has more details.

src/main/java/io/appium/java_client/ios/IOSTouchAction.java
mykola-mokhnach: This class will be needed to add further TouchAction extensions

src/main/java/io/appium/java_client/ios/IOSTouchAction.java
mykola-mokhnach: It looks redundant to me. doubleTap can be easily replaced with tap(count:2) or double tap-wait-tap chains.
Would you like to keep it?

src/main/java/io/appium/java_client/ios/touch/DoubleTapOptions.java
mykola-mokhnach: but this should be visible for TouchAction options as well.

mykola-mokhnach:
> why there some empty classes? May some things be generilized?

See my comment above about builder generics. Also, it will be possible to add more stuff into these classes later, for example some special platform-specific gestures, which we don't support yet.

> is it safe to declare something like public T waitAction(ActionOptions waitOptions)? What is user will try something like to.wait(optionsWithAbsolutePositioning)? Should methods have more strict signature?

I thought about it... From code creation perspective the current approach should not be a problem, since it's pretty obvious to see an error in the method like `tap(new LongPressOptions())`, which is the case for dynamically typed non-Java world. And this gives us more freedom if we want to change these options in the future without affecting the client code much. However, its good as an additional protection level from dumb users. So my point is we can always make these types more strict later if we observe complains about it.

mykola-mokhnach:
Closed because of https://github.com/appium/java-client/pull/760

PR #760, opened 2017-11-03
Mykola mokhnach's actions params:  The addition to the #756

TikhomirovSergey:
## Change list

The addition to the #756
 
## Types of changes


TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget 
I finished. Could you take a look at  this one?
@SrinivasanTarget could you run iOS tests?

src/main/java/io/appium/java_client/TouchAction.java
```diff
@@ -78,36 +110,61 @@ public TouchAction press(int x, int y) {
...
     /**
      * Move current touch to center of an element.
      *
-     * @param el element to move to.
+     * @param moveToOptions see {@link AbsoluteOffsetOption}.
      * @return this TouchAction, for chaining.
      */
-    public TouchAction moveTo(WebElement el) {
-        ActionParameter action = new ActionParameter("moveTo", (HasIdentity) el);
+    public T moveTo(AbsoluteOffsetOption moveToOptions) {
```
mykola-mokhnach: moveTo does not support absolute offset.

src/main/java/io/appium/java_client/TouchAction.java
```diff
@@ -118,13 +175,11 @@ public TouchAction moveTo(WebElement el) {
...
      */
-    public TouchAction moveTo(int x, int y) {
-        ActionParameter action = new ActionParameter("moveTo");
-        action.addParameter("x", x);
-        action.addParameter("y", y);
-        parameterBuilder.add(action);
-        return this;
+    @Deprecated
+    public T moveTo(int x, int y) {
+        return moveTo(useAbsolute(x, y));
```
mykola-mokhnach: * @param x **change** in x coordinate to move through.
       * @param y **change** in y coordinate to move through.

src/main/java/io/appium/java_client/touch/AbsoluteOffsetOption.java
```diff
@@ -0,0 +1,68 @@
...
+    public AbsoluteOffsetOption withAbsoluteOffset(int xOffset, int yOffset) {
+        this.absoluteOffset = new Point(xOffset, yOffset);
+        //noinspection unchecked
+        return this;
+    }
+
+    @Override
+    protected void verify() {
+        ofNullable(absoluteOffset).orElseThrow(() -> new IllegalArgumentException(
+                "Absolute offset must not be defined"));
```
mykola-mokhnach: why must not?

src/main/java/io/appium/java_client/touch/OptionsCombinedWithOffset.java
```diff
@@ -0,0 +1,49 @@
...
+     * @return self-reference
+     */
+    public T withOffset(RelativeOffsetOption offset) {
+        offsetOption = offset;
+        return (T) this;
+    }
+
+    protected void verify() {
+        ofNullable(offsetOption).orElseThrow(() ->
+                new IllegalArgumentException("Some relative or absolute offset should "
```
mykola-mokhnach: ->Either relative or absolute

src/main/java/io/appium/java_client/touch/OptionsCombinedWithOffset.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param offset is the values of required offset from the left corner of an element.     *
+     * @return self-reference
+     */
+    public T withOffset(RelativeOffsetOption offset) {
+        offsetOption = offset;
+        return (T) this;
+    }
+
+    protected void verify() {
+        ofNullable(offsetOption).orElseThrow(() ->
```
mykola-mokhnach: we should consider here, that useRelative(x,y) won't work as expected for press/tap/longTap. I think the best way to avoid it would be to extract element parameter from RelativeOffset into _withElement_ like we discussed before

src/main/java/io/appium/java_client/touch/RelativeOffsetOption.java
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     *  * It creates an instance of {@link RelativeOffsetOption } which takes an element
+     * and 0-value x and y offsets.
+     *
+     * @param element is the target element
+     * @return a built option
+     */
+    public static RelativeOffsetOption useRelative(WebElement element) {
+        return useRelative(element, 0, 0);
```
mykola-mokhnach: `0,0` should not be the case, but separate withElement should address this problem.

src/main/java/io/appium/java_client/touch/OptionsCombinedWithOffset.java
mykola-mokhnach: the main reason of this is that we anyway translate both relative and absolute offset into `x,y` JSON keys, so there is no way for the server do distinguish them (unless we start using [W3C JSON](https://github.com/jlipps/simple-wd-spec#perform-actions) format to represent the stuff, which is already awaiting in the queue).

src/main/java/io/appium/java_client/touch/RelativeOffsetOption.java
```diff
@@ -0,0 +1,119 @@
...
+     * @param yOffset the relative distance from the top element corner
+     *                (if set) or from the top corner of the preceding chain action.
+     *                This value might be zero if it is necessary.
+     * @return this instance for chaining.
+     */
+    public RelativeOffsetOption withRelativeOffset(WebElement element, int xOffset, int yOffset) {
+        checkNotNull(element);
+        this.elementId = ((HasIdentity) element).getId();
+        this.relativeOffset = new Point(xOffset, yOffset);
+        //noinspection unchecked
```
mykola-mokhnach: I assume this comment is not needed here

mykola-mokhnach: same below

src/main/java/io/appium/java_client/touch/RelativeOffsetOption.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Override
+    public Map<String, Object> build() {
+        final Map<String, Object> result = super.build();
+        ofNullable(relativeOffset).ifPresent(point -> {
+            result.put("x", point.x);
+            result.put("y", point.y);
+        });
+
```
mykola-mokhnach: the empty line is not needed

src/main/java/io/appium/java_client/touch/WaitOptions.java
```diff
@@ -0,0 +1,67 @@
...
+        checkNotNull(duration);
+        checkArgument(duration.toMillis() >= 0,
+                "Duration value should be greater or equal to zero");
+        this.duration = duration;
+        return this;
+    }
+
+    @Override
+    protected void verify() {
+        ofNullable(duration).orElseThrow(() ->
```
mykola-mokhnach: this verification is redundant, since we already have _checkNotNull_ verification inside _withDuration_ method

src/test/java/io/appium/java_client/touch/FailsWithMatcher.java
```diff
@@ -0,0 +1,43 @@
+package io.appium.java_client.touch;
```
mykola-mokhnach: This stuff is special for tests, shouldn't it be in `tests` folder then?

src/test/java/io/appium/java_client/touch/DummyElement.java
```diff
@@ -0,0 +1,103 @@
+package io.appium.java_client.touch;
```
mykola-mokhnach: shouldn't this be in tests?

src/test/java/io/appium/java_client/touch/TouchOptionsTests.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Map;
+
+public class TouchOptionsTests {
+    private static final WebElement DUMMY_ELEMENT = new DummyElement();
+
+    @Test
+    public void invalidAbsolutePositionOptionsShouldFailOnBuild() throws Exception {
+        final List<ActionOptions> invalidOptions = new ArrayList<>();
+        invalidOptions.add(new AbsoluteOffsetOption());
+        for (ActionOptions opts : invalidOptions) {
```
mykola-mokhnach: does it make sense to create an array with a single item?

src/test/java/io/appium/java_client/touch/TouchOptionsTests.java
```diff
@@ -0,0 +1,98 @@
...
+            assertThat(opts::build, failsWith(IllegalArgumentException.class));
+        }
+    }
+
+    @Test
+    public void invalidRelativePositionOptionsShouldFailOnBuild() throws Exception {
+        final List<ActionOptions> invalidOptions = new ArrayList<>();
+        invalidOptions.add(new RelativeOffsetOption());
+        for (ActionOptions opts : invalidOptions) {
```
mykola-mokhnach: same here

src/test/java/io/appium/java_client/touch/TouchOptionsTests.java
```diff
@@ -24,34 +25,28 @@
 public class TouchOptionsTests {
     private static final WebElement DUMMY_ELEMENT = new DummyElement();
 
-    @Test
+    @Test(expected = IllegalArgumentException.class)
```
mykola-mokhnach: I didn't know about such possibility %)

src/main/java/io/appium/java_client/TouchAction.java
```diff
@@ -177,49 +227,91 @@ public TouchAction tap(int x, int y) {
...
+        return (T) this;
+    }
+
+    /**
+     * Press and hold the at the center of an element until the context menu event has fired.
+     *
+     * @param longPressOptions see {@link LongPressOptions}.
+     * @return this TouchAction, for chaining.
+     */
+    public T longPress(LongPressOptions longPressOptions) {
```
mykola-mokhnach: do we still need this method?

src/main/java/io/appium/java_client/touch/PositionOffsetOption.java
```diff
@@ -0,0 +1,51 @@
...
+     * @param xOffset is x-axis offset.
+     * @param yOffset is y-axis offset.
+     * @return a built option
+     */
+    public static PositionOffsetOption positionOffsetOption(int xOffset, int yOffset) {
+        return new PositionOffsetOption().withOffet(xOffset, yOffset);
+    }
+
+    /**
+     * Sets the offset from the upper left corner of the screen/element or
```
mykola-mokhnach: I'd rather have two separate methods to set relative and absolute offset, so then we can distinguish these in _verify_ and show an error if the client wants, for example, use absolute coordinates with _moveTo_ call.

src/main/java/io/appium/java_client/touch/PositionOffsetOption.java
```diff
@@ -0,0 +1,51 @@
...
+     * from the corresponding action.
+     *
+     * @param xOffset is x-axis offset.
+     * @param yOffset is y-axis offset.
+     * @return this instance for chaining.
+     */
+    public T withOffet(int xOffset, int yOffset) {
```
mykola-mokhnach: O**ff**et

src/main/java/io/appium/java_client/touch/PositionOffsetOption.java
```diff
@@ -0,0 +1,51 @@
+package io.appium.java_client.touch;
+
+import static java.util.Optional.ofNullable;
+
+import org.openqa.selenium.Point;
+
+import java.util.Map;
+
+public class PositionOffsetOption<T extends PositionOffsetOption<T>> extends ActionOptions<T> {
+    private Point offset = new Point(0, 0);
```
mykola-mokhnach: like I mentioned before, the default value should not be set to `0,0`, but rather to null. When only element is set the the server code selects appropriate coordinates for touch and this is not always the top left corner of the element.

src/main/java/io/appium/java_client/touch/WebElementOption.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    /**
+     * This method creates a build instance of the {@link WebElementOption}
+     *
+     * @param element is the element to calculate offset from.
+     * @return the built option
+     */
+    public static WebElementOption elementOption(WebElement element) {
+        return elementOption(element, 0, 0);
```
mykola-mokhnach: should be no zeroes

src/main/java/io/appium/java_client/touch/WebElementOption.java
```diff
@@ -0,0 +1,66 @@
...
+     * from the upper left corner of the given element.
+     *
+     * @param element is the element to calculate offset from.
+     * @return self-reference
+     */
+    public WebElementOption withElement(WebElement element) {
+        checkNotNull(element);
+        checkArgument(true, "Element should be an instance of the class which "
+                + "implements org.openqa.selenium.internal.HasIdentity",
+                (HasIdentity.class.isAssignableFrom(element.getClass())));
```
mykola-mokhnach: `element instanceof HasIdentity` is enough

src/main/java/io/appium/java_client/touch/WebElementOption.java
```diff
@@ -0,0 +1,66 @@
...
+import static java.util.Optional.ofNullable;
+
+import org.openqa.selenium.WebElement;
+import org.openqa.selenium.internal.HasIdentity;
+
+import java.util.Map;
+
+public class WebElementOption extends PositionOffsetOption<WebElementOption> {
+
+    private HasIdentity elementId;
```
mykola-mokhnach: We could save only String id value here instead of the whole element

src/main/java/io/appium/java_client/touch/WebElementOption.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * This method creates a build instance of the {@link WebElementOption}
+     *
+     * @param element is the element to calculate offset from.
+     * @param x is the x-offset from the upper left corner of the given element.
+     * @param y is the y-offset from the upper left corner of the given element.
+     * @return the built option
+     */
+    public static WebElementOption elementOption(WebElement element, int x, int y) {
```
mykola-mokhnach: I'd rather say it is elementOffset. _Option_ suffix it's not needed - let's keep these names short

src/test/java/io/appium/java_client/android/AndroidTouchTest.java
```diff
@@ -31,8 +35,10 @@ public void setUp() throws Exception {
         WebElement dragText = driver.findElement(By.id("io.appium.android.apis:id/drag_text"));
         assertEquals("Drag text not empty", "", dragText.getText());
 
-        TouchAction dragNDrop =
-                new TouchAction(driver).longPress(dragDot1).moveTo(dragDot3).release();
+        TouchAction dragNDrop = new TouchAction(driver)
+                        .longPress(elementOption(dragDot1))
```
mykola-mokhnach: perhaps `withOffset(elementOption(...))`

src/main/java/io/appium/java_client/touch/WebElementOption.java
mykola-mokhnach: I'd rather rename these to simply `element`, `relative` and `absolute`, so we can pass it like

`withOffset(element(el, x, y)), withOffset(element(el)), withOffset(relative(x,y)), withOffset(absolute(x,y))`

TikhomirovSergey: @mykola-mokhnach 
I think it would be better to have two static methods
```java
element()
offset() //to not confuse user when he/she waits for an action with absolute offset but it acts with 
//relative or does the opposite. 
```

TikhomirovSergey: If there was a flag that could make action use offset as some relative/absolute value forcefully I think that it would have sense.

mykola-mokhnach: I think it's easier to perform the verification on the client side and throw an exception then if we explicitly define which type of offset we have - relative or absolute.

src/main/java/io/appium/java_client/TouchAction.java
TikhomirovSergey: @mykola-mokhnach 
Yes because it has the additional 
```
withDuration
```

TikhomirovSergey:
@mykola-mokhnach 
It seems I still don't understand something. 
First of all it would be good to understand the action flow and how do actions use offsets accordingly their position in the action sequence.
 Also which actions may use only absolute/relative offsets.

TikhomirovSergey:
@mykola-mokhnach
It seems I still don't understand something.
First of all it would be good to understand the action flow and how do actions use offsets accordingly their position in the action sequence.
Also which actions may use only absolute/relative offsets.

mykola-mokhnach:
The point is to make this implementation backward compatible to the existing one. And the existing one defines:
- tap, longPress, press: absolute coordinates (x, y; element; element, x, y)
- moveTo: relative coordinates (x, y) and partially absolute coordinates (element; element, x ,y)

src/main/java/io/appium/java_client/touch/offset/AbstractPositionOption.java
```diff
@@ -0,0 +1,18 @@
+package io.appium.java_client.touch.offset;
+
+import io.appium.java_client.touch.ActionOptions;
+import org.openqa.selenium.Point;
+
+public abstract class AbstractPositionOption<T extends AbstractPositionOption<T>> extends ActionOptions<T> {
+    protected Point offset;
```
mykola-mokhnach: = null

src/main/java/io/appium/java_client/touch/offset/PointOption.java
```diff
@@ -0,0 +1,56 @@
...
+import static java.lang.String.format;
+import static java.util.Optional.ofNullable;
+
+import org.openqa.selenium.Point;
+
+import java.util.Map;
+
+public class PointOption extends AbstractPositionOption<PointOption> {
+
+    private static final String ERROR_MESSAGE_TEMPLATE = "%s coordinate value should be equal or greater than zero";
```
mykola-mokhnach: I'd rather have it of type `Function<String, String>`

src/main/java/io/appium/java_client/touch/offset/PointOption.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * It defines x and y coordinates.
+     * This is offset from the upper left corner of the screen.
+     *
+     * @param xOffset is x value.
+     * @param yOffset is y value.
+     * @return self-reference
+     */
+    @Override
+    public PointOption position(int xOffset, int yOffset) {
```
mykola-mokhnach: why not AbsoluteOffsetOption ?

src/test/java/io/appium/java_client/android/AndroidTouchTest.java
```diff
@@ -97,9 +100,10 @@ public void setUp() throws Exception {
 
         TouchAction dragNDrop = new TouchAction(driver)
                 .longPress(longPressOptions()
-                        .withOffset(offset(center1.x, center1.y))
+                        .withPosition(coordinates(center1.x, center1.y))
                         .withDuration(ofSeconds(2)))
-                .moveTo(offset(center2.x, center2.y))
+                .moveTo(position()
+                        .withPosition(coordinates(center2.x, center2.y)))
```
mykola-mokhnach: how this expression visually helps to understand, that this position is calculated relatively to the previous element and the previous one is absolute?

TikhomirovSergey: @mykola-mokhnach 
The signature of methods prevents the setting of the wrong offset. Like 
```java
public T tap(Position<PointOption> tapOptions) 
```
It means that user may set only absolute offset (`PointOption`, `coordinates(x, y)`) or element-option there. 

You said before
>moveTo: relative coordinates (x, y) and partially absolute coordinates (element; element, x ,y)

It means that `moveTo` is not bounded by absolute/relative/element offsets formally.  So I desided to 

```java
public <S extends AbstractPositionOption> T moveTo(Position<S> moveToOptions) {
```

and it is possible to

```java
moveTo(position().withOffset(coordinates()))
moveTo(position().withOffset(offset()))
moveTo(position().withOffset(element()))
```

I thought that one of ideas behind these changes was to minimize method count and exclude unnecessary overloading/code repeating. But, yes, it should be safe and it should exclude accidental user's actions.

If not everything is ok with this approach then I think that we could only split `moveTo` into
```java
moveTo(absolute)
moveTo(relative)
```
but it is still not clear when each method should be invoked.

TikhomirovSergey: May be there is some problem with the method naming

mykola-mokhnach: I like the idea in general. What would you say if we don't split parameter signatures between move/non-moveTo, but just add a possibility to throw a runtime verification exception if withOffset(coordinates()) is provided  to moveTo and/or withOffset(offset())) is provided to non-moveTo method?

src/main/java/io/appium/java_client/touch/offset/PointOption.java
TikhomirovSergey: I think it is not completely correct. When we say `absolute offset` it means `coordinates` or 'point' on the screen.

src/test/java/io/appium/java_client/android/AndroidTouchTest.java
TikhomirovSergey: @mykola-mokhnach I think that the good way is to inform users by method signature and compilation errors. I suppose users won't have fun if compiled will throw runtime verification errors.

The `Position` class has at least two purposes:
- unify signature of methods. Most of touch actions (present and further) can use either coordinate position/offset or offset from some element. So the class has 2 methods `withOffset` and `withElement` which invocation replaces previously stored value.

- bound used coordinate strategy (absolute or relative) by generic parameter. 

Currently most of actions use only points/elements, but I think there will be new actions which will use only  offsets/elements.

mykola-mokhnach: If something breaks the spec then it should throw an error. However I see there is a major issue in the server implementation. iOS threats moveTo arguments as relative and Android as absolute.  Then it why such expression

```
       TouchAction dragNDrop = new TouchAction(driver)
                .longPress(longPressOptions()
                        .withPosition(coordinates(center1.x, center1.y))
                        .withDuration(ofSeconds(2)))
                .moveTo(position()
                        .withPosition(coordinates(center2.x, center2.y)))
                .release();
```

works perfectly for Android, but causes the unexpected behaviour for iOS

SrinivasanTarget: @mykola-mokhnach Yes i remember i had an issue open on this at server end reg. `moveTo`

TikhomirovSergey: Ok. I need to think a bit. 

One of the possible ways could be the ovelloading of such methods by iOS/Androind/Windows (if it will be added), maybe the making of TouchAction the abstract class with generic signature of such merhods, overloaded with bounds by sublasses .

SrinivasanTarget: https://github.com/appium/appium/issues/7486

TikhomirovSergey:
@mykola-mokhnach @SrinivasanTarget 
There is one more change. It was done according our decision to use only absolute coordinates or an offset from an element. Please review the last change.

src/main/java/io/appium/java_client/touch/offset/ElementOption.java
```diff
@@ -37,6 +39,19 @@ public static ElementOption element(WebElement element) {
     }
 
     /**
+     * It defines x and y offset from the upper left corner of an element.
+     *
+     * @param xOffset is x value.
```
mykola-mokhnach: I'd rather say `the x/y offset from the top left point of the screen`

src/main/java/io/appium/java_client/touch/offset/ElementOption.java
```diff
@@ -37,6 +39,19 @@ public static ElementOption element(WebElement element) {
     }
 
     /**
+     * It defines x and y offset from the upper left corner of an element.
```
mykola-mokhnach: not element, but screen

mykola-mokhnach: of an element (if provided) of the whole screen (if the element is not provided)

src/main/java/io/appium/java_client/touch/offset/PointOption.java
```diff
@@ -4,11 +4,14 @@
...
 
+import io.appium.java_client.touch.ActionOptions;
 import org.openqa.selenium.Point;
 
 import java.util.Map;
 
-public class PointOption extends AbstractPositionOption<PointOption> {
+public class PointOption<T extends PointOption<T>> extends ActionOptions<T> {
+
+    protected Point coordinates;
```
mykola-mokhnach: = null

src/main/java/io/appium/java_client/touch/offset/ElementOption.java
```diff
@@ -37,6 +39,19 @@ public static ElementOption element(WebElement element) {
...
+     *
+     * @param xOffset is x value.
+     * @param yOffset is y value.
+     * @return self-reference
+     */
+    @Override
+    public ElementOption coordinates(int xOffset, int yOffset) {
```
mykola-mokhnach: should it be `withCoordinates` to follow the common style?

src/main/java/io/appium/java_client/touch/offset/PointOption.java
```diff
@@ -0,0 +1,58 @@
...
+     * This is offset from the upper left corner of the screen.
+     *
+     * @param xOffset is x value.
+     * @param yOffset is y value.
+     * @return self-reference
+     */
+    public T coordinates(int xOffset, int yOffset) {
+        checkArgument(xOffset >= 0, format(ERROR_MESSAGE_TEMPLATE, "X"));
+        checkArgument(yOffset >= 0, format(ERROR_MESSAGE_TEMPLATE, "Y"));
+        coordinates = new Point(xOffset, yOffset);
```
SrinivasanTarget: `withCoordinates` as mykola suggested above.

src/test/java/io/appium/java_client/touch/FailsWithMatcher.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public static <E extends Throwable> Matcher<Runnable> failsWith(
+            final Class<E> throwableType) {
+        return new FailsWithMatcher<>(instanceOf(throwableType));
+    }
+
+    public static <E extends Throwable> Matcher<Runnable> failsWith(
+            final Class<E> throwableType, final Matcher<? super E> throwableMatcher) {
+        return new FailsWithMatcher<>(allOf(instanceOf(throwableType), throwableMatcher));
```
SrinivasanTarget: Also compilation error here

TikhomirovSergey: ??? ок

TikhomirovSergey: @SrinivasanTarget 
I can't reproduce it even after removal of the project from the disc and cloning it again. 
Could you try again? If you are facing the same issue could you provide some text of the error.

TikhomirovSergey:
I am waiting for @SrinivasanTarget 's approve.

src/test/java/io/appium/java_client/touch/FailsWithMatcher.java
TikhomirovSergey: @SrinivasanTarget even Travis could compile the project

SrinivasanTarget: Yes travis wasn't able to catch it. Only IDE throws it on file level but still project builds fine. `no instance of type variable exists so that E conforms to capture of ? super....` is error.

SrinivasanTarget: looks fine in latest IDE

mykola-mokhnach: Do you use Eclipse?

SrinivasanTarget: Intellij IDEA 2017.1.5 throws this error but no errors in Intellij IDEA 2017.2

Issue #764, opened 2017-11-10
Client asking to many times for session


vrunoa:
## Description

I'm seeing the java client asking for getSession too many times, probably like 5,6 times between commands, making the test to run slower. This make a big difference when running on a testing cloud since this calls are adding ~4 seconds between commands in the test. 

## Environment
This is a problem with the client itself. 
* java client build version or git revision if you use some shapshot: 5.0.4 (havent tried another version)
* Appium server version or git revision if you use some shapshot: Any
* Desktop OS/version used to run Appium if necessary: Any
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: Any
* Mobile platform/version under test: Any
* Real device or emulator/simulator: Any

## Code To Reproduce Issue [ Good To Have ]

Running a simple test using the java client, during logs we can see all the getSession commands. 
For ex.
```
INFO: Executed: [f8b4312dda7c43bfba2620ef50605eec, findElements {using=id, value=com.test.app:id/text1}]
Nov 09, 2017 5:39:45 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executing: getSession [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:45 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executed: [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:45 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executing: getSession [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:46 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executed: [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:46 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executing: getSession [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:46 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executed: [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
Nov 09, 2017 5:39:46 PM org.openqa.selenium.remote.RemoteWebDriver log
INFO: Executing: getSession [f8b4312dda7c43bfba2620ef50605eec, getSession {}]
[...]

PR #765, opened 2017-11-10
Cache session details to avoid making unnecessary queries

mykola-mokhnach:
## Change list

Addresses https://github.com/appium/java-client/issues/764
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
+++ b/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
@@ -38,7 +38,7 @@ import static org.apache.commons.lang3.StringUtils.isBlank;
 abstract class DefaultGenericMobileDriver<T extends WebElement> extends RemoteWebDriver
     implements MobileDriver<T> {
 
-    private volatile ImmutableMap sessionDetails = null;
+    private ImmutableMap sessionDetails = null;
 
     public DefaultGenericMobileDriver(CommandExecutor executor, Capabilities desiredCapabilities) {
         super(executor, desiredCapabilities);
@@ -163,7 +163,7 @@ abstract class DefaultGenericMobileDriver<T extends WebElement> extends RemoteWe
      */
     @SuppressWarnings("unchecked")
     @Override
-    public Map<String, Object> getSessionDetails() {
+    public synchronized Map<String, Object> getSessionDetails() {
         if (sessionDetails == null) {
             Response response = execute(GET_SESSION);
             Map<String, Object> resultMap = Map.class.cast(response.getValue());
--- a/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
+++ b/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
@@ -30,9 +30,15 @@ import org.openqa.selenium.remote.Response;
 import java.util.List;
 import java.util.Map;
 
+import static io.appium.java_client.MobileCommand.GET_SESSION;
+import static java.util.stream.Collectors.toMap;
+import static org.apache.commons.lang3.StringUtils.isBlank;
+
 @SuppressWarnings({"unchecked", "rawtypes"})
 abstract class DefaultGenericMobileDriver<T extends WebElement> extends RemoteWebDriver
     implements MobileDriver<T> {
+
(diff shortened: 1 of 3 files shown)


vrunoa:
Works great! reduced a test from 2hs to 25minutes.

saikrishna321:
Wow, that's a great time difference .....

saikrishna321:
@vrunoa was this tested on parallel sessions ? if not I can take it ...

vrunoa:
@saikrishna321 no, just a test on a cloud service. Havent test parallel sessions.

saikrishna321:
done I will do it today

src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
```diff
@@ -151,6 +157,32 @@ public List findElementsByXPath(String using) {
         return super.getMouse();
     }
 
+    /**
+     * @return a map with values that hold session details.
+     *
+     */
+    @SuppressWarnings("unchecked")
+    @Override
+    public synchronized Map<String, Object> getSessionDetails() {
```
TikhomirovSergey: I am against this change. Session details data is not immutable. It may change when driver is switched to another context (NATIVE APP or WEB VIEW). 

I think the right way is to find places where it is invoked frequently.

TikhomirovSergey:
@vrunoa There are some question to make it clear:
- do you use get session details frequently at the client side?
- do you use page objects?

vrunoa:
@TikhomirovSergey the test I'm running is not calling to getSessionDetails. 
If the session will change at some point, should be getSessionDetails only at that point and not every time ?

Not sure about page object, can you help me figure out if it is used.

saikrishna321:
@vrunoa as my tests are running locally, i don't see much time difference its 2mins difference in total

vrunoa:
@saikrishna321 yeah, the big difference it's when running on a cloud service since those getSessionDetails calls take like 5 seconds or more between commands but locally you just lose a few milliseconds. 
@TikhomirovSergey do we know where are session details changed ?

TikhomirovSergey:
@vrunoa I suppose we don't all the actions which can change session parameters.
I know one `driver.context(someContext)`

>If the session will change at some point, should be getSessionDetails only at that point and not every time ?

We know only one place in java_client where `getSessionDetails` is asked very often. And I am going to optimize it. I have asked the question before. Does your code build and use page objects using `@AndroidFindBy` /`@iOSFindBy` etc and `AppiumFieldDecorator`? Otherwise could you show the sample of your code which has the performance issue?

vrunoa:
@TikhomirovSergey yes, it does use those decorators. Sorry but I cannot share the source code of this particular test.

TikhomirovSergey:
@vrunoa ok. So it seems I was right. Working on improvements.

vrunoa:
thanks @TikhomirovSergey. Let me know if there's something I can help with.

vrunoa:
@TikhomirovSergey Sorry to bother, but do you have any updates on this fix?

TikhomirovSergey:
@vrunoa Yep. Please look at the related issue #764. There are 3 commits. Tomorrow I am going to propose the new PR.

mykola-mokhnach:
Closed in favour of https://github.com/appium/java-client/pull/769

src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
fhoeben: If full copy is used, shouldn't there at least be an 'expiry time' or a method to force the details to be refreshed?

When I ran into this issue I decided to use an expiry time, based on the time it took to retrieve the information, in my (not so nice) workaround in https://github.com/fhoeben/hsac-fitnesse-appium.
The approach I currently use there is to have some cache at the element level, and using [my own 'json to element' converter](https://github.com/fhoeben/hsac-fitnesse-appium/blob/master/src/main/java/nl/hsac/fitnesse/fixture/util/mobile/element/MobileElementConverter.java). 
My custom elements require less retrieval of session details (for me the 'foundBy' information does not require session details, which is currently required for the driver's `toString()`, and they 'cache' information retrieved from the driver about an element for a short while (so my test code does not need to worry too much about for instance checking `isDisplayed()` twice)).

Issue #102, opened 2014-09-08
Inconsistent use of MissingParameterException


0x1mason:
java-client checks on the client side for missing parameters in MultiTouchAction::perform and throws a custom exception.  However, there aren't other checks for valid params.  We should check parameters consistently in the client or not at all.

Also, we should probably use `Preconditions.checkArgument(foo, msg);`, since this is part of the Guava idiom, and add `throws IllegalArgumentException` to function signatures that assert precondtions.

TikhomirovSergey:
Fixed

PR #772, opened 2017-11-23
#102 FIX

TikhomirovSergey:
## Change list
- MissingParameterException was removed
- some code improvements in MultiTouchAction. Now it will throw `IllegalArgumentException` when no action is defined.
 
#102

## Types of changes


mykola-mokhnach:
+ 💯 for picking up issues that have been waiting for three years in the backlog 👍

src/main/java/io/appium/java_client/MultiTouchAction.java
```diff
@@ -62,16 +65,14 @@ public MultiTouchAction add(TouchAction action) {
...
-        if (size > 1) {
+        List<TouchAction> touchActions = actions.build();
+        checkArgument(touchActions.size() > 0,
+                "MultiTouch action must have at least one TouchAction added before it can be performed");
+        if (touchActions.size() > 1) {
             performsTouchActions.performMultiTouchAction(this);
-        } else if (size == 1) {
+        } else {
             //android doesn't like having multi-touch actions with only a single TouchAction...
             performsTouchActions.performTouchAction(actions.build().get(0));
```
mykola-mokhnach: perhaps `touchActions.get(0)` ?

TikhomirovSergey: yes. Will improve soon

src/main/java/io/appium/java_client/MultiTouchAction.java
```diff
@@ -62,16 +65,14 @@ public MultiTouchAction add(TouchAction action) {
      * Perform the multi-touch action on the mobile performsTouchActions.
      */
     public MultiTouchAction perform() {
-        int size = actions.build().size();
-        if (size > 1) {
+        List<TouchAction> touchActions = actions.build();
+        checkArgument(touchActions.size() > 0,
+                "MultiTouch action must have at least one TouchAction added before it can be performed");
+        if (touchActions.size() > 1) {
             performsTouchActions.performMultiTouchAction(this);
```
mykola-mokhnach: you can `returns this` immediately here, so then `else` block is not needed

src/main/java/io/appium/java_client/MultiTouchAction.java
```diff
@@ -80,9 +81,8 @@ public MultiTouchAction perform() {
         ImmutableList.Builder<Object> listOfActionChains = ImmutableList.builder();
         ImmutableList<TouchAction> touchActions = actions.build();
 
-        touchActions.forEach(action -> {
-            listOfActionChains.add(action.getParameters().get("actions"));
-        });
+        touchActions.forEach(action ->
```
mykola-mokhnach: why not to replace it with something like `ImmutableList.Builder<Object> listOfActionChains = ImmutableList.copyOf(touchActions.stream().map(x -> x.getParameters().get("actions")));` ?

src/main/java/io/appium/java_client/MultiTouchAction.java
TikhomirovSergey: @mykola-mokhnach 
I can't understand how to do it
```java
performsTouchActions.performMultiTouchAction(this);
```
is `void`

```java
performsTouchActions.performTouchAction(touchActions.get(0));
```
returns `TouchAction`

TikhomirovSergey:
@mykola-mokhnach There is one more change.
Yes. There is some legacy (from Java 7 times) code which still can be here. Sometimes it is hard to see it because of force of habit :)
@SrinivasanTarget your rewiev is is also valuable.

src/main/java/io/appium/java_client/MultiTouchAction.java
mykola-mokhnach:
```java
        if (touchActions.size() > 1) {
            performsTouchActions.performMultiTouchAction(this);
            resturn this;
        }
        //android doesn't like having multi-touch actions with only a single TouchAction...
        performsTouchActions.performTouchAction(touchActions.get(0));
        return this;
```

PR #795, opened 2017-12-19
Less strict verification

mykola-mokhnach:
## Change list

`moveTo` fix is not deployed to Appium yet, so it's OK to allow clients to still use relative coordinates. I've also added a note about the problem into the docstring.
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/touch/offset/PointOption.java
+++ b/src/main/java/io/appium/java_client/touch/offset/PointOption.java
@@ -1,7 +1,5 @@
 package io.appium.java_client.touch.offset;
 
-import static com.google.common.base.Preconditions.checkArgument;
-import static java.lang.String.format;
 import static java.util.Optional.ofNullable;
 
 import io.appium.java_client.touch.ActionOptions;
@@ -12,8 +10,6 @@ import java.util.Map;
 public class PointOption<T extends PointOption<T>> extends ActionOptions<T> {
 
     protected Point coordinates;
-
-    private static final String ERROR_MESSAGE_TEMPLATE = "%s coordinate value should be equal or greater than zero";
 
     /**
      * It creates a built instance of {@link PointOption} which takes x and y coordinates.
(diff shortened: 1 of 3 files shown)


src/main/java/io/appium/java_client/touch/offset/PointOption.java
```diff
@@ -36,14 +32,14 @@ public static PointOption point(int xOffset, int yOffset) {
      * @return self-reference
      */
     public T withCoordinates(int xOffset, int yOffset) {
-        checkArgument(xOffset >= 0, format(ERROR_MESSAGE_TEMPLATE, "X"));
```
TikhomirovSergey: Is it temporary change for backward compatibility with older servers versions?
I think it is not necessary to remove it. Just comment it, maybe with the TODO mark.
And before the publishing of the 6.0.0 we can remove `//`

src/test/java/io/appium/java_client/touch/TouchOptionsTests.java
```diff
@@ -44,8 +44,6 @@ public void invalidOptionsArgumentsShouldFailOnAltering() throws Exception {
         final List<Runnable> invalidOptions = new ArrayList<>();
         invalidOptions.add(() -> waitOptions(ofMillis(-1)));
         invalidOptions.add(() -> new ElementOption().withCoordinates(0, 0).withElement(null));
-        invalidOptions.add(() -> new PointOption().withCoordinates(0, -1));
```
TikhomirovSergey: Just comment I think

src/main/java/io/appium/java_client/touch/offset/PointOption.java
mykola-mokhnach: There might be a situation when it is necessary to provide negative absolute coordinates. For example, if the device has two screens and the second screen is located on the left side of the main one. That is why I'd consider this restriction as not very useful.

mykola-mokhnach: Also, it is possible that one wants to drag an element to an off-screen area and negative coordinates might also have sense for such case.

valfirst: I agree with @mykola-mokhnach. I tried using java-client 6.0.0-BETA1, but I was needed to rollback due to the new limitation: we have cases when actions start at the screen, but end out off screen.

TikhomirovSergey:
Ok

TikhomirovSergey:
Will publish it on this weekend

PR #799, opened 2018-01-05
Add handlers for lock/unlock in iOS

mykola-mokhnach:
## Change list

Please do not merge it unless https://github.com/appium/appium-xcuitest-driver/pull/595 is published
 
## Types of changes


src/main/java/io/appium/java_client/ios/LocksIOSDevice.java
```diff
@@ -27,11 +29,30 @@
```
SrinivasanTarget: @mykola-mokhnach Code LGTM. Have few suggestions though,

* Can we deprecate both `LocksIOSDevice` and `LocksAndroidDevice`. Have `LocksDevice` interface instead. Currently based on this PR both android and iOS both does supports the API except this [method](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/android/LocksAndroidDevice.java#L42) in android doesn't accept duration. May be a server fix will make the API work like iOS.

* https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/android/AndroidMobileCommandHelper.java#L182 can be removed since `MobileCommand` will have it.

* `isDeviceLocked` as like in this PR makes more sense to me instead of `isLocked` as in android now. So we  might need to deprecate that and rename in android to make it consistent.

Your thoughts?

mykola-mokhnach: > Can we deprecate both LocksIOSDevice and LocksAndroidDevice. Have LocksDevice interface instead. Currently based on this PR both android and iOS both does supports the API except this method in android doesn't accept duration. May be a server fix will make the API work like iOS.

Done

> https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/android/AndroidMobileCommandHelper.java#L182 can be removed since MobileCommand will have it.

Done

> isDeviceLocked as like in this PR makes more sense to me instead of isLocked as in android now. So we might need to deprecate that and rename in android to make it consistent.

Done

TikhomirovSergey:
@mykola-mokhnach It is loking good. I will try it on Android soon

mykola-mokhnach:
@TikhomirovSergey I've added integration tests for iOS. The isLocked method has been marked as deprecated.

PR #800, opened 2018-01-07
"deviceOrientation" MobileCapabilityType

MirzaSB:
## Change list

Added a capability type for "deviceOrientation". Primarily used in mobile device capabilities and remote drivers like Sauce Labs as well.
 
## Types of changes

## Details

Just added a MobileCapabilityType constant value.

### DIFF
--- a/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
+++ b/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
@@ -45,6 +45,11 @@ public interface MobileCapabilityType extends CapabilityType {
      * The kind of mobile device or emulator to use.
      */
     String DEVICE_NAME = "deviceName";
+    
+    /**
+     * The device orientation.
+     */
+    String DEVICE_ORIENTATION = "deviceOrientation";
 
     /**
      * How long (in seconds) Appium will wait for a new command from the


src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
```diff
@@ -45,6 +45,11 @@
      * The kind of mobile device or emulator to use.
      */
     String DEVICE_NAME = "deviceName";
+    
+    /**
+     * The device orientation.
+     */
+    String DEVICE_ORIENTATION = "deviceOrientation";
```
SrinivasanTarget: How is it different from `orientation` that we already have [here](https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java#L93)? I couldn't find one either in server doc's.

MirzaSB: For example, Sauce uses "deviceOrientation".  Please have a look at the following screenshot:

image

SrinivasanTarget: @MirzaSB We haven't added vendor specific capabilities here so far. Here since we already have `Orientation` it will add confusion regardless of documentation. So two possibilities, adding it with proper java doc comments and other option is to leave it as is. leaving it to @TikhomirovSergey

TikhomirovSergey: I am not sure that it is the good practice to add and support vendor-specific capabilities to the general project.
@MirzaSB There is nothing prevents you to add the local constant tp your project and use it.

MirzaSB:
Ok, I will close the PR.

PR #809, opened 2018-01-15
Use standard Java plugin source sets

valfirst:
## Change list

The Java plugin defines two standard source sets, called `main` and `test`. There is no reason to duplicate their declaration. (https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_source_sets)
 
## Types of changes


TikhomirovSergey:
@valfirst Will have a look ASAP

PR #812, opened 2018-01-17
Fix the second argument type for pushFile

mykola-mokhnach:
## Change list

Fixes the second argument type for `pushFile` endpoint. Appium expects it to be base64-encoded string, but we send array of bytes instead.
 
## Types of changes

## Details

See https://github.com/appium/appium-xcuitest-driver/pull/605

### DIFF
--- a/src/main/java/io/appium/java_client/MobileCommand.java
+++ b/src/main/java/io/appium/java_client/MobileCommand.java
@@ -377,11 +377,13 @@ public class MobileCommand {
      */
     public static Map.Entry<String, Map<String, ?>> pushFileCommand(String remotePath, byte[] base64Data) {
         String[] parameters = new String[] {"path", "data"};
+        String data;
         try {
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/MobileCommand.java
```diff
@@ -376,7 +377,13 @@ public static AppiumCommandInfo deleteC(String url) {
      */
     public static Map.Entry<String, Map<String, ?>> pushFileCommand(String remotePath, byte[] base64Data) {
         String[] parameters = new String[] {"path", "data"};
-        Object[] values = new Object[] {remotePath, base64Data};
+        String data;
+        try {
+            data = new String(base64Data, "UTF-8");
```
valfirst:
```java 
import java.nio.charset.StandardCharsets;
...
data = new String(base64Data, StandardCharsets.UTF_8)
```
?

mykola-mokhnach: np

PR #855, opened 2018-03-28
Add clipboard handlers

mykola-mokhnach:
## Change list

this adds possibility to set and get clipboard content for Android and iOS
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/ios/HasIOSClipboard.java
+++ b/src/main/java/io/appium/java_client/ios/HasIOSClipboard.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import io.appium.java_client.clipboard.ClipboardContentType;
 import io.appium.java_client.clipboard.HasClipboard;
 
-import javax.imageio.ImageIO;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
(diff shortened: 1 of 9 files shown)


mykola-mokhnach:
@KazuCocoa FYI

src/main/java/io/appium/java_client/ios/HasIOSClipboard.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /**
+     * Set an URL to the clipboard.
+     *
+     * @param url the actual URL to set.
+     */
+    default void setClipboardUrl(URL url) {
+        setClipboard(ClipboardContentType.URL, Base64
+                .getEncoder()
```
SrinivasanTarget: `getURLEncoder()` & `getURLDecoder()` should be good than basic encoding/decoding as it is not URL safe. Also for text encoding, i would recommend to use MIME encoding over basic encoding. what say @mykola-mokhnach ?

SrinivasanTarget:
waiting for drivers to get published. will play around after that.

src/main/java/io/appium/java_client/ios/HasIOSClipboard.java
mykola-mokhnach: It looks like this is not the case for Apple's NSData class, since it requires the data to be always padded: https://stackoverflow.com/questions/21406482/nsdata-wont-accept-valid-base64-encoded-string

TikhomirovSergey:
@mykola-mokhnach Will review today

SrinivasanTarget:
https://github.com/appium/WebDriverAgent/pull/70

Issue #846, opened 2018-03-09
Update to Selenium >= 3.10.x


TikhomirovSergey:
## Description

It needs to update the dependency on selenium to 3.10.0 or later version. But it has some major changes. So it needs for some efforts.

## Details

It seems that Selenium team decided to use `java.util.time.Duration` for the explicit waiting instead of  `java.util.conurent.TimeUnit` and `long` value. Current functionality should be refactored.

Related to: #742

joerg1985:
it looks like they are switching from ApacheHttpClient to OkHttpClient, see commit: https://github.com/SeleniumHQ/selenium/commit/a731205f3ff342e2ca7fcaf036200878163ea8d7

perhapse it makes sense for appium to switch to OkHttpClient too?

Issue #742, opened 2017-09-30
Re-design of the TimeOutDuration and WithTimeout. Use java.time.Duration instead of java.util.concurrent.TimeUnit


TikhomirovSergey:
## Description

- mark old methods and parameters of the annotation `@Deprecated`

- add get/set methods to the TimeOutDuration. These methods should get and set an instance of 'java.time.Duration'

- WithTimeout should have a parameter of the `java.time.temporal.ChronoUnit`

## Details

Java 8 is more powerful than 7. Also I think it is more correct way to use the native and general `java.time` package instead of more specifical `java.util.concurrent` in our case.

TikhomirovSergey:
Is going to be published soon.

PR #863, opened 2018-04-08
 #846 and #742 FIX

TikhomirovSergey:
## Change list

- FIX of #846. Update to Selenium 3.11
- #846 and #742 fixes. Migration to java.time.Duration
 
## Types of changes

## Details
I am sorry for the long break. Was busy.

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -66,7 +66,7 @@ class AppiumElementLocator implements CacheableLocator {
      * @deprecated This constructor is going to be removed. Use {@link #AppiumElementLocator(SearchContext, By,
      * boolean, Duration)} instead.
      */
-
+    @Deprecated
     public AppiumElementLocator(SearchContext searchContext, By by, boolean shouldCache,
         TimeOutDuration duration) {
         this(searchContext, by, shouldCache,
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -21,6 +21,8 @@ import static io.appium.java_client.pagefactory.ThrowableUtil.isInvalidSelectorR
 import static io.appium.java_client.pagefactory.ThrowableUtil.isStaleElementReferenceException;
 import static io.appium.java_client.pagefactory.utils.WebDriverUnpackUtility.getCurrentContentType;
 import static java.lang.String.format;
+import static java.time.Duration.ofMillis;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import io.appium.java_client.pagefactory.bys.ContentMappedBy;
 import io.appium.java_client.pagefactory.locator.CacheableLocator;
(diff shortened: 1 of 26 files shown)


src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -59,11 +62,31 @@
      * @param by                a By locator strategy
      * @param shouldCache       is the flag that signalizes that elements which
      *                          are found once should be cached
-     * @param duration          is a POJO which contains timeout parameters for the element to be searched
+     * @param duration          is a POJO which contains timeout parameters for the element to be found
+     * @deprecated This constructor is going to be removed. Use {@link #AppiumElementLocator(SearchContext, By,
```
mykola-mokhnach: I 'd also add `@Deprecated` annotation to the constructor itself

src/main/java/io/appium/java_client/pagefactory/AppiumFieldDecorator.java
```diff
@@ -62,16 +66,20 @@
     private static final List<Class<? extends WebElement>> availableElementClasses = ImmutableList.of(WebElement.class,
             RemoteWebElement.class, MobileElement.class, AndroidElement.class,
             IOSElement.class, WindowsElement.class);
+    public static Duration DEFAULT_WAITING_TIMEOUT = ofSeconds(1);
```
mykola-mokhnach: final?

src/main/java/io/appium/java_client/pagefactory/WithTimeout.java
```diff
@@ -38,7 +45,28 @@
...
+    ChronoUnit chronoUnit() default FOREVER;
+
+    class DurationBuilder {
+        static Duration build(WithTimeout withTimeout) {
+            //providing backward compatibility
+            if (!FOREVER.equals(withTimeout.chronoUnit())) {
+                return Duration.of(withTimeout.time(), withTimeout.chronoUnit());
+            }
+
+            return Duration.of(MILLISECONDS.convert(withTimeout.time(), withTimeout.unit()), MILLIS);
```
mykola-mokhnach: is FOREVER the same as milliseconds?

src/test/java/io/appium/java_client/android/BaseAndroidTest.java
```diff
@@ -48,7 +48,7 @@
         File appDir = new File("src/test/java/io/appium/java_client");
         File app = new File(appDir, "ApiDemos-debug.apk");
         DesiredCapabilities capabilities = new DesiredCapabilities();
-        capabilities.setCapability(MobileCapabilityType.AUTOMATION_NAME, AutomationName.ANDROID_UIAUTOMATOR2);
+        capabilities.setCapability(MobileCapabilityType.AUTOMATION_NAME, AutomationName.APPIUM);
```
mykola-mokhnach: why this is changed?

src/main/java/io/appium/java_client/pagefactory/WithTimeout.java
TikhomirovSergey: I was forced to add defalt values (FOREVER and NANOSECONDS) for bacward compatibility (temporary). Comilation problems will occure otherewise. User will be forced to set both time units.

TikhomirovSergey: @mykola-mokhnach However
Can we consider FOREVER illegal value? I think it has some sense.

mykola-mokhnach: I don't know, but it for sure does not help to read and understand the code.

PR #865, opened 2018-04-10
Refactor network connection setting on Android

mykola-mokhnach:
## Change list

With the previous implementation it was only possible to enable connections, but not disable. I think the current approach will make the feature much more flexible than it was before.
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/android/connection/ConnectionState.java
+++ b/src/main/java/io/appium/java_client/android/connection/ConnectionState.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client.android.connection;
+
+public class ConnectionState {
+    public static final int AIRPLANE_MODE_MASK = 0b001;
+    public static final int WIFI_MASK = 0b010;
+    public static final int DATA_MASK = 0b100;
+
+    private final int bitMask;
+
+    public int getBitMask() {
+        return bitMask;
+    }
+
+    public ConnectionState(int bitMask) {
+        this.bitMask = bitMask;
+    }
+
+    /**
+     * @return true if airplane mode is enabled.
+     */
+    public boolean isAirplaneModeEnabled() {
+        return (bitMask & AIRPLANE_MODE_MASK) != 0;
+    }
+
+    /**
+     * @return true if Wi-Fi connection is enabled.
+     */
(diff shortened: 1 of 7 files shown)


src/main/java/io/appium/java_client/android/connection/ConnectionStateBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+     * @return self instance for chaining
+     */
+    public ConnectionStateBuilder withAirplaneModeEnabled() {
+        this.bitMask = bitMask | AIRPLANE_MODE_MASK;
+        return this;
+    }
+
+    /**
+     * Sets airplane mode to disabled state if it was enabled.
+     * This option only works up to Android 6.
```
SrinivasanTarget: If i remember right, even data and wifi through set/get connection wasn't working with Android 7. Also i thought retrieving connection state was only possible through shell. may be i'm wrong. Let me try this today.

mykola-mokhnach: wifi should work. for data there is a comment, that it only works on emulator and on rooted devices

src/main/java/io/appium/java_client/android/connection/ConnectionState.java
```diff
@@ -0,0 +1,55 @@
...
+
+    public ConnectionState(long bitMask) {
+        this.bitMask = bitMask;
+    }
+
+    /**
+     * @return true if airplane mode is enabled.
+     */
+    public boolean isAirplaneModeEnabled() {
+        return (bitMask & AIRPLANE_MODE_MASK) != 0;
```
vmaxim: It seems that all these bitwise operations can be simplified via ```java.utils.BitSet```

mykola-mokhnach: do you find the current implementation not readable enough? I think BitSet usage might add an unnecessary overhead, since the actual count of bitwise operations here is pretty limited anyway

vmaxim: ok, np. However from my point of view `bs.set(AIRPLANE_MODE_BIT, false)` looks more naturaly than `this.bitMask = bitMask & ~AIRPLANE_MODE_MASK;` (Maybe I just don't like bitwise logic :)

src/test/java/io/appium/java_client/android/AndroidConnectionTest.java
```diff
@@ -26,23 +28,32 @@
...
-        assertEquals(Connection.NONE,
-            driver.getConnection());
-        driver.setConnection(Connection.AIRPLANE);
-        assertEquals(Connection.AIRPLANE,
-            driver.getConnection());
+        driver.setConnection(new ConnectionStateBuilder()
+                .withAirplaneModeDisabled()
+                .build());
+        ConnectionState state = driver.getConnection();
+        assertTrue(!state.isAirplaneModeEnabled() && !state.isWiFiEnabled() && !state.isDataEnabled());
```
SrinivasanTarget: check for wifi and data is not necessary i believe as disabling airplane mode will only disable airplane mode incase if it was already enabled and leave's the rest of connection as is. Isn't it?

Enabling airplane mode will disable both data and wifi but viseversa is not true. Correct me if i'm wrong.

SrinivasanTarget: Here also we might need to remove the additional checks.

mykola-mokhnach: These I'd prefer to keep

TikhomirovSergey:
Rewieving...

PR #869, opened 2018-04-13
Addition to #846 and #855:

TikhomirovSergey:
## Change list

- OkHttpClient instead of ApacheHttpClient
- fix of java.lang.IllegalArgumentException: Input byte array has incorrect ending byte at...
 
## Types of changes

## Details
I could reproduce `java.lang.IllegalArgumentException: Input byte array has incorrect ending byte at...` on the getting of clipboard text

### DIFF
--- a/src/main/java/io/appium/java_client/clipboard/HasClipboard.java
+++ b/src/main/java/io/appium/java_client/clipboard/HasClipboard.java
@@ -61,7 +61,7 @@ public interface HasClipboard extends ExecutesMethod {
      */
     default void setClipboardText(String text) {
         setClipboard(ClipboardContentType.PLAINTEXT, Base64
-                .getEncoder()
+                .getMimeEncoder()
                 .encode(text.getBytes(StandardCharsets.UTF_8)));
     }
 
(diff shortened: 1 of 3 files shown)


TikhomirovSergey:
@SrinivasanTarget Could you check it on iOS?

src/main/java/io/appium/java_client/clipboard/HasClipboard.java
```diff
@@ -72,7 +72,7 @@ default void setClipboardText(String text) {
      */
     default String getClipboardText() {
         byte[] base64decodedBytes = Base64
-                .getDecoder()
+                .getMimeDecoder()
```
SrinivasanTarget: Did suggested this in last PR. How did i missed it. Good catch @TikhomirovSergey

SrinivasanTarget: @mykola-mokhnach Last time when i locally built Appium's WDA(Forked) and tested clipboard of URL's where passing in iOS after your fix [here](https://github.com/appium/WebDriverAgent/pull/70) but now i upgraded my XCUI driver to latest [release](https://github.com/appium/appium-xcuitest-driver/blob/master/package.json#L11) but URL's stopped working after that. 

URL's are working only if i use mime decoding for URL.

SrinivasanTarget: Can you make mime decoding for getClipboardUrl too? Else it is failing...

TikhomirovSergey:
@SrinivasanTarget I made some changes. Could you run iOS test one more time?

PR #899, opened 2018-05-15
Send keys helper

mykola-mokhnach:
## Change list

This is for now poor-man's implementation of Android send keys support using W3C actions. Perhaps, the keys transformation algorithm might be improved by importing some routines from roboelectric lib, but I don't really want to include it into the dependencies list.
 
## Types of changes


mykola-mokhnach:
The test won't work until https://github.com/appium/appium-uiautomator2-server/pull/161 is published

mykola-mokhnach:
I think this can be still slightly improved on the server side

PR #961, opened 2018-06-29
update to selenium 3.13.0

SrinivasanTarget:
## Change list

update to selenium 3.13.0 
 
## Types of changes

@mykola-mokhnach ping.
Looks like checkstyle fixes broke the build in master. will fix it in sometime.


src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -96,6 +99,16 @@ public AppiumServiceBuilder() {
         withEnvironment(System.getenv());
     }
 
+    @Override public int score(Capabilities capabilites) {
```
mykola-mokhnach: What is the purpose of this method?

SrinivasanTarget: Grid sessions are going to be created only if the capability score is greater than zero. Higher the score higher the probability of getting sessions created sooner as per my chat with Simon. so we need this, probably used in future versions of grid.

mykola-mokhnach: It might be useful to add a docstring there with an explanation

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -96,6 +99,31 @@ public AppiumServiceBuilder() {
...
+     * Higher the score, higher the possibility of getting grid sessions created sooner.
+     */
+    @Override public int score(Capabilities capabilites) {
+        int score = 0;
+
+        if (capabilites.getCapability(PLATFORM_NAME) != null) {
+            score++;
+        }
+
+        String browserName = capabilites.getBrowserName();
```
mykola-mokhnach: may this ever return null?

mykola-mokhnach: ah, hm. It cannot : `return String.valueOf(Optional.ofNullable(getCapability("browserName")).orElse(""));`

PR #968, opened 2018-07-04
better handling of invocationtargetexception

SrinivasanTarget:
## Change list

Better handling of invocationtargetexception thrown during creating session
 
Thanks @iddol.

## Types of changes

@mykola-mokhnach FYI

### DIFF
--- a/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
+++ b/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
@@ -200,7 +200,7 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
                                 long.class.getSimpleName()), e);
                     } catch (InvocationTargetException e) {
                         throw new SessionNotCreatedException(
-                                format("Unable to create new remote session. desired capabilities = %s", desired), e);
+                                format("Unable to create new remote session. Desired capabilities: %s", desired), e);
                     }
                 } finally {
                     os.reset();
--- a/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
+++ b/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
@@ -199,10 +199,8 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
                                 InputStream.class.getSimpleName(),
                                 long.class.getSimpleName()), e);
                     } catch (InvocationTargetException e) {
-                        Throwable cause = e.getCause();
                         throw new SessionNotCreatedException(
-                                format("Unable to create new remote session. desired capabilities = %s"
-                                        + "and reason is " + cause.getMessage(), desired));
+                                format("Unable to create new remote session. desired capabilities = %s", desired), e);
                     }
                 } finally {
                     os.reset();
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -191,13 +191,18 @@ public Result createSession(HttpClient client, Command command)
...
                                         + "because 'createSession' which takes %s, %s and %s was not found "
                                         + "or it is not accessible",
                                 HttpClient.class.getSimpleName(),
                                 InputStream.class.getSimpleName(),
                                 long.class.getSimpleName()), e);
+                    } catch (InvocationTargetException e) {
+                        Throwable cause = e.getCause();
+                        throw new SessionNotCreatedException(
+                                format("Unable to create new remote session. desired capabilities = %s"
+                                        + "and reason is " + cause.getMessage(), desired));
```
mykola-mokhnach: why not to use format specifiers for both arguments?

mykola-mokhnach: Also, it might be handy to pass the original exception as `cause` constructor argument, so it would be easier to debug it.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -191,13 +191,18 @@ public Result createSession(HttpClient client, Command command)
...
-                    } catch (NoSuchMethodException | IllegalAccessException | InvocationTargetException e) {
+                    } catch (NoSuchMethodException | IllegalAccessException e) {
                         throw new WebDriverException(format("It is impossible to create a new session "
                                         + "because 'createSession' which takes %s, %s and %s was not found "
                                         + "or it is not accessible",
                                 HttpClient.class.getSimpleName(),
                                 InputStream.class.getSimpleName(),
                                 long.class.getSimpleName()), e);
+                    } catch (InvocationTargetException e) {
+                        Throwable cause = e.getCause();
```
mykola-mokhnach: can it be that getCause returns null?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
SrinivasanTarget:
```
Throwable cause = e.getCause();
                        throw new SessionNotCreatedException(
                                format("Unable to create new remote session. desired capabilities = %s"
                                        + "and reason is %s", desired, cause != null ? cause.getMessage() : null,
                                        e.getTargetException()));
```

@mykola-mokhnach Do you mean it this way?

mykola-mokhnach: almost. It's just instead of setting reason to null if cause is null you set it to the original exception instead

mykola-mokhnach: also SessionNotCreatedException has the second constructor where one can provide the cause:

`SessionNotCreatedException(String msg, Throwable cause)`

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -199,10 +199,8 @@ public Result createSession(HttpClient client, Command command)
                                 InputStream.class.getSimpleName(),
                                 long.class.getSimpleName()), e);
                     } catch (InvocationTargetException e) {
-                        Throwable cause = e.getCause();
                         throw new SessionNotCreatedException(
-                                format("Unable to create new remote session. desired capabilities = %s"
-                                        + "and reason is " + cause.getMessage(), desired));
+                                format("Unable to create new remote session. desired capabilities = %s", desired), e);
```
mykola-mokhnach: `desired capabilities = %s` -> `Desired capabilities: %s`

PR #997, opened 2018-08-13
Expand touch options API to accept coordinates as Point

valfirst:
## Change list

Expand touch options API to accept coordinates as `org.openqa.selenium.Point`
 
## Types of changes


src/test/java/io/appium/java_client/touch/TouchOptionsTests.java
```diff
@@ -71,7 +72,7 @@ public void longPressOptionsShouldBuildProperly() {
     @Test
     public void tapOptionsShouldBuildProperly() {
         final Map<String, Object> actualOpts = tapOptions()
-                .withPosition(point(0, 0))
+                .withPosition(point(new Point(0, 0)))
```
mykola-mokhnach: so now both would be possible `point(0, 0`) and `point(new Point(0, 0))` ?

SrinivasanTarget: Yes both are possible now.

PR #1001, opened 2018-08-14
Access caps from travis env variables

SrinivasanTarget:
## Change list

Access caps from travis env variables
 
## Types of changes

@mykola-mokhnach ping

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ matrix:
     - language: java
       install: true
       os: osx
-      osx_image: xcode9.3
+      osx_image: $xcodeVersion
       before_install:
         - brew update
         - brew outdated xctool || brew upgrade xctool
--- a/src/test/java/io/appium/java_client/ios/AppIOSTest.java
+++ b/src/test/java/io/appium/java_client/ios/AppIOSTest.java
@@ -27,9 +27,8 @@ public class AppIOSTest extends BaseIOSTest {
         File appDir = new File("src/test/java/io/appium/java_client");
         File app = new File(appDir, "TestApp.app.zip");
         DesiredCapabilities capabilities = new DesiredCapabilities();
-        capabilities.setCapability(MobileCapabilityType.BROWSER_NAME, "");
-        capabilities.setCapability(MobileCapabilityType.PLATFORM_VERSION, "11.3");
-        capabilities.setCapability(MobileCapabilityType.DEVICE_NAME, "iPhone 8");
+        capabilities.setCapability(MobileCapabilityType.PLATFORM_VERSION, PLATFORM_VERSION);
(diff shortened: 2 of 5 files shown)


SrinivasanTarget:
Looks like travis is gone mad... their doc says JDK 8 is default but i could see JDK 10 is picked up by default and its not possible to switch to 8...will try later...

SrinivasanTarget:
@imurchie Can you help on this if you know any other options to force travis to JDK 8?

imurchie:
```yml
jdk:
  - oraclejdk8
```
Seems to work.  https://travis-ci.org/appium/appium-espresso-driver/jobs/416092729

SrinivasanTarget:
@imurchie Looks like an issue specific to their OSX build images. may be we cant do much...

saikrishna321:
still uses jdk 10

Java version: 10.0.1, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk-10.0.1.jdk/Contents/Home

.travis.yml
```diff
@@ -1,18 +1,18 @@
-language: java
-
 sudo: required
 
 matrix:
   include:
-    - language: java
-      install: true
+    - language: node_js
```
imurchie: This won't work. With language as 'node_js' but no node version, you'll get 0.10, which Appium will not install with. Further, `jdk` will be ignored.

SrinivasanTarget: Yes

imurchie:
Weird. I just pushed a basically change-less branch (just adding logging the available Java versions) and it's getting 1.8 as far as I can tell.

https://travis-ci.org/appium/java-client/builds/416390977

SrinivasanTarget:
@imurchie Looks like its a problem only in `xcode 9.4` containers.

imurchie:
Oh yeah! That does seem to be the case.

imurchie:
The following works:
```diff
diff --git a/.travis.yml b/.travis.yml
index 6702255d..8b1b30ad 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,17 @@ sudo: required
 matrix:
   include:
     - language: java
+      jdk:
+        - oraclejdk8
       install: true
       os: osx
-      osx_image: xcode9.3
+      osx_image: xcode9.4
       before_install:
-        - brew update
-        - brew outdated xctool || brew upgrade xctool
+        - brew tap caskroom/versions
+        - brew cask install java8
+        - export JAVA_HOME=`/usr/libexec/java_home -v 1.8`
+        - java -version
+
         - npm install -g appium
       script:
         - ./gradlew clean build xcuiTest -x test -x signArchives
```
https://travis-ci.org/appium/java-client/builds/416414956

The `jdk` part is probably unnecessary, but you could have that, and some sort of check to see if the environment is already using 8 and skip the install step (for if Travis fixes the issue).

PR #1014, opened 2018-08-31
Add SLF4J logging support for Appium local service

alexander-poulikakos:
## Change list

Add possibilty to enable server output data logging through  [SLF4J](http://slf4j.org) loggers. This allow server output data to be configured with your preferred logging frameworks (e.g. java.util.logging, logback, log4j).
 
## Types of changes

## Details

**Normal use case:**
```Java
	    AppiumDriverLocalService service = ...
	    service.clearOutPutStreams();
	    service.enableDefaultSlf4jLoggingOfOutputData(); // <-- add this method call
	    service.start();
```
 By default log messages are:
* logged at `INFO` level, unless log message is pre-fixed by `[debug]` then logged at `DEBUG` level.
* logged by a [SLF4J](http://slf4j.org) logger instance with a logger name corresponding to the appium sub module as prefixed in log message (logger name is transformed to lower case, no spaces and prefixed with "appium.service.").

Example of log messages:
* Log message `"[ADB] Cannot read version codes of "` is logged by logger: `appium.service.adb` at level `INFO`
* Log message `"[debug] [XCUITest] Xcode version set to 'x.y.z' "` is logged by logger `appium.service.xcuitest` at level `DEBUG`

**More general purposes**
These methods are provided for more general use cases:

```java
service.addSlf4jLogMessageConsumer(BiConsumer);
service.addLogMessageConsumer(Consumer)
```

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
@@ -53,6 +53,7 @@ public final class AppiumDriverLocalService extends DriverService {
     private static final Logger LOG = LoggerFactory.getLogger(AppiumDriverLocalService.class);
     private static final Pattern LOG_MESSAGE_PATTERN = Pattern.compile("^(.*)\\R");
     private static final Pattern LOGGER_CONTEXT_PATTERN = Pattern.compile("^(\\[debug\\] )?\\[(.+?)\\]");
+    private static final String APPIUM_SERVICE_SLF4J_LOGGER_PREFIX = "appium.service";
     private final File nodeJSExec;
     private final ImmutableList<String> nodeJSArgs;
     private final ImmutableMap<String, String> nodeJSEnvironment;
@@ -264,10 +265,10 @@ public final class AppiumDriverLocalService extends DriverService {
      */
     public void enableDefaultSlf4jLoggingOfOutputData() {
         addSlf4jLogMessageConsumer((logMessage, ctx) -> {
-            if (ctx.level().equals(DEBUG)) {
-                ctx.logger().debug(logMessage);
+            if (ctx.getLevel().equals(DEBUG)) {
+                ctx.getLogger().debug(logMessage);
             } else {
-                ctx.logger().info(logMessage);
+                ctx.getLogger().info(logMessage);
             }
         });
     }
@@ -310,7 +311,7 @@ public final class AppiumDriverLocalService extends DriverService {
 
     static Slf4jLogMessageContext parseSlf4jContextFromLogMessage(String logMessage) {
         Matcher m = LOGGER_CONTEXT_PATTERN.matcher(logMessage);
-        String loggerName = "appium.service";
+        String loggerName = APPIUM_SERVICE_SLF4J_LOGGER_PREFIX;
         Level level = INFO;
         if (m.find()) {
             loggerName += "." + m.group(2).toLowerCase().replaceAll("\\s+", "");
--- a/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
@@ -238,28 +238,30 @@ public final class AppiumDriverLocalService extends DriverService {
      * Enables server output data logging through
      * <a href="http://slf4j.org">SLF4J</a> loggers. This allow server output
      * data to be configured with your preferred logging frameworks (e.g.
-     * java.util.logging, logback, log4j).</br>
-     * </br>
+     * java.util.logging, logback, log4j).
+     * <p>
      * NOTE1: You might want to call method {@link #clearOutPutStreams()} before
-     * calling this method. </br>
-     * NOTE2: it is required that {@link --log-timestamp} server flag is
-     * {@link false}. </br>
-     * </br>
+     * calling this method. 
+     * <br>
+     * NOTE2: it is required that {@code --log-timestamp} server flag is
+     * {@code false}.
+     * <p>
      * By default log messages are:
+     * <ul>
      * <li>logged at {@code INFO} level, unless log message is pre-fixed by
      * {@code [debug]} then logged at {@code DEBUG} level.</li>
      * <li>logged by a <a href="http://slf4j.org">SLF4J</a> logger instance with
      * a name corresponding to the appium sub module as prefixed in log message
      * (logger name is transformed to lower case, no spaces and prefixed with
-     * "appium.service.").</li> </br>
+     * "appium.service.").</li>
+     * </ul>
      * Example log-message: "[ADB] Cannot read version codes of " is logged by
-     * logger: {@code appium.service.adb} at level {@code INFO}. </br>
+     * logger: {@code appium.service.adb} at level {@code INFO}.
+     * <br>
      * Example log-message: "[debug] [XCUITest] Xcode version set to 'x.y.z' "
      * is logged by logger {@code appium.service.xcuitest} at level
      * {@code DEBUG}.
-     * 
-     * </br>
-     * </br>
+     * <br>
      * 
      * @see #addSlf4jLogMessageConsumer(BiConsumer)
      */
@@ -277,26 +279,29 @@ public final class AppiumDriverLocalService extends DriverService {
      * When a complete log message is available (from server output data) that
      * message is parsed for its slf4j context (logger name, logger level etc.)
      * and the specified {@code BiConsumer} is invoked with the log message and
-     * slf4j context.</br>
-     * </br>
+     * slf4j context.
+     * <p>
      * Use this method only if you want a behavior that differentiates from the
      * default behavior as enabled by method
-     * {@link #enableDefaultSlf4jLoggingOfOutputData()}. </br>
-     * </br>
+     * {@link #enableDefaultSlf4jLoggingOfOutputData()}.
+     * <p>
      * NOTE: You might want to call method {@link #clearOutPutStreams()} before
-     * calling this method. </br>
-     * </br>
+     * calling this method.
+     * <p>
      * implementation detail:
+     * <ul>
      * <li>if log message begins with {@code [debug]} then log level is set to
      * {@code DEBUG}, otherwise log level is {@code INFO}.</li>
      * <li>the appium sub module name is parsed from the log message and used as
      * logger name (prefixed with "appium.service.", all lower case, spaces
      * removed). If no appium sub module is detected then "appium.service" is
-     * used as logger name.</li> </br>
+     * used as logger name.</li>
+     * </ul>
      * Example log-message: "[ADB] Cannot read version codes of " is logged by
-     * {@code appium.service.adb} at level {@code INFO} </br>
+     * {@code appium.service.adb} at level {@code INFO} <br>
      * Example log-message: "[debug] [XCUITest] Xcode version set to 'x.y.z' "
      * is logged by {@code appium.service.xcuitest} at level {@code DEBUG}
+     * <br>
      * 
      * @param slf4jLogMessageConsumer
      *            BiConsumer block to be executed when a log message is
@@ -324,13 +329,14 @@ public final class AppiumDriverLocalService extends DriverService {
 
     /**
      * When a complete log message is available (from server output data), the
-     * specified {@code Consumer} is invoked with that log message. </br>
-     * </br>
+     * specified {@code Consumer} is invoked with that log message.
+     * <p>
      * NOTE: You might want to call method {@link #clearOutPutStreams()} before
-     * calling this method. </br>
-     * </br>
+     * calling this method.
+     * <p>
      * If the Consumer fails and throws an exception the exception is logged (at
      * WARN level) and execution continues.
+     * <br>
      * 
      * @param consumer
      *            Consumer block to be executed when a log message is available.
--- a/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static org.slf4j.event.Level.DEBUG;
 import static org.slf4j.event.Level.INFO;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
@@ -313,6 +314,7 @@ public final class AppiumDriverLocalService extends DriverService {
         });
     }
 
+    @VisibleForTesting
     static Slf4jLogMessageContext parseSlf4jContextFromLogMessage(String logMessage) {
         Matcher m = LOGGER_CONTEXT_PATTERN.matcher(logMessage);
         String loggerName = APPIUM_SERVICE_SLF4J_LOGGER_PREFIX;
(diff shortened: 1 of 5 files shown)


src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +232,127 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     * {@code DEBUG}.
+     * 
+     * </br>
+     * </br>
+     * 
+     * @see #addSlf4jLogMessageConsumer(BiConsumer)
+     */
+    public void enableDefaultSlf4jLoggingOfOutputData() {
+        addSlf4jLogMessageConsumer((logMessage, ctx) -> {
+            if (ctx.level().equals(DEBUG)) {
```
mykola-mokhnach: This mapping is not sufficient, since Appium server uses more levels, for example `warn` and `error`

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +232,127 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+    public void addSlf4jLogMessageConsumer(
+            BiConsumer<String, Slf4jLogMessageContext> slf4jLogMessageConsumer) {
+        addLogMessageConsumer(logMessage -> {
+            slf4jLogMessageConsumer.accept(logMessage, parseSlf4jContextFromLogMessage(logMessage));
+        });
+    }
+
+    static Slf4jLogMessageContext parseSlf4jContextFromLogMessage(String logMessage) {
+        Matcher m = LOGGER_CONTEXT_PATTERN.matcher(logMessage);
+        String loggerName = "appium.service";
```
mykola-mokhnach: This should be in the constant

src/main/java/io/appium/java_client/service/local/Slf4jLogMessageContext.java
```diff
@@ -0,0 +1,44 @@
...
+	Slf4jLogMessageContext(String loggerName, Level level) {
+		this.level = level;
+		this.logger = LoggerFactory.getLogger(loggerName);
+	}
+
+	/**
+	 * @return {@link Logger} instance associated with this context.
+	 * 
+	 */
+	public Logger logger() {
```
mykola-mokhnach: java getters usually start with `get`

alexander-poulikakos:
thanks for comments! What should I adjust indentation to? it seems to match the rest of the code lines.

src/main/java/io/appium/java_client/service/local/Slf4jLogMessageContext.java
alexander-poulikakos: will fix that

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: Will add the other levels as well.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: I'm not following you on this. What should be in the constant? Which constant?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: Hmm, I have never seen an actual log message on either `warn`/`error` level. Do you know if they follow same format as `debug`? i.e. log message is prefixed with log level. `Info` level tag is not added in log message.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +232,127 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     *            available.
+     */
+    public void addSlf4jLogMessageConsumer(
+            BiConsumer<String, Slf4jLogMessageContext> slf4jLogMessageConsumer) {
+        addLogMessageConsumer(logMessage -> {
+            slf4jLogMessageConsumer.accept(logMessage, parseSlf4jContextFromLogMessage(logMessage));
+        });
+    }
+
+    static Slf4jLogMessageContext parseSlf4jContextFromLogMessage(String logMessage) {
```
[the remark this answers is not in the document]
alexander-poulikakos: I dissagree to this comment. package private is needed for unit tests to access this method.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: I was just able to trigger a `WARN`message, but there is no log level indication in the log message it self. So will ignore `warn` and `error` levels for now.

Example
```
[Appium] Deprecated server args: 
[Appium]   --device-name => --default-capabilities '{"deviceName":"someName"}'
```
According to this, the above log message is at `WARN`level:
https://github.com/appium/appium/blob/7a00b3f875991ba1d95359ed53a3eda47911f64b/lib/main.js#L44

alexander-poulikakos:
@mykola-mokhnach Can you please give more context regarding you comment `Please adjust indentation in all source files`. What should the indentation? I don't see it differ from other code lines.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
mykola-mokhnach: the actual level is added there if logging timestamp is enabled:

```
2018-08-31 08:17:13:813 - info: [debug] [JSONWP Proxy] Got response with status 200: {"value":{"using":"class chain","value":"**/XCUIElementTypeCollectionView[`name == 'add_participants.list'`]/XCUIElementTypeCell[$name == 'user_cell.name' AND value == 'qrj7pcz5'$]","description":"unable to find an element"},"sessionId":"A4FBAD14-ED57-46D4-9407-B91C76BCB415","status":7}
2018-08-31 08:17:13:814 - info: [debug] [MJSONWP] Matched JSONWP error code 7 to NoSuchElementError
2018-08-31 08:17:13:815 - error: [W3C] Encountered internal error running command: NoSuchElementError: An element could not be located on the page using the given search parameters.
2018-08-31 08:17:13:815 - error: [W3C]     at XCUITestDriver.callee$0$0$ (/usr/local/lib/node_modules/appium/node_modules/appium-xcuitest-driver/lib/commands/find.js:130:13)
```

alexander-poulikakos:
@mykola-mokhnach How many spaces should it be per indentation?

mykola-mokhnach:
4 spaces

alexander-poulikakos:
@mykola-mokhnach 
getter names fixed
indentation fixed
constant fixed
javadoc fixed

regarding loglevel `error` and `warn` I'm skipping those according to comment above (log level tag is not present in log message -> can't parse it)

Anything else that I have missed?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +233,132 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     * Example log-message: "[debug] [XCUITest] Xcode version set to 'x.y.z' "
+     * is logged by logger {@code appium.service.xcuitest} at level
+     * {@code DEBUG}.
+     * <br>
+     * 
+     * @see #addSlf4jLogMessageConsumer(BiConsumer)
+     */
+    public void enableDefaultSlf4jLoggingOfOutputData() {
+        addSlf4jLogMessageConsumer((logMessage, ctx) -> {
+            if (ctx.getLevel().equals(DEBUG)) {
```
mykola-mokhnach: this is anyway not very accurate, since this `[debug]` prefix is also not always added to debug logs

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +233,132 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     *            available.
+     */
+    public void addSlf4jLogMessageConsumer(
+            BiConsumer<String, Slf4jLogMessageContext> slf4jLogMessageConsumer) {
+        addLogMessageConsumer(logMessage -> {
+            slf4jLogMessageConsumer.accept(logMessage, parseSlf4jContextFromLogMessage(logMessage));
+        });
+    }
+
+    static Slf4jLogMessageContext parseSlf4jContextFromLogMessage(String logMessage) {
```
mykola-mokhnach: codacy does not like package-level visibility. Please add `public`

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +233,132 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     * <p>If the Consumer fails and throws an exception the exception is logged (at
+     * WARN level) and execution continues.
+     * <br>
+     * 
+     * @param consumer
+     *            Consumer block to be executed when a log message is available.
+     * 
+     */
+    public void addLogMessageConsumer(Consumer<String> consumer) {
+        addOutPutStream(new OutputStream() {
```
mykola-mokhnach: Output

src/test/java/io/appium/java_client/service/local/AppiumDriverLocalServiceTest.java
```diff
@@ -0,0 +1,40 @@
...
+        assertLoggerContext(INFO, "appium.service", 
+            "no-prefix log message.");
+        assertLoggerContext(INFO, "appium.service", 
+            "no-prefix log [not-a-logger-name] message.");
+        assertLoggerContext(DEBUG, "appium.service.mjsonwp", 
+            "[debug] [MJSONWP] Calling AppiumDriver.getStatus() with args: []");
+        assertLoggerContext(DEBUG, "appium.service.xcuitest", 
+            "[debug] [XCUITest] Xcode version set to 'x.y.z' ");
+        assertLoggerContext(DEBUG, "appium.service.jsonwpproxy", 
+            "[debug] [JSONWP Proxy] Proxying [GET /status] to [GET http://localhost:18218/status] with no body");
```
mykola-mokhnach: How would you parse messages with timestamps like `2018-08-31 08:17:13:741 - info: [debug] [JSONWP Proxy] Matched '/element' to command name 'findElement'` ?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: @mykola-mokhnach There is a big difference between public and package private. Making it public will make it part of the public API, which is not desired in this case.

I don't think we should make an implementation detail public just because codacy says so, or?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: @mykola-mokhnach So what do you sugests instead? Should all be logged at ÌNFO` level?
As it is now log messages with `[debug]` prefix is logged at `DEBUG` level. That is better than nothing if you ask me.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: I'm not following this comment? `addOutPutStream(...)` is the name of an already existing method.

src/test/java/io/appium/java_client/service/local/AppiumDriverLocalServiceTest.java
alexander-poulikakos: as mentioned in javadoc. this feature is not supported when `--log-timestamp` is enabled.

from javadoc:
`NOTE2: it is required that {@code --log-timestamp} server flag is {@code false}.`

alexander-poulikakos: @mykola-mokhnach  ^

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
mykola-mokhnach: There is a typo in this word

alexander-poulikakos: @mykola-mokhnach 
I'm still not following what this has to do with this PR?
As I said the `addOutPutStream(...)` method is already existing. Changing its name would be a breaking change.

mykola-mokhnach: then yes, it is better to keep it as is if it was there

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -114,6 +115,15 @@ public void tearDown() throws Exception {
         ofNullable(PATH_TO_APPIUM_NODE_IN_PROPERTIES).ifPresent(s -> setProperty(APPIUM_PATH, s));
     }
 
+    @Test public void checkAbilityToAddLogMessageConsumer() {
+        List<String> log = new ArrayList<>();
+        service = buildDefaultService();
+        service.clearOutPutStreams();
+        service.addLogMessageConsumer(log::add);
```
mykola-mokhnach: should we also have a possibility to remove a consumer?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -219,4 +233,132 @@ public void addOutPutStreams(List<OutputStream> outputStreams) {
...
+     * {@code appium.service.adb} at level {@code INFO} <br>
+     * Example log-message: "[debug] [XCUITest] Xcode version set to 'x.y.z' "
+     * is logged by {@code appium.service.xcuitest} at level {@code DEBUG}
+     * <br>
+     * 
+     * @param slf4jLogMessageConsumer
+     *            BiConsumer block to be executed when a log message is
+     *            available.
+     */
+    public void addSlf4jLogMessageConsumer(
```
mykola-mokhnach: I find it useful to return self instance from such methods, so one can chain them

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
alexander-poulikakos: Essentially a consumer in this case is an OutputStream (as added with `addOutPutStream(...)` method). So consumers can be removed with the `clearOutPutStreams()` method.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
alexander-poulikakos: I agree it is nice to return `self` to chain method calls. But no other methods in this class does that, so I figured to follow same pattern as other methods. Would be strange if only this method returned `self`.

alexander-poulikakos:
I just realized these methods does not do a null check of given arguments. Will fix that.

```
addSlf4jLogMessageConsumer(...)
addLogMessageConsumer(...)
```

alexander-poulikakos:
Will this be merged?

PR #1028, opened 2018-09-19
Clean up

valfirst:
## Change list

- Remove redundant 'throws' clauses
- Simplify assertions in unit tests
- Replace 'Collection.addAll()' with parametrized constructor call
- Remove redundant array creation for calling varargs method
- Remove redundant type arguments
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
+++ b/src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
@@ -96,7 +96,7 @@ public abstract class AppiumByBuilder extends AbstractAnnotations {
             }
 
             try {
-                String strategyParameter = value.invoke(mobileBy, new Object[] {}).toString();
+                String strategyParameter = value.invoke(mobileBy).toString();
                 if (!"".equals(strategyParameter)) {
                     return value.getName();
                 }
--- a/src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
+++ b/src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
@@ -97,7 +97,7 @@ public abstract class AppiumByBuilder extends AbstractAnnotations {
 
             try {
                 String strategyParameter = value.invoke(mobileBy).toString();
-                if (!"".equals(strategyParameter)) {
+                if (!strategyParameter.isEmpty()) {
                     return value.getName();
                 }
             } catch (IllegalAccessException
--- a/src/test/java/io/appium/java_client/service/local/ThreadSafetyTest.java
+++ b/src/test/java/io/appium/java_client/service/local/ThreadSafetyTest.java
@@ -1,5 +1,6 @@
 package io.appium.java_client.service.local;
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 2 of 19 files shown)


src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
```diff
@@ -96,7 +96,7 @@ private static String getFilledValue(Annotation mobileBy) {
             }
 
             try {
-                String strategyParameter = value.invoke(mobileBy, new Object[] {}).toString();
+                String strategyParameter = value.invoke(mobileBy).toString();
```
mykola-mokhnach: does this call return the same result as the previous one?

src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
```diff
@@ -96,7 +96,7 @@ private static String getFilledValue(Annotation mobileBy) {
...
                 if (!"".equals(strategyParameter)) {
```
mykola-mokhnach: we could use isEmpty() here

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -138,6 +138,6 @@ public void getDeviceTimeTest() {
         driver.toggleTouchIDEnrollment(true);
         driver.performTouchID(true);
         driver.performTouchID(false);
-        assertEquals(true, true);
+        assertTrue(true);
```
mykola-mokhnach: %)

src/test/java/io/appium/java_client/service/local/ThreadSafetyTest.java
```diff
@@ -86,8 +87,8 @@
                 throw isRunningTestThread2.throwable;
             }
 
-            assertTrue(isRunningTestThread.result.equals(true));
-            assertTrue(isRunningTestThread2.result.equals(true));
+            assertEquals(true, isRunningTestThread.result);
```
mykola-mokhnach: cannot we just test that `assertTrue(isRunningTestThread.result)` ?

src/main/java/io/appium/java_client/pagefactory/bys/builder/AppiumByBuilder.java
valfirst: yes, the method signature is:
```java
java.lang.reflect.Method.invoke(Object, Object...)
```
so it's not required to pass an empty array as varargs

src/test/java/io/appium/java_client/service/local/ThreadSafetyTest.java
valfirst: we can't: `assertTrue` requires `boolean` input parameter, but `isRunningTestThread.result` is of `Object` type

mykola-mokhnach: I'd rather use Boolean.valueOf then

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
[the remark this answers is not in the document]
mykola-mokhnach: yep, now I remember why `assertEquals(true, true);` has been used ;)

PR #1059, opened 2018-11-19
Set up CI with Azure Pipelines

SrinivasanTarget:


azure-pipelines.yml
```diff
@@ -0,0 +1,12 @@
...
+# Add steps that analyze code, save build artifacts, deploy, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/java
+
+pool:
+  vmImage: 'macOS 10.13'
+
+steps:
+- script: |
+    npm install -g appium@beta
+    ./gradlew clean build xcuiTest -x test -x signArchives
```
KazuCocoa: What about using https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/gradle?view=vsts like below?

```
- task: Gradle@2
  inputs:
    gradleWrapperFile: 'gradlew' 
    tasks: 'clean build' 
    options: 'xcuiTest -x test -x signArchives'
```

According to https://github.com/Microsoft/azure-pipelines-image-generation/blob/master/images/macos/macos-Readme.md, we probably can use the `gradle` task.

SrinivasanTarget: @KazuCocoa Noted. Neat and clear.

src/test/java/io/appium/java_client/appium/element/generation/BaseElementGenerationTest.java
```diff
@@ -34,6 +35,7 @@ protected boolean check(Supplier<DesiredCapabilities> serverCapabilitiesSupplier
                          BiPredicate<By, Class<? extends WebElement>> filter,
                          By by, Class<? extends WebElement> clazz) {
         AppiumServiceBuilder builder = new AppiumServiceBuilder()
+                .withAppiumJS(new File("/usr/local/lib/node_modules/appium/build/lib/main.js"))
```
mykola-mokhnach: shouldn't this be cross-platform?

SrinivasanTarget: Sorry didn't get you?

mykola-mokhnach: we could find npm root by executing `npm root -g`. The second part of the path is always constant and can be added using path joining

SrinivasanTarget: Yup client does the same internally [here](https://github.com/appium/java-client/blob/master/src/main/resources/scripts/get_path_to_default_node.sh) but not sure why it couldn't find it inside the azure mac host. So i harcoded the same.

mykola-mokhnach: I would say it would be better to put  appium.js path calculation to a separate helper method rather than hardcoding it

SrinivasanTarget: yup will do that

mykola-mokhnach: Something like `Paths.get(<Path_to_npm_root>, "appium", "build" ,"lib", "main.js")`

mykola-mokhnach:
I would try to perform

```
npm config delete prefix 
nvm alias default node
```

in order to fix azure failures

mykola-mokhnach:
https://stackoverflow.com/questions/34718528/nvm-is-not-compatible-with-the-npm-config-prefix-option

src/test/java/io/appium/java_client/appium/element/generation/BaseElementGenerationTest.java
SrinivasanTarget: I'm thinking why does client doesn't detect globally installed server path automatically when there is nvm in place in a system... I do remember there are some issues related to it.

mykola-mokhnach:
@SrinivasanTarget Probably, it makes sense to install nvm and node same way it is done on Travis instead of using the built-in one.

SrinivasanTarget:
@mykola-mokhnach Its fixed now. Can you look into it?

PR #1081, opened 2018-12-31
Migrate CI to Azure for both iOS and Android

SrinivasanTarget:
## Change list

Migrate CI to Azure for both iOS and Android
 
## Types of changes


SrinivasanTarget:
https://github.com/appium/java-client/issues/1064

azure-pipelines.yml
```diff
@@ -13,6 +13,14 @@ steps:
 
 - script: |
     echo Configuring Environment
+    echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'system-images;android-27;google_apis;x86'
+    echo "no" | $ANDROID_HOME/tools/bin/avdmanager create avd -n xamarin_android_emulator -k 'system-images;android-27;google_apis;x86' --force
+    echo $ANDROID_HOME/emulator/emulator -list-avds
+    echo "Starting emulator"
+    nohup $ANDROID_HOME/emulator/emulator -avd xamarin_android_emulator -no-snapshot > /dev/null 2>&1 &
+    $ANDROID_HOME/platform-tools/adb wait-for-device shell 'while [[ -z $(getprop sys.boot_completed | tr -d '\r') ]]; do sleep 1; done; input keyevent 82'
```
mykola-mokhnach: checking `sys.boot_completed` might sometimes be not reliable if CI is too slow, because the package manager might take more time to start. I'd rather verify if `adb shell pm list packages` returns non-zero exit code

azure-pipelines.yml
```diff
@@ -13,6 +13,14 @@ steps:
 
 - script: |
     echo Configuring Environment
+    echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'system-images;android-27;google_apis;x86'
+    echo "no" | $ANDROID_HOME/tools/bin/avdmanager create avd -n xamarin_android_emulator -k 'system-images;android-27;google_apis;x86' --force
+    echo $ANDROID_HOME/emulator/emulator -list-avds
+    echo "Starting emulator"
+    nohup $ANDROID_HOME/emulator/emulator -avd xamarin_android_emulator -no-snapshot > /dev/null 2>&1 &
```
mykola-mokhnach: can we have a separate variable for the emulator name?

azure-pipelines.yml
```diff
@@ -13,6 +13,14 @@ steps:
 
 - script: |
     echo Configuring Environment
+    echo "y" | $ANDROID_HOME/tools/bin/sdkmanager --install 'system-images;android-27;google_apis;x86'
```
mykola-mokhnach: can we have a variable for image name? Also, it might be useful to set platform and tools versions separately

.travis.yml
```diff
@@ -1,47 +0,0 @@
-sudo: required
```
mykola-mokhnach: so we'd like to get rid of Travis completely...
Is Azure stable enough? is it supposed to be always free for open source?

azure-pipelines.yml
SrinivasanTarget: Yeah I accept that. I see we do that at many places but I went with azure suggestions and so far their CI is faster. Travis took more than 20mins to complete but azure agents took only 10mins. Will definitely check and improve if required.

.travis.yml
SrinivasanTarget: Yes as per their documents and offerings in their website. I see it is much stable than Travis.

azure-pipelines.yml
mykola-mokhnach: Travis is so slow because their VMs don't support KVM virtualization required for x86 emulator. And, of course, emulating ARM instructions set on Intel processors is slow

SrinivasanTarget: True

SrinivasanTarget:
@mykola-mokhnach sorry I missed this PR. Please review when you get time.

SrinivasanTarget:
@mykola-mokhnach I added Travis back. We can check both parallelly for some time and remove either of one if any of this is not working for us.

PR #1094, opened 2019-01-29
Avoid null pointer exceptions when calling getCapabilities

etanol:
## Change list

Add null checks to `getCapabilities()` that prevent spurious null pointer exceptions.

## Types of changes

## Details

The `capabilities` field in `RemoteWebDriver` remains null until a new session is successfully created.  If something fails before the Selenium client can consolidate the returned capabilities, a `NullPointerException` is thrown from the Appium overriden `getCapabilities` methods that masks the real failure.

In other words:

- Appium [overrides `getCapabilities`][2]
- Selenium first [requests a new session][3], then [saves the returned capabilities][1]
- If the new session command fails, [`getCapabilities` is called][4]
- Because `RemoteWebDriver` returns null, the Appium override [tries to call a method on a null value][5]

This exception masking results in very confusing and misleading troubleshooting.

The fix proposed by this pull request should fix the problem.

[1]: https://github.com/SeleniumHQ/selenium/blob/e82be7d3584062a0d16af8d562d387e3d4855aa1/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L252
[2]: https://github.com/appium/java-client/blob/98daabe3943b606c5416818b73add493d889cc94/src/main/java/io/appium/java_client/android/AndroidDriver.java#L209-L213
[3]: https://github.com/SeleniumHQ/selenium/blob/e82be7d3584062a0d16af8d562d387e3d4855aa1/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L213
[4]: https://github.com/SeleniumHQ/selenium/blob/e82be7d3584062a0d16af8d562d387e3d4855aa1/java/client/src/org/openqa/selenium/remote/RemoteWebDriver.java#L577
[5]: https://github.com/appium/java-client/blob/98daabe3943b606c5416818b73add493d889cc94/src/main/java/io/appium/java_client/android/AndroidDriver.java#L211

### DIFF
--- a/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
+++ b/src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
@@ -157,7 +157,8 @@ abstract class DefaultGenericMobileDriver<T extends WebElement> extends RemoteWe
 
     @Override
     public String toString() {
+        Capabilities capabilities = getCapabilities();
         return String.format("%s, Capabilities: %s", getClass().getCanonicalName(),
-                getCapabilities().asMap().toString());
+                capabilities != null ? capabilities.asMap().toString() : "null");
     }
 }
--- a/src/main/java/io/appium/java_client/android/AndroidDriver.java
+++ b/src/main/java/io/appium/java_client/android/AndroidDriver.java
@@ -47,6 +47,7 @@ import org.openqa.selenium.remote.http.HttpClient;
 import java.net.URL;
 import java.util.Collections;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
  * Android driver implementation.
@@ -206,9 +207,12 @@ public class AndroidDriver<T extends WebElement>
      *
      * @return given {@link Capabilities}
      */
+    @Nullable
     public Capabilities getCapabilities() {
         MutableCapabilities capabilities = (MutableCapabilities) super.getCapabilities();
-        capabilities.setCapability(PLATFORM_NAME, ANDROID_PLATFORM);
+        if (capabilities != null) {
+            capabilities.setCapability(PLATFORM_NAME, ANDROID_PLATFORM);
+        }
         return capabilities;
     }
 
--- a/src/main/java/io/appium/java_client/ios/IOSDriver.java
+++ b/src/main/java/io/appium/java_client/ios/IOSDriver.java
@@ -48,6 +48,7 @@ import java.net.URL;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.Map;
+import javax.annotation.Nullable;
 
 /**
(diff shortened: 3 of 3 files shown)


src/main/java/io/appium/java_client/DefaultGenericMobileDriver.java
```diff
@@ -157,7 +157,8 @@ public List findElementsByXPath(String using) {
 
     @Override
     public String toString() {
+        Capabilities capabilities = getCapabilities();
         return String.format("%s, Capabilities: %s", getClass().getCanonicalName(),
-                getCapabilities().asMap().toString());
+                capabilities != null ? capabilities.asMap().toString() : "null");
```
mykola-mokhnach: I'd rather use checkNotNull macros everywhere

mykola-mokhnach: or this is normal that `getCapabilities` can sometimes return null?

etanol: `getCapabilities` will be null **until** the response of the *new session* request, where the real capabilities are returned by the server.

As I mentioned in the description, if the *new session* command fails (which can be pretty often when you are trying to set up the proper version combination of Appium, Automation and device), this `NullPointerException` masks the real failure (with **no** exception nesting).

So, overall, this makes the initial set up incredibly frustrating.

mykola-mokhnach: yep, I've checked the Selenium code and it's how it currently works. Although I would like to get the response to https://github.com/SeleniumHQ/selenium/issues/6903. What if this is just a bug and not the expected behaviour?

etanol: I'm ok about learning the *official posture*.  However, if `RemoteWebDriver` does indeed have a typo (missing a `this.`), then the exception masking would still remain.  But instead of `NullPointerException`, a `ClassCastException` would be thrown:

https://github.com/appium/java-client/blob/fa71401de3fdb74eef7e4e7d513cf0c8e6adbe36/src/main/java/io/appium/java_client/android/AndroidDriver.java#L210

In such a case, the fix would be much uglier.

src/main/java/io/appium/java_client/android/AndroidDriver.java
```diff
@@ -208,7 +208,9 @@ public AndroidBatteryInfo getBatteryInfo() {
      */
     public Capabilities getCapabilities() {
```
mykola-mokhnach: Can you please add Nullable annotation to methods which can return null? It seems like we are not going to get any response from Selenium contributors, so will merge this fix as is.

PR #1142, opened 2019-05-12
Add apple tvOS platform constant

ajeeshvl:
## Change list

This is to add support for apple tvOS.
 
## Types of changes

## Details

This adds tvOS platform.
tvOS also uses xcuitest-driver.

Related PRs 
https://github.com/appium/appium-xcuitest-driver/pull/911
https://github.com/appium/appium/pull/12401


src/main/java/io/appium/java_client/ios/TVOSDriver.java
```diff
@@ -0,0 +1,176 @@
...
+ *           {@link WebElement}
+ *           {@link org.openqa.selenium.remote.RemoteWebElement}
+ *           {@link io.appium.java_client.MobileElement}
+ *           {@link IOSElement}
+ */
+public class TVOSDriver<T extends WebElement>
+    extends AppiumDriver<T>
+    implements HidesKeyboardWithKeyName, HasIOSSettings, HasOnScreenKeyboard,
+        LocksDevice, FindsByIosNSPredicate<T>, FindsByIosClassChain<T>,
+        PushesFiles, CanRecordScreen, HasIOSClipboard {
```
KazuCocoa: Did you check all of commands work? (just a confirmation.)
Clipboad does not work on this list.

mykola-mokhnach:
For me it does not make sense to have a separate class, which, actually, adds nothing. We already had youidriver like that and removed it.

ajeeshvl:
@mykola-mokhnach sounds good. Since AppiumDriver is not abstract, we can use that. Do you think, platform name `tvOS` can still be added in MobilePlatform class?

mykola-mokhnach:
> Do you think, platform name tvOS can still be added in MobilePlatform class?

Yes, this will be useful to have a constant for it

ajeeshvl:
@mykola-mokhnach @KazuCocoa @SrinivasanTarget removed TVOSDriver class.

src/main/java/io/appium/java_client/remote/MobilePlatform.java
```diff
@@ -22,4 +22,5 @@
     String IOS = "iOS";
     String FIREFOX_OS = "FirefoxOS";
     String WINDOWS = "Windows";
+    String TVOS = "tvOS";
```
mykola-mokhnach: Does it make sense to have it inside `AutomationName` interface instead?

KazuCocoa: No, I do not think so in this case.
It confuses users a bit since when we use tvOS, the capability should be like:

```
...
automationName: 'xcuites', # AutomationName.IOS_XCUI_TEST
platformName: 'tvOS'      # AutomationName.IOS_TVOS or MobilePlatform.TVOS (or APPLE_TVOS is proper?)
...
```

ajeeshvl: Unlike youiengine tvos uses xcuitest as automation name. I still think having a TVOSdriver would also be helpful. That would avoid confusion for users around what capability needs to be used for tvos.

KazuCocoa: I don't have a strong opinion about the new class.
I would respect @mykola-mokhnach, @saikrishna321 and @SrinivasanTarget

saikrishna321: I would say not to have TVOSDriver class. We should just document what capability to use for TVOS.

Issue #1137, opened 2019-04-30
Feature: Add support for "driver.status()"


amedvedjev:
## Description

Please add support for driver status.

http://appium.io/docs/en/commands/status/
https://github.com/appium/appium/issues/12490

## Environment

* java client build version or git revision if you use some shapshot: 7.0.0
* Appium server version or git revision if you use some shapshot: 1.13.0-beta.3

anilkrverma:
The implementation for this issue is merged in the master branch and should be live in next appium-java-client release.

saikrishna321:
You can get the latest master commit from jitpack to use and need not wait for release.

PR #1153, opened 2019-06-13
Implemented driver status check.

anilkrverma:
## Change list
Feature: Add support for "driver.status()" - #1137 
 
## Types of changes

- [x ] No changes in production code.

## Details
- Implemented driver status check as requested in #1137 which returns AppiumServerStatus object containing a field for returning the version of running appium server.
- Added a sample test to validate if the version is of form "x.y"


anilkrverma:
The tests are failing even when no changes are made in the code.
Refer https://github.com/appium/java-client/pull/1155  which has a change in Readme text only.

This failure shouldn't be related to the code implemented here.

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -223,6 +223,15 @@ protected static Capabilities substituteMobilePlatform(Capabilities originalCapa
         return contextName;
     }
 
+    public AppiumServerStatus getStatus() {
```
mykola-mokhnach: I'd rather simply return a map out of it. The status output is not constant and differs for different drivers

anilkrverma: This is related to Appium Server status that shows the build version info when we call "http://localhost:4723/wd/hub/status" while running appium locally. The idea was to add the ability in the driver to return its build version when someone calls getStatus() method which returns AppiumServerStatus object having getVersion() method to return the build version value. 
By returning the map, we might be forcing the user to know the key (i.e. "version" in this case) and pass this to get the value from the map.
Am I missing anything here?

mykola-mokhnach: > By returning the map, we might be forcing the user to know the key (i.e. "version" in this case) and pass this to get the value from the map.

I'm fine with that

anilkrverma: @mykola-mokhnach, @jlipps , @saikrishna321, @SrinivasanTarget 
My idea was to keep it simple and scalable by having a separate Class for managing the status (and to have ease in adding new fields, if any in the future) rather than expecting users to know about the internal fields. 
But if you guys ok with returning the versionMap from getStatus() method implemented here, then I can go ahead and change the code to return versionMap info?

saikrishna321: I would go with just returning the versionMap and let the user pick the values from it.

anilkrverma: Updated the code as suggested and done formatting. CI tests are still breaking which needs to be fixed separately.

anilkrverma:
> Code looks good now. Can you keep styling as per the style guides we use?

@SrinivasanTarget Current changes have been done through standard auto-formatting in Java. Any reference to style guides doc for my understanding so that I can update it accordingly and be proactive in my future commits?

SrinivasanTarget:
@anilkrverma https://github.com/appium/java-client/blob/master/google-style.xml

anilkrverma:
> @anilkrverma https://github.com/appium/java-client/blob/master/google-style.xml
> Gradle task: `./gradlew clean checkstyleMain`

@SrinivasanTarget Pushed the changes after fixing the violations reported. Please review.

SrinivasanTarget:
@anilkrverma Can you rebase your branch with the latest master once as a final nit. CI should be green after that.

anilkrverma:
Hurray, CI is green now. :)

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -202,28 +208,42 @@ protected static Capabilities substituteMobilePlatform(Capabilities originalCapa
...
-    @Override public DeviceRotation rotation() {
+    /**
+     * This method is used to get build version status of running Appium server.
+     *
+     * @return map containing version details
+     */
+    public Map<String, String> getStatus() {
+        Map<String, Object> statusMap = (Map<String, Object>) execute(DriverCommand.STATUS).getValue();
+
+        return (Map<String, String>) statusMap.get("build");
```
mykola-mokhnach: why cannot we just return the statusMap ?

mykola-mokhnach: Also, this is not a very good idea to cast `Map<String, Object>` to `Map<String, String>`

anilkrverma: @mykola-mokhnach For simplicity, it has been implemented this way.
@jlipps , @saikrishna321 , @SrinivasanTarget What do you guys think? Does it need further changes or can be merged?
If we really want to enhance it further, I believe the better approach will to have pojo classes created for JSON response (i.e. {"build":{"version":"1.13.0"}}, in this case 2 pojo classes will be created one parent pojo containing an instance of Build class and getter/setter for it and another pojo file "Build" containing a field version and getter/setter for that) and then we do the parsing for extracting version field. Please suggest.

SrinivasanTarget: We don't need to add more POJO's here. We just need not do the casting and send the status map as is. POJO is overkill to me. Heavy lifting should be done by users as per their needs.

anilkrverma: @SrinivasanTarget 
You mean getStatus() should look like this:

**public Map<String, Object> getStatus() {
        Map<String, Object> statusMap = (Map<String, Object>) execute(DriverCommand.STATUS).getValue();
        return statusMap;
    }**

anilkrverma:
@SrinivasanTarget , @mykola-mokhnach Rebased from master. Please review.

anilkrverma:
@SrinivasanTarget @mykola-mokhnach @jlipps @saikrishna321 We are getting below error again:
Error: /Users/vsts/agent/2.153.2/work/1/s/gradlew failed with return code: 1

It was working before, Did we broke Azure CI again or am I missing something here?

mykola-mokhnach:
This does not look like a rebase

anilkrverma:
> This does not look like a rebase
Should be good now. @mykola-mokhnach

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -202,28 +208,42 @@ protected static Capabilities substituteMobilePlatform(Capabilities originalCapa
...
     }
 
-    @Override public DeviceRotation rotation() {
+    /**
+     * This method is used to get build version status of running Appium server.
+     *
+     * @return map containing version details
+     */
+    public Map<String, Object> getStatus() {
+        Map<String, Object> statusMap = (Map<String, Object>) execute(DriverCommand.STATUS).getValue();
```
mykola-mokhnach: return (Map<String, Object>) execute(DriverCommand.STATUS).getValue();

anilkrverma: Done.

saikrishna321:
Thanks for addressing all the comments. Looks good👍

jlipps:
🎉  congrats @anilkrverma!

anilkrverma:
Thanks @jlipps , @saikrishna321 @SrinivasanTarget @mykola-mokhnach for all the support.

PR #1163, opened 2019-06-24
Moving cglib dependency to an older version before 3.2.6 to resolve i…

canaangifford:
Referencing https://github.com/appium/java-client/issues/1100

## Change list

Moving cglib dependency to a previous version to resolve issue with the Enhancer class.
 
## Types of changes

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -68,7 +68,7 @@ dependencies {
     }
     compile 'com.google.code.gson:gson:2.8.5'
     compile 'org.apache.httpcomponents:httpclient:4.5.8'
-    compile 'cglib:cglib:3.2.10'
+    compile 'cglib:cglib:3.2.5'
     compile 'commons-validator:commons-validator:1.6'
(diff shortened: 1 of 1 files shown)


canaangifford:
See details in the bug discussion:
https://github.com/appium/java-client/issues/1100#issuecomment-501813322

I'm not sure why this cglib version is being used by appium, more insight into why would be wonderful, thanks!

mykola-mokhnach:
@valfirst Are you ok with this change?

valfirst:
downgrading `cglib` may make `java-client` incompatible with newer Java versions, I would propose to perform deep investigation and find out actual reason (in case of `cglib` or dependent library bug downgrade could be reasonable, otherwise I would postpone blind downgrade)

valfirst:
@mykola-mokhnach I've investigated the issue a bit, it seems it's not caused by version upgrade: https://github.com/appium/java-client/issues/1100#issuecomment-505208545

Issue #1044, opened 2018-10-18
Attach to an already-existing session


i0:
Hi everyone,

Is there a way to attach the java client to an existing session? 
For javascript-wd driver there is an "attach" method which connects the driver to an existing session:
https://github.com/admc/wd#attach-to-an-already-existing-session

So, I was just wondering if there is a similar method in the java client library?

Thanks

zyziyun:
same question
+1

saikrishna321:
@Jonahss This is something which we discussed at the appium conf

Jonahss:
Totally! This is a really crucial feature that I used to use all the time. The Java client just extends the selenium project's java client, and when they updated they may have made this harder.

I'll see what I can do. It wasn't obvious how to do this with the new versions. Maybe this can be the next AppiumPro article ;)

ganjarpanji:
This will help with app debugging purpose, sometimes there are app error cases that only occurred when the app navigated by appium. looking forward to these :)

boris779:
+1 would really like to see that.

PR #1181, opened 2019-07-16
add a new constructor for creating a client attached to an existing s…

Jonahss:
…ession

## Change list

A new constructor which attaches to an existing Appium session
 
## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Details

Related to https://github.com/appium/java-client/issues/1044

It was rather tricky to do something so simple. The reason being that the selenium client constructor calls the POST /session command. We needed to construct a client with all the right values but _can't_ call the original constructor, because if we do, it will attempt to create a session.

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -24,6 +24,7 @@ import com.google.common.collect.ImmutableMap;
 
 import io.appium.java_client.internal.JsonToMobileElementConverter;
 import io.appium.java_client.remote.AppiumCommandExecutor;
+import io.appium.java_client.remote.AppiumW3CHttpCommandCodec;
 import io.appium.java_client.remote.MobileCapabilityType;
 import io.appium.java_client.service.local.AppiumDriverLocalService;
 import io.appium.java_client.service.local.AppiumServiceBuilder;
@@ -39,6 +40,7 @@ import org.openqa.selenium.WebElement;
 import org.openqa.selenium.html5.Location;
 
 import org.openqa.selenium.remote.DesiredCapabilities;
+import org.openqa.selenium.remote.Dialect;
 import org.openqa.selenium.remote.DriverCommand;
 import org.openqa.selenium.remote.ErrorHandler;
 import org.openqa.selenium.remote.ExecuteMethod;
@@ -84,11 +86,7 @@ public class AppiumDriver<T extends WebElement>
      */
     public AppiumDriver(HttpCommandExecutor executor, Capabilities capabilities) {
         super(executor, capabilities);
-        this.executeMethod = new AppiumExecutionMethod(this);
-        locationContext = new RemoteLocationContext(executeMethod);
-        super.setErrorHandler(errorHandler);
-        this.remoteAddress = executor.getAddressOfRemoteServer();
-        this.setElementConverter(new JsonToMobileElementConverter(this, this));
+        init(executor);
     }
 
     public AppiumDriver(URL remoteAddress, Capabilities desiredCapabilities) {
@@ -129,6 +127,35 @@ public class AppiumDriver<T extends WebElement>
 
     public AppiumDriver(Capabilities desiredCapabilities) {
         this(AppiumDriverLocalService.buildDefaultService(), desiredCapabilities);
+
+    }
+
+    /**
+     * For creating a client connected to an existing Appium session.
+     * The other constructors all create a new session, sending the POST /session request.
+     * This constructor does not.
+     */
+    public AppiumDriver(URL remoteAddress, String sessionId) {
+        this.setSessionId(sessionId);
+        
+        AppiumCommandExecutor executor = new AppiumCommandExecutor(MobileCommand.commandRepository, remoteAddress);
+        executor.setCommandCodec(new AppiumW3CHttpCommandCodec());
+        executor.setResponseCodec(Dialect.W3C.getResponseCodec());
+        executor.configureW3CMode();
+        
+        this.setCommandExecutor(executor);
+        init(executor);
+    }
+
+    /**
+     * Used by constructors
+     */
+    private void init(HttpCommandExecutor executor) {
+        this.executeMethod = new AppiumExecutionMethod(this);
+        locationContext = new RemoteLocationContext(executeMethod);
+        super.setErrorHandler(errorHandler);
+        this.remoteAddress = executor.getAddressOfRemoteServer();
+        this.setElementConverter(new JsonToMobileElementConverter(this, this));
     }
 
     /**
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -139,9 +139,8 @@ public class AppiumDriver<T extends WebElement>
         this.setSessionId(sessionId);
         
         AppiumCommandExecutor executor = new AppiumCommandExecutor(MobileCommand.commandRepository, remoteAddress);
-        executor.setCommandCodec(new AppiumW3CHttpCommandCodec());
-        executor.setResponseCodec(Dialect.W3C.getResponseCodec());
         executor.configureW3CMode();
+        executor.setResponseCodec(Dialect.W3C.getResponseCodec());
         
         this.setCommandExecutor(executor);
         init(executor);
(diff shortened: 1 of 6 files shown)


src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -129,6 +127,35 @@ public AppiumDriver(HttpClient.Factory httpClientFactory, Capabilities desiredCa
...
+    /**
+     * For creating a client connected to an existing Appium session.
+     * The other constructors all create a new session, sending the POST /session request.
+     * This constructor does not.
+     */
+    public AppiumDriver(URL remoteAddress, String sessionId) {
+        this.setSessionId(sessionId);
+        
+        AppiumCommandExecutor executor = new AppiumCommandExecutor(MobileCommand.commandRepository, remoteAddress);
+        executor.setCommandCodec(new AppiumW3CHttpCommandCodec());
```
mykola-mokhnach: how do we know that this is a w3c sessions?

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -129,6 +127,35 @@ public AppiumDriver(HttpClient.Factory httpClientFactory, Capabilities desiredCa
...
+        executor.setResponseCodec(Dialect.W3C.getResponseCodec());
+        executor.configureW3CMode();
```
mykola-mokhnach: both `configureW3CMode` and this method call `setCommandCodec`

src/main/java/io/appium/java_client/AppiumDriver.java
Jonahss: I'd like to limit this feature to W3C sessions only. I had a nice error before but refactored it away, I'll find a way to surface an error if it's not W3C. Any idea how?

src/test/java/io/appium/java_client/ios/ConstructorTest.java
```diff
@@ -0,0 +1,60 @@
...
+import org.openqa.selenium.support.ui.WebDriverWait;
+
+import java.net.URL;
+import java.time.Duration;
+
+public class ConstructorTest extends AppIOSTest {
+
+    @Test
+    public void attachToSessionTest() {
+		String sessionId = driver.getSessionId().toString();
```
mykola-mokhnach: indentation please

src/main/java/io/appium/java_client/AppiumDriver.java
Jonahss: Good point. I refactored at the last second. Will remove that line.

src/test/java/io/appium/java_client/ios/ConstructorTest.java
Jonahss: you got it. sorry.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -260,10 +260,14 @@ public Response execute(Command command) throws WebDriverException {
 
         if (DriverCommand.NEW_SESSION.equals(command.getName())
                 && getCommandCodec() instanceof W3CHttpCommandCodec) {
-            setCommandCodec(new AppiumW3CHttpCommandCodec());
-            getAdditionalCommands().forEach(this::defineCommand);
+        	configureW3CMode();
```
mykola-mokhnach: wrong indentation here

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -136,16 +136,16 @@ protected void setPrivateFieldValue(String fieldName, Object newValue) {
         return getPrivateFieldValue("additionalCommands", Map.class);
     }
 
-    protected CommandCodec<HttpRequest> getCommandCodec() {
```
mykola-mokhnach: I'd rather keep them protected if possible

mykola-mokhnach: at least setters

Jonahss: How could that be achieved here?

Jonahss: Oh, I mean. I figured that as long as the setters were made public, you might as well get access to the getters. It's not a big secret.

Jonahss:
noticed my own TODO. need to add this to IOSDriver and AndroidDriver too.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: this is not a secret, but rather protection from shutting your leg. Setting an incorrect codec in runtime will break the driver completely. 
Also I expect public methods to be properly documented

mykola-mokhnach: these can also be made package-private

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -260,10 +260,14 @@ public Response execute(Command command) throws WebDriverException {
...
                 && getCommandCodec() instanceof W3CHttpCommandCodec) {
-            setCommandCodec(new AppiumW3CHttpCommandCodec());
-            getAdditionalCommands().forEach(this::defineCommand);
+            configureW3CMode();
         }
 
         return response;
     }
+    
+    public void configureW3CMode() {
```
mykola-mokhnach: configureW3CMode -> I would rather change it to `configureCooandCodec` and pass the codec instance as an argument

src/main/java/io/appium/java_client/AppiumDriver.java
mykola-mokhnach: the simplest way would be to get /status response and analyse its format. The response in W3C format does not contain response code field (the same is done in appium base driver)

Jonahss: So the current code sets this from the initial response to create a session. The new code never sends that request, so it needs to be set. BUT, the new code calls GET /session/:sessionId when setting the JsonToMobile converter thing. So the codec needs to be set before the constructor finishes.

I could change the executor code to set this if the codecs are null, rather than just on create session. I'll try that.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
Jonahss: The executor and the AppiumDriver are in different packages though :(

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
Jonahss: Sure. Makes sense.

src/main/java/io/appium/java_client/AppiumDriver.java
mykola-mokhnach: I would try to make the request directly using okhttp wrapper

Jonahss: So I was going to change the execute method to set the codecs if they are currently null, rather than only if it's a create_session command. BUT, in order to send and encode any command, it needs codecs already. So there's no way to capture the protocol type when the first command is sent, because the command can't be sent :(

options I can think of:
 - make this a W3C-only feature
 - remove jsonwire protocol support in a major version
 - add a third term to the constructor, requiring the user pass in a dialect (I wanted to avoid this because it's an extra complication)

Any idea when we can stop supporting json wire?

Jonahss: I ran the test with the `MobileCapabilityType.FORCE_MJSONWP` desired cap, and the test still passed.

chawlashashi10:
@Jonahss 
Please help me to integrate add a new constructor for creating a client attached to an existing session in Appium dotnet

ahmadsayeed3:
Hi Mykola, I am not sure If I should write this here. 
This sounds like a very useful feature for us and these days I am looking for implementation which can share sessions. 

Unfortunately, I don't think I am that much capable yet to contribute on this.  May I know when we can expect this merge.

mykola-mokhnach:
@ahmadsayeed3 Currently no one is working on this PR. It's a bit more complicated that it might look like. Consider hiring a consultant if there's a need in such functionality. I could help with reviews.

kandji-joel:
@mykola-mokhnach any idea if this is something that will ever be supported? I'm using the python client and would like to instantiate a driver object by attach to an existing session

PR #1216, opened 2019-08-26
fix: Avoid using getSession call for capability values retrieval

mykola-mokhnach:
## Change list

W3C protocol does not contain getSession API in its spec, which means all calls to `HasSessionDetails` interface methods will end up with `no such method` error. As a result, I have replaced all the "static" calls  with simple capabilities extraction and the "dynamic" call uses the fact that none of native mobile drivers support javascript execution, - this only works in browser context.
 
## Types of changes

## Details

See https://github.com/appium/appium/issues/13124#issuecomment-524850823


mykola-mokhnach:
The error in e2e is a flakiness issue

SrinivasanTarget:
> The error in e2e tests is a CI flakiness issue

Yeah, I did notice in master. Will fix it asap.

Issue #1236, opened 2019-09-15
Doc: AndroidMobileCapabilityType: missing chromedriverPort/chromedriverPorts


amedvedjev:
## Environment

* java client build version: https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java

PR #1238, opened 2019-09-16
Update AndroidMobileCapabilityType

a-en:
## Change list

Added missing capabilities from [docs](http://appium.io/docs/en/writing-running-appium/caps/) to AndroidMobileCapabilityType interface. Also, this will fix [this issue](https://github.com/appium/java-client/issues/1236)
 
## Types of changes

## Details

### DIFF
--- a/src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
+++ b/src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
@@ -57,6 +57,7 @@ public interface AndroidMobileCapabilityType extends CapabilityType {
 
     /**
      * Timeout in milliseconds used to wait for the appWaitActivity to launch (default 20000).
+     * @since 1.6.0
      */
     String APP_WAIT_DURATION = "appWaitDuration";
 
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
```diff
@@ -276,4 +373,54 @@
...
+    String UNLOCK_TYPE = "unlockType";
+
+    /**
+     * A key pattern to unlock used by {@code unlockType}.
+     */
+    String UNLOCK_KEY = "unlockKey";
+
+    /**
+     * Initializing the app under test automatically.
+     * Appium does not install/launch the app under test if this is {@code false}. Defaults to {@code true}
```
mykola-mokhnach: it does install the app, but skips the launch

mykola-mokhnach:
Thanks. It might also be useful to mention since which Appium version these caps are available. (aka `@since` doc option)

src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
a-en: @mykola-mokhnach I've changed description to 'Appium does not launch the app...'. Should it be also changed in http://appium.io/docs/en/writing-running-appium/caps/ ?

a-en:
@mykola-mokhnach  I've went through releases of Appium on github and collected '@since' for those caps I was able to find. Are there any other sources that might be used to add this information?

mykola-mokhnach:
Usually version changelist is the only source to find it out

src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
mykola-mokhnach: yes, it makes sense to update the description in the documentation

PR #1262, opened 2019-11-04
feat: add support for logCustomEvent command

jlipps:
Depends on https://github.com/appium/appium-base-driver/pull/364

(I couldn't get tests to run due to some Class Not Found error that seems unrelated.)

### DIFF
--- a/src/main/java/io/appium/java_client/LogsEvents.java
+++ b/src/main/java/io/appium/java_client/LogsEvents.java
@@ -16,10 +16,19 @@
 
 package io.appium.java_client;
 
+import static io.appium.java_client.MobileCommand.GET_EVENTS;
 import static io.appium.java_client.MobileCommand.LOG_EVENT;
 
-import java.util.HashMap;
+import com.google.common.collect.ImmutableMap;
+import io.appium.java_client.serverevents.CommandEvent;
+import io.appium.java_client.serverevents.CustomEvent;
+import io.appium.java_client.serverevents.TimedEvent;
+import io.appium.java_client.serverevents.ServerEvents;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import org.openqa.selenium.json.Json;
+import org.openqa.selenium.remote.Response;
 
 public interface LogsEvents extends ExecutesMethod {
 
@@ -27,14 +36,45 @@ public interface LogsEvents extends ExecutesMethod {
      * Log a custom event on the Appium server.
      *
      * @since Appium 1.16
-     * @param vendor the name of the vendor prefix to avoid event collisions
-     * @param eventName the name of the event
+     * @param event the event to log
      * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
      */
-    default void logEvent(String vendor, String eventName) {
-        Map<String, Object> data = new HashMap<>();
-        data.put("vendor", vendor);
-        data.put("event", eventName);
-        execute(LOG_EVENT, data);
+    default void logEvent(CustomEvent event) {
+        execute(LOG_EVENT, ImmutableMap.of("vendor", event.getVendor(), "event", event.getEventName()));
+    }
+
+    /**
+     * Log a custom event on the Appium server.
+     *
+     * @since Appium 1.16
+     * @return ServerEvents object wrapping up the various command and event timestamps
+     * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
+     */
+    default ServerEvents getEvents() {
+        Response response = execute(GET_EVENTS);
+        String jsonData = new Json().toJson(response.getValue());
+
+        //noinspection unchecked
+        Map<String, Object> value = (Map<String, Object>) response.getValue();
+
+        //noinspection unchecked
+        List<CommandEvent> commands = ((List<Map<String, Object>>) value.get("commands"))
+            .stream()
+            .map((Map<String, Object> cmd) -> new CommandEvent(
+                (String) cmd.get("cmd"),
+                ((Long) cmd.get("startTime")),
+                ((Long) cmd.get("endTime"))
+            ))
+            .collect(Collectors.toList());
+
+        List<TimedEvent> events = value.keySet().stream()
+            .filter((String name) -> !name.equals("commands"))
+            .map((String name) -> {
+                //noinspection unchecked
+                return new TimedEvent(name, (List<Long>) value.get(name));
+            })
+            .collect(Collectors.toList());
+
+        return new ServerEvents(commands, events, jsonData);
     }
 }
--- a/src/main/java/io/appium/java_client/LogsEvents.java
+++ b/src/main/java/io/appium/java_client/LogsEvents.java
@@ -1,0 +1,40 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client;
+
+import static io.appium.java_client.MobileCommand.LOG_EVENT;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public interface LogsEvents extends ExecutesMethod {
+
+    /**
+     * Log a custom event on the Appium server.
+     *
+     * @since Appium 1.16
+     * @param vendor the name of the vendor prefix to avoid event collisions
+     * @param eventName the name of the event
+     * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
+     */
+    default void logEvent(String vendor, String eventName) {
+        Map<String, Object> data = new HashMap<>();
+        data.put("vendor", vendor);
+        data.put("event", eventName);
+        execute(LOG_EVENT, data);
+    }
+}
--- a/src/main/java/io/appium/java_client/serverevents/CommandEvent.java
+++ b/src/main/java/io/appium/java_client/serverevents/CommandEvent.java
@@ -1,0 +1,10 @@
+package io.appium.java_client.serverevents;
+
+import lombok.Data;
+
+@Data
+public class CommandEvent {
+    public final String name;
+    public final long startTimestamp;
+    public final long endTimestamp;
+}
--- a/src/main/java/io/appium/java_client/serverevents/ServerEvents.java
+++ b/src/main/java/io/appium/java_client/serverevents/ServerEvents.java
@@ -1,0 +1,19 @@
+package io.appium.java_client.serverevents;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class ServerEvents {
+
+    public final List<CommandEvent> commands;
+    public final List<TimedEvent> events;
+    public final String jsonData;
+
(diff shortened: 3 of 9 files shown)


src/main/java/io/appium/java_client/LogsEvents.java
```diff
@@ -0,0 +1,40 @@
...
+     * @since Appium 1.16
+     * @param vendor the name of the vendor prefix to avoid event collisions
+     * @param eventName the name of the event
+     * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
+     */
+    default void logEvent(String vendor, String eventName) {
+        Map<String, Object> data = new HashMap<>();
+        data.put("vendor", vendor);
+        data.put("event", eventName);
+        execute(LOG_EVENT, data);
```
mykola-mokhnach: `execute(LOG_EVENT, ImmutableMap.of("vendor", vendor, "event", eventName))`

src/test/java/io/appium/java_client/android/LogEventTest.java
```diff
@@ -0,0 +1,42 @@
...
+import static org.hamcrest.core.Is.is;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertThat;
+
+public class LogEventTest extends BaseAndroidTest {
+
+    @Test
+    public void verifyLoggingCustomEvents() {
+        driver.logEvent("appium", "funEvent");
+        Object value = driver.getSessionDetail("events");
```
mykola-mokhnach: session details is not a part of W3C spec. Should we have events under "mobile:" endpoint?

jlipps: hmm. the problem with the mobile endpoints is that they are supported in individual drivers, not basedriver, i think. i'm not sure right now of a clean way to cascade mobile calls from basedriver to subdrivers. we'd have to pull a lot of logic in from other drivers and clean it up somehow.

mykola-mokhnach: perhaps, similarly to how we use executeDriverScript

jlipps: @mykola-mokhnach to make this functionality available in w3c, should we create a `GET /session/:sessionid/appium/log_events` method?

mykola-mokhnach: yes, we would need something like that. Also, the method might require additional filtering parameters, so I don't know if it makes sense to change the method to POST

jlipps: https://github.com/appium/appium-base-driver/pull/365

jlipps: ok, i created the various java objects needed for first class support of that route.

jlipps:
The Azure failure is a timeout of some kind and it seems to have existed since before this PR; not sure what to do about that.

src/main/java/io/appium/java_client/LogsEvents.java
```diff
@@ -0,0 +1,84 @@
...
+
+    /**
+     * Log a custom event on the Appium server.
+     *
+     * @since Appium 1.16
+     * @param vendor the name of the vendor prefix to avoid event collisions
+     * @param eventName the name of the event
+     * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
+     */
+    default void logEvent(String vendor, String eventName) {
```
mykola-mokhnach: should the argument be also an object here?

src/main/java/io/appium/java_client/LogsEvents.java
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Log a custom event on the Appium server.
+     *
+     * @since Appium 1.16
+     * @return ServerEvents object wrapping up the various command and event timestamps
+     * @throws org.openqa.selenium.WebDriverException if there was a failure while executing the script
+     */
+    default ServerEvents getEvents() {
+        Response response = execute(GET_EVENTS);
+        ArrayList<CommandEvent> commands = new ArrayList<>();
```
mykola-mokhnach: the basic type should be an interface (`List`) rather than concrete type (`ArrayList`)

mykola-mokhnach: same below

src/main/java/io/appium/java_client/LogsEvents.java
```diff
@@ -0,0 +1,84 @@
...
+        ArrayList<ServerEvent> events = new ArrayList<>();
+        String jsonData = new Json().toJson(response.getValue());
+        //noinspection unchecked
+        Map<String, Object> value = (Map<String, Object>) response.getValue();
+        for (String name : value.keySet()) {
+            if (name.equals("commands")) {
+                //noinspection unchecked
+                ArrayList<Map<String, Object>> cmds = (ArrayList<Map<String, Object>>) value.get(name);
+                for (Map<String, Object> cmd : cmds) {
+                    commands.add(new CommandEvent(
```
mykola-mokhnach: why cannot we use collect here?

src/main/java/io/appium/java_client/LogsEvents.java
```diff
@@ -0,0 +1,84 @@
...
+                        (String) cmd.get("cmd"),
+                        ((Long) cmd.get("startTime")).intValue(),
+                        ((Long) cmd.get("endTime")).intValue()
+                    ));
+                }
+            } else {
+                //noinspection unchecked
+                ArrayList<Long> timestamps = (ArrayList<Long>) value.get(name);
+                List<Integer> intStamps = timestamps
```
mykola-mokhnach: why cannot we keep timestamps as Long?

src/main/java/io/appium/java_client/serverevents/CommandEvent.java
```diff
@@ -0,0 +1,13 @@
+package io.appium.java_client.serverevents;
+
+public class CommandEvent {
+    public final String name;
+    public Integer startTimestamp;
```
mykola-mokhnach: why not final?

src/main/java/io/appium/java_client/serverevents/CommandEvent.java
```diff
@@ -0,0 +1,13 @@
...
+    public final Integer endTimestamp;
```
mykola-mokhnach: can the value be null, that we use Integer instead of int?

src/main/java/io/appium/java_client/serverevents/CommandEvent.java
```diff
@@ -0,0 +1,13 @@
+package io.appium.java_client.serverevents;
+
+public class CommandEvent {
```
mykola-mokhnach: Consider using Lombok annotation for data class creation

src/main/java/io/appium/java_client/serverevents/ServerEvent.java
```diff
@@ -0,0 +1,13 @@
+package io.appium.java_client.serverevents;
+
+import java.util.List;
+
+public class ServerEvent {
```
mykola-mokhnach: same above

src/test/java/io/appium/java_client/android/LogEventTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+public class LogEventTest extends BaseAndroidTest {
+
+    @Test
+    public void verifyLoggingCustomEvents() {
+        driver.logEvent("appium", "funEvent");
+        ServerEvents events = driver.getEvents();
+        boolean hasCustomEvent = false;
+        boolean hasCommandName = false;
+        for (ServerEvent event : events.events) {
```
mykola-mokhnach: consider using streams + filter + anyMatch

src/main/java/io/appium/java_client/serverevents/ServerEvents.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+public class ServerEvents {
+
+    public final List<CommandEvent> commands;
+    public final List<ServerEvent> events;
```
mykola-mokhnach: this naming looks a bit unlogical to me. We have ServerEvents class, but it consist of lists with two absolutely different types: CommandEvent and ServerEvent. Perhaps, we should change the `ServerEvent` to something else to avoid confusion?

src/main/java/io/appium/java_client/serverevents/CommandEvent.java
jlipps: I initially did this, but the project wouldn't build. I did it exactly as we do with the execute driver script results, but for some reason it did not work.

src/main/java/io/appium/java_client/serverevents/ServerEvents.java
jlipps: i call it ServerEvents because it is the response to the "get server events" call. Definitely open to suggestions for a different name. Maybe instead of ServerEvent we could call it TimedEvent?

src/main/java/io/appium/java_client/LogsEvents.java
jlipps: seems a bit overkill to me, but if you feel strongly i can make it one.

src/main/java/io/appium/java_client/LogsEvents.java
jlipps: 👍

src/test/java/io/appium/java_client/android/LogEventTest.java
jlipps: 👍

jlipps: java is actually much more pleasant with these functional bits.

src/main/java/io/appium/java_client/LogsEvents.java
mykola-mokhnach: the function would be easier to extend if it would be an object. Although, I don't have strong opinion on that

src/main/java/io/appium/java_client/serverevents/TimedEvent.java
```diff
@@ -0,0 +1,13 @@
+package io.appium.java_client.serverevents;
+
+import java.util.List;
+
+public class TimedEvent {
```
mykola-mokhnach: I would still consider using lombok. Otherwise  I would prefer removing this dependency completely if it cannot be used

src/main/java/io/appium/java_client/serverevents/CommandEvent.java
```diff
@@ -0,0 +1,13 @@
...
+    public final String name;
+    public final Long startTimestamp;
+    public final Long endTimestamp;
+
+    public CommandEvent(String name, Long startTimestamp, Long endTimestamp) {
```
mykola-mokhnach: can we use long instead of Long?

jlipps: why I try that I am unable to decompose results into `List<long>` because list cannot take a primitive type.

jlipps: n/m seems to work in the definition anyway.

mykola-mokhnach: have you heard about autoboxing in Java? ;)

jlipps: remember, i am not the world's best java person.

mykola-mokhnach: not yet, not yet :)

src/main/java/io/appium/java_client/serverevents/ServerEvents.java
```diff
@@ -0,0 +1,19 @@
...
+import java.nio.file.Path;
+import java.util.List;
+import lombok.Data;
+
+@Data
+public class ServerEvents {
+
+    public final List<CommandEvent> commands;
+    public final List<TimedEvent> events;
+    public final String jsonData;
```
mykola-mokhnach: should it be string or JSONObject?

jlipps: string, because its only purpose is to write out to disk.

Issue #1269, opened 2019-11-20
AndroidFindBy View Tag Support


iChip:
I'd like to use the @AndroidFindBy annotation for finding elements by view tag on Android, I am specifying a test ID through this attribute as we cannot use resource ids for targeting views when using React Native.

It does not seem like @AndroidFindBy currently supports this. I see MobileBy.ByAndroidViewTag, but I believe this cannot be used with the page object model to initially locate elements.

Is the only workaround an xpath selector (which I believe requires knowing the tag name in advance in addition to the view tag)?
e.g. driver.findElementByXPath("//android.widget.TextView[@view-tag='some-tag']"

iChip:
Tagging @dpgraham since he's my guy on the inside + [appears to have some context](https://github.com/appium/appium/issues/6025#issuecomment-322266065)

jayandran-Sampath:
@iChip : As per current implementation, AndroidFindBy(tagName="some-tag") itself will work.  Just give a try and let us know.

iChip:
Thanks, It does work, but I am confused by the naming. Your solution seems more sensible, the existing tagName annotation does not seem to correctly describe this locator strategy.

I added some comments on your closed PR:
https://github.com/appium/java-client/pull/1271#discussion_r356983682

vaishaliidnani:
is this project run in Android because i m not able to run this please give the link or share the project .

VivekShahare04:
@iChip @dpgraham try to Create a custom annotation:

just like 
```
@Retention(RetentionPolicy.RUNTIME)
@Target({ElementType.FIELD, ElementType.TYPE})
public @interface AndroidFindByViewTag {
    String value();
}
```

ranjeetksingh55:
Hi @iChip,

In React Native apps where resource-id isn’t available, we often rely on view-tag (mapped from testID). While MobileBy.AndroidViewTag("some-tag") works, it’s not usable with @AndroidFindBy, limiting Page Object Model usage.

Currently, the workaround is using XPath like:
`@AndroidFindBy(xpath = "//android.widget.TextView[@view-tag='some-tag']")
`But this requires knowing the class name, making it fragile.

Support for something like **@AndroidFindBy(viewTag = "some-tag")** would be a clean solution. 
Happy to contribute if this is feasible!

Thanks 🙌

PR #1271, opened 2019-11-23
AndroidFindBy annotation support for android view-tag selector

jayandran-Sampath:
## Change list

Added support to provide Android view-tag selector in PageFactory-AndroidFindBy annotation
 
## Types of changes

## Details

Related to #1269 

Android View-tag support in Android Pagefactory

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -96,7 +96,7 @@ public @interface AndroidFindBy {
     int priority() default 0;
 
     /**
-     * Its Android View Tag element value
+     * Its Android View Tag element value.
      *
      * @return a android view tag value to the target element
      */
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -94,4 +94,11 @@ public @interface AndroidFindBy {
      * @return priority of the searching
      */
     int priority() default 0;
+
+    /**
+     * Its Android View Tag element value
+     *
+     * @return a android view tag value to the target element
+     */
+    String androidViewTag() default "";
 }
--- a/src/main/java/io/appium/java_client/pagefactory/bys/builder/Strategies.java
+++ b/src/main/java/io/appium/java_client/pagefactory/bys/builder/Strategies.java
@@ -109,8 +109,8 @@ enum Strategies {
         }
     },
 
-    BY_ANDROID_VIEWTAG("androidViewTag"){
-        @Override By getBy(Annotation annotation){
+    BY_ANDROID_VIEWTAG("androidViewTag") {
+        @Override By getBy(Annotation annotation) {
             String value = getValue(annotation, this);
             if (annotation.annotationType().equals(AndroidFindBy.class)
                     || annotation.annotationType().equals(AndroidBy.class)) {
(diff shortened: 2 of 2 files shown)


src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
```diff
@@ -94,4 +94,11 @@
      * @return priority of the searching
      */
     int priority() default 0;
+
+    /**
+     * Its Android View Tag element value.
+     *
+     * @return a android view tag value to the target element
+     */
+    String androidViewTag() default "";
```
mykola-mokhnach: isn't it the same as `tagName`?

jayandran-Sampath: Hi, both does different work. There is a example for this in #1269.

mykola-mokhnach: I still have doubts on that. AndroidFindBy(tagName) should work just fine. Is there a reason to create another synonym?

saikrishna321: @jayandran-Sampath As @mykola-mokhnach says this should work

```
@AndroidFindBy(tagName = "something")
```

jayandran-Sampath: @saikrishna321 & @mykola-mokhnach sure i will check once from my side and see how it works and then we can close this PR if its not needed. 

Just to be clear, I did this PR to enable PageFactory Annotation option for Locator Strategy -  Android View Tag (Espresso only) & also i see TagName method points to default Selenium By.TagName.

jayandran-Sampath: @mykola-mokhnach & @saikrishna321 Thanks both for clearing it.. Tried with hello world app, as you guys said its working with AndroidFindBy(TagName) locator itself.

Just a thought, Internally using Tagname annotation we will still depend on Core By.tagName Locator Stragey to make it work but having seperate androidViewTag annotation will actually represent MobileBy.AndroidView Locator Stragey so in future this independent tag will work if there is any change in finding the element in espresso. 

image

If still we feel it's not required, We can close this PR.

SrinivasanTarget: @jayandran-Sampath Sorry for getting back late on this. I think we should not add one more locator strategy instead we should have test specific to `android view` tag.

jayandran-Sampath: okay @SrinivasanTarget . i will close this PR. Thanks

iChip: If this is your decision, documenting this functionality for users would be helpful. I cannot locate documentation for the tagName annotation.

According to the AndroidFindBy javado, tagName returns "A desired element tag".

TBH, I would have expected the annotation to be called viewTag for consistency, 'name' some times refers to accessibility or 'tag name' to me suggests class (tag) name. Since this query will only be supported on a single driver, this naming seems problematic/unintuitive. What does tagName do on UIAutomator2?

e.g. [findElementByName/By.Name clearly resolves to content-desc (accessibility)](https://stackoverflow.com/a/26989612)


e.g. in [these docs](https://appium.readthedocs.io/en/latest/en/commands/element/attributes/name/) a tag name also corresponds to an accessibility label

PR #1311, opened 2020-03-05
FindBy doc update

Cliabhach:
## Change list

* New `Advanced-By.md` file for advanced AndroidBy & iOSXCUITFindBy usage
* Updated javadoc links for UiAutomator and Class Chain concepts
 
## Types of changes

## Details

The main goal here was to document these `uiAutomator` and `iOSClassChain` attributes _within_ the java client documentation. You'll note that the README is incomplete - I tried to use [the server source code](https://github.com/appium/appium-uiautomator2-server/blob/514ff583d18e27d0335fcc86b349656d44504e0b/app/src/main/java/io/appium/uiautomator2/utils/UiScrollableParser.java) to figure out how UiScrollable works, but didn't get very far.

I did find [these](https://github.com/appium/appium/blob/master/commands-yml/commands/element/find-elements.yml) [two files](https://github.com/appium/appium/blob/master/docs/en/drivers/android-uiautomator2.md) in the main appium source. Should I make a pull request to update either of those instead/as well?


src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,69 @@
...
+Our XCUiTest integration has full support for the 'Class Chain' concept. This
+can do much of what XPath does...but faster.
+
+### External References
+
+Refer to [the official WebDriverAgent query docs](https://github.com/facebookarchive/WebDriverAgent/wiki/Class-Chain-Queries-Construction-Rules)
+to learn more about the general concept.
+
+### Sample usage:
+
```
mykola-mokhnach: please use markdown formatters for code blocks: https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks

mykola-mokhnach:
Thanks for the PR @Cliabhach , the documentation is never enough

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Certainly. I'll get right on that.

Cliabhach:
Also, it looks like appium has its own [WebDriverAgent](https://github.com/appium/webdriveragent) repository - should I make the inline links point there instead of to facebook's old project?

mykola-mokhnach:
> Also, it looks like appium has its own WebDriverAgent repository - should I make the inline links point there instead of to facebook's old project?

Yes, we maintain our own fork. Although, it only contains the code. Wiki does not get copied

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,75 @@
...
+```java
+// Selector for image elements
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeImage")
+
+// Selector for every cell with the name 'Foo'
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name == \"Foo\"`]")
```
mykola-mokhnach: NSPredicate allows to use both double and single quotes in the equal manner. In java strings we could use the single ones to avoid unnecessary escaping.

mykola-mokhnach: one should also be careful if the quoted string is a variable, which can also contain the same type of quotes that are used in the predicate. In such case this would be a developer's responsibility to take care quotes escaping inside the string variable. See, for example: https://stackoverflow.com/questions/14116217/nspredicate-single-quote-issue

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,75 @@
...
+and
+* [UiSelector API docs](https://developer.android.com/reference/androidx/test/uiautomator/UiSelector)
+
+### Sample Strings
+
+Here are some ways you could configure a UiSelector in your project:
+
+```java
+// Create a selector that looks for the text "Hello World":
+@AndroidFindBy(uiAutomator = "new UiSelector().text(\"Hello World\")")
```
mykola-mokhnach: I can observe that people are mostly looking for methods to scroll into views, like https://discuss.appium.io/t/cant-get-setmaxsearchswipes-to-work-with-driver-findelement-mobileby-androiduiautomator/29935

It would be nice to put some examples of such locators here as well

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Good points. I'll try to present one or two more examples with different quoting styles.

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Right - I'll need to read up on that.

Cliabhach:
I'll need to add more examples, I think, but how does that look so far?

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,133 @@
...
+// Selector for every cell with the name "Foo" (double quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name == \"Foo\"`]")
+
+// Selector for every cell with a name that starts with 'Foo' (single quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name BEGINSWITH 'Foo'`]")
+// Selector for every cell with a name that starts with "Foo" (double quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name BEGINSWITH \"Foo\"`]")
+
+// Selector for every cell with a name that starts with "it's not"
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name BEGINSWITH \"it's not\"`]")
```
mykola-mokhnach: class chain also supports indexing, including negative indexes (e.g. -1 means the last matched element, -2 - the one before the last one, etc) . The only limitation is that the index should be the last specifier in the chain, for example ```**/XCUIElementTypeCell[`name == 'bla'`][-2]``` is valid, but ```**/XCUIElementTypeCell[-2][`name == 'bla'`]``` is not

mykola-mokhnach:
yep, the doc looks great. I've also added a comment regarding indexing and deep attribute search

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+// Selector for every cell with the name 'Foo' (single quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name == 'Foo'`]")
+// Selector for every cell with the name "Foo" (double quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name == \"Foo\"`]")
+
+// Selector for every cell with a name that starts with 'Foo' (single quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name BEGINSWITH 'Foo'`]")
+// Selector for every cell with a name that starts with "Foo" (double quote style)
+@iOSXCUITFindBy(iOSClassChain = "**/XCUIElementTypeCell[`name BEGINSWITH \"Foo\"`]")
```
mykola-mokhnach: there is also a possibility to perform deep search by attributes. predicates enclosed in backticks will only look through attributes of the current element, whenever predicates enclosed with dollar sign `$` will look through all descendant attributes, for example: ```**/XCUIElementTypeCell[`name == 'Foo'`]``` looks for cells with name `Foo`, but ```**/XCUIElementTypeCell[$name == 'Foo'$]``` looks for cells, which contain at least one element having the name `Foo`

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: These are 1-based indices, it looks like. What does an index of 0 do?

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Oh, interesting.

Cliabhach:
Ok, I've added items for those. I'm still not 100% on putting the README in this folder - should we cross-reference it from [the main uiautomator2 Appium docs](https://github.com/appium/appium/blob/master/docs/en/drivers/android-uiautomator2.md)?

Cliabhach:
Oh, and I should rebase these commits so they have the `docs:` prefix.....

src/main/java/io/appium/java_client/pagefactory/README.md
mykola-mokhnach: throws an error :)

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,160 @@
...
+// Followed by `
+// End with ]
+[`label != 'a'`]
+[`isWDVisible == 1`]
+[`value < 0`]
+```
+
+#### Searching Descendents
+
+If you replace the backticks (`) around a predicate with dollar signs ($), then
```
mykola-mokhnach: backticks are also used for GitHub markup. Please escape them properly

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,160 @@
...
+the Class Chain will match against the children, grandchildren, and other
+descendents of each element.
```
mykola-mokhnach: descendants

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,160 @@
...
+XCUIElementTypeCell[`label == 'here'`]
+// Find a cell element which contains SOMETHING ELSE that has the label 'here'
+XCUIElementTypeCell[$label == 'here'$]
+```
+
+#### Handling Quote Marks
+
+Most of the time, you can treat pairs of single quotes or double quotes
+interchangably. If you're searching with a string that contains quote marks,
+though, you'll [want to be careful](https://stackoverflow.com/q/14116217).
```
mykola-mokhnach: ...you need to be careful

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: If I knew how I would....the only thing I can think of is an html entity like &#96;. Is that allowed?

Cliabhach: I found a github issue https://github.com/github/markup/issues/363 with some sort of guidance. I'll try using double backticks around it, like so: `` ` ``

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Changed!

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Done.

src/main/java/io/appium/java_client/pagefactory/README.md
Cliabhach: Fair enough.

src/main/java/io/appium/java_client/pagefactory/README.md
```diff
@@ -0,0 +1,160 @@
+# Standard Selectors
```
mykola-mokhnach: Please move this document to `docs` folder. There should be no documents inside java sources

Cliabhach: I can do that.

Cliabhach:
Not sure about the name - I don't suppose anyone has a good idea for what to call it?

mykola-mokhnach:
the current name is ok. We can always rename it if necessary ;)

PR #1319, opened 2020-03-29
implement driver options

titusfortner:
## Change list
* Added `IOSMobileOptions` and `AndroidMobileOptions` for easier capability generation
 
## Types of changes

## Details

This is to follow the same approach that Selenium Java bindings have taken with their Browser Options classes. Since `DesiredCapabilities` will soon be deprecated, users will either need to update to `MutableCapabilities` or otherwise use an Options class like this. An Options class has the advantage of encapsulating specific behaviors to IOS and Android.

Note that this is a draft to get feedback before I implement all of the IOS & Android specific capabilities.

Honestly, this would be a lot less time consuming if we do some reflection magic with Java 9, but I don't think Appium is at the point where it is willing to require > 8?


src/main/java/io/appium/java_client/remote/AndroidMobileOptions.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client.remote;
+
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.remote.CapabilityType;
+
+public class AndroidMobileOptions extends MobileOptions<AndroidMobileOptions> {
```
mykola-mokhnach: I would say platform-specific option classes should go to android/ios packages

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,211 @@
...
+package io.appium.java_client.remote;
+
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.MutableCapabilities;
+import org.openqa.selenium.ScreenOrientation;
+import org.openqa.selenium.remote.CapabilityType;
+
+import java.net.URL;
+
+public abstract class MobileOptions<T extends MobileOptions<T>> extends MutableCapabilities {
```
mykola-mokhnach: I wouldn't make it abstract. That way we could still use this class for newly attached drivers without modifying java client

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,211 @@
...
+
```
mykola-mokhnach: do we need toString implementation for quick debug?

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,211 @@
...
+package io.appium.java_client.remote;
+
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.MutableCapabilities;
+import org.openqa.selenium.ScreenOrientation;
+import org.openqa.selenium.remote.CapabilityType;
+
+import java.net.URL;
+
```
mykola-mokhnach: I personally like the concept of capabilities builder more like MutableCapabilities one. It would ensure more safety during the session life cycle, since capabilities, once passed to the session constructor are immutable in principle. And changing them in the code does not affect the session at all.

src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
```diff
@@ -134,4 +134,16 @@
      * or Safari's (on iOS) performance logging (default {@code false}).
      */
     String ENABLE_PERFORMANCE_LOGGING = "enablePerformanceLogging";
+
+
+    /**
+     * App or list of apps (as a JSON array) to install prior to running tests. Note that it will not work with
+     * automationName of Espresso and iOS real devices.
+     */
+    String OTHER_APPS = "otherApps";
```
mykola-mokhnach: how about extracting this change into a separate PR?

src/main/java/io/appium/java_client/remote/AndroidMobileOptions.java
titusfortner: Makes sense, I put them in `remote` because that's where `IOSMobileCapabilityType` and `AndroidMobileCapabilityType` were.

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Superclass `AbstractCapabilities` implementation should be sufficient: https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/AbstractCapabilities.java#L142

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: I get what you are saying, but there is value in following the same pattern as Selenium. There's also no reason you couldn't provide multiple options for users if you choose..

src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
titusfortner: #1323

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 06815760ab95fe3ecb0e3aae841bcf4b8ae1b3d5

src/main/java/io/appium/java_client/remote/AndroidMobileOptions.java
titusfortner: 3dd48dd54f1c7c6e54202b7e473a380e0e79424b

src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
```diff
@@ -124,7 +124,7 @@
     String EVENT_TIMINGS = "eventTimings";
```
mykola-mokhnach: Please revert this file. It has been merged in the other PR

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,219 @@
...
+    public MobileOptions(Capabilities source) {
+        this();
+        merge(source);
+    }
+
+    public String getPlatformName() {
+        return (String) getCapability(CapabilityType.PLATFORM_NAME);
+    }
+
+    public T setApp(String url) {
```
mykola-mokhnach: it would be nice to have some docstrings

mykola-mokhnach: same below

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,219 @@
...
+
+import java.net.URL;
+
+public class MobileOptions<T extends MobileOptions<T>> extends MutableCapabilities {
+
+    public MobileOptions() {
+    }
+
+    public MobileOptions(Capabilities source) {
+        this();
```
mykola-mokhnach: is this call needed?

src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
titusfortner: Sure, but git magics it away when/if you squash/rebase. :)

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Oh, nope, was a copy paste from the subclass that set the platformName;

titusfortner: https://github.com/appium/java-client/pull/1319/commits/499a385e6c067be52b2c8597fdd9d3bf49c547cf

titusfortner:
So overall this approach is good?

I'm ok to go through & add docstrings to `MobileOptions` and setters/getters/tests/docstrings for `AndroidMobileOptions` & `IOSMobileOptions` and the PR will be accepted?

mykola-mokhnach:
> So overall this approach is good?

Like I said I'm more into immutable data structures, but if Selenium is already doing it this way then there is no point for us not to follow it and provide different experience to users.

@SrinivasanTarget Please share your thoughts on the feature

SrinivasanTarget:
Approach looks fine to me in terms of giving same experience as of Selenium.

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,483 @@
...
+        return (String) getCapability(MobileCapabilityType.UDID);
+    }
+
+    @Override
+    public T merge(Capabilities extraCapabilities) {
+        super.merge(extraCapabilities);
+        return (T) this;
+    }
+
+    protected T amend(String optionName, Object value) {
```
mykola-mokhnach: is there a possibility to set a custom capability name as string?

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,483 @@
...
+     * @see MobileCapabilityType#ORIENTATION
+     */
+    public ScreenOrientation getOrientation() {
+        return (ScreenOrientation) getCapability(MobileCapabilityType.ORIENTATION);
+    }
+
+    /**
+     * Set the location of the app(s) to install before running a test.
+     *
+     * TODO: Consider supporting String as well as List since, but any String value can be put into a List of one
```
mykola-mokhnach: TODOs should be comments

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    /**
+     * Set the timeout for new commands.
+     *
+     * @param seconds is the number of seconds to timeout before seeing a new command.
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#NEW_COMMAND_TIMEOUT
+     */
+    public T setNewCommandTimeout(int seconds) {
```
mykola-mokhnach: Please use Duration type for for time units

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    /**
+     * Set the url for the location of the App.
+     *
+     * @param url is the URL representing the location of the App
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#APP
+     */
+    public T setApp(URL url) {
```
mykola-mokhnach: why only URL? this could also be a path to a local file

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: e6b795065e6d7086cc29b1027f0488c4ea1df6ef

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Well, this is a private method. Users can either use `setCapability()` or we can handle specific converting custom capabilities to String keys in the public setter.

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: removed

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: It supports String as well. https://github.com/appium/java-client/pull/1319/files#diff-66811b841580b850e3df6d3581ea3b83R72

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 59d0ef99e17548c614f221e648db196aadeea623

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    /**
+     * Set the url for the location of the App.
+     *
+     * @param url is a String representing the location of the App
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#APP
+     */
+    public T setApp(String url) {
```
mykola-mokhnach: I'd rather rename the argument to location

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,514 @@
...
+     * @see org.openqa.selenium.remote.CapabilityType#PLATFORM_NAME
+     */
+    public String getPlatformName() {
+        return (String) getCapability(CapabilityType.PLATFORM_NAME);
+    }
+
+    /**
+     * Set the url for the location of the App.
+     *
+     * @param url is a String representing the location of the App
```
mykola-mokhnach: I'd rather explicitly mention this could be either a full path to a file located on the `same` machine where the server is running or an URL to an application archive/package

src/main/java/io/appium/java_client/remote/MobileOptions.java
mykola-mokhnach: same below

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,514 @@
...
+     * @see MobileCapabilityType#CLEAR_SYSTEM_FILES
+     */
+    public T setClearSystemFiles() {
+        return setClearSystemFiles(true);
+    }
+
+    /**
+     * Set whether the app clears system files.
+     *
+     * @param bool is whether the app clears system files.
```
mykola-mokhnach: I think this cap requires better explanation on what `system files` means. These are leftovers from native test frameworks, like logs, debugging data, etc.

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,514 @@
...
+     *
+     * @return true if the app clears system files.
+     * @see MobileCapabilityType#CLEAR_SYSTEM_FILES
+     */
+    public boolean doesClearSystemFiles() {
+        return (boolean) getCapability(MobileCapabilityType.CLEAR_SYSTEM_FILES);
+    }
+
+    /**
+     * Set the name of the device.
```
mykola-mokhnach: this cap, actually, is only used on Sauce and it does not affect the actual `real` device, which is used for testing. One must use the `udid` one for this purpose.

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: We need to keep this tied to capability type. So to change this to location, we'd also need to change this: https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java#L61

mykola-mokhnach: nah, I mean simply `String url -> String location`

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Sounds good, updating to:

`String`-
`Set the absolute local path for the location of the App.` 
`URL`-
`Set the remote http URL for the location of the App.`

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: It isn't just for Sauce, but I'll update this to `Set the kind of mobile device or emulator to use`

titusfortner: 4ae42ed81cfa222c8067702c53db70d73056d730

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 10726da364ade09fbed223e3a6301459b2055c5a

titusfortner: For most of these, additional details can be found by following the `@see` link

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 4ae42ed81cfa222c8067702c53db70d73056d730

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: ohhh, I see, sounds good.

titusfortner: eecafb29ea553236e1bf8d627ae05a89f3c8f8d0

titusfortner:
I'm still trying to suss out the IOS code. Let me know if there are any issues with this one. I'll create a new PR for the IOS specific options. Thanks

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the port used to connect to the ADB server.
+     *
+     * @param port is the port to connect to the ADB server.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ADB_PORT
+     */
+    public AndroidMobileOptions setAdbPort(String port) {
```
mykola-mokhnach: port number should be an integer

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the Timeout in milliseconds used to wait for adb command execution.
+     *
+     * @param ms is the number of milliseconds to wait for adb command execution.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ADB_EXEC_TIMEOUT
+     */
+    public AndroidMobileOptions setAdbExecTimeout(Duration ms) {
```
mykola-mokhnach: Please change the arg name to `duration` and mention in docs that the minimum time resolution unit is one millisecond. Same for other caps, which use duration - they need to be time-unit agnostic, since the Duration entity already implements when we need.

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+        return amend(AndroidMobileCapabilityType.ADB_EXEC_TIMEOUT, ms);
```
mykola-mokhnach: this will be now `duration.toMillis()`

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the Timeout in milliseconds used to wait for adb command execution.
+     *
+     * @return Duration to wait for adb command execution.
+     * @see AndroidMobileCapabilityType#ADB_EXEC_TIMEOUT
+     */
+    public Duration getAdbExecTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.ADB_EXEC_TIMEOUT);
```
mykola-mokhnach: this will be now `return Duration.ofMillis(getCapability(AndroidMobileCapabilityType.ADB_EXEC_TIMEOUT))`

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the port used to connect to the ADB server.
+     *
+     * @return String of the port used to connect to the ADB server.
+     * @see AndroidMobileCapabilityType#ADB_PORT
+     */
+    public String getAdbPort() {
+        return (String) getCapability(AndroidMobileCapabilityType.ADB_PORT);
```
mykola-mokhnach: I think we should handle both numbers and strings here properly.

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the timeout to wait for a device to become ready after booting.
+     *
+     * @param seconds is the number of seconds to wait for a device to become ready after booting.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_DEVICE_READY_TIMEOUT
+     */
+    public AndroidMobileOptions setAndroidDeviceReadyTimeout(Duration seconds) {
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the timeout to wait for a device to become ready after booting.
+     *
+     * @return Duration to wait for a device to become ready after booting.
+     * @see AndroidMobileCapabilityType#ANDROID_DEVICE_READY_TIMEOUT
+     */
+    public Duration getAndroidDeviceReadyTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.ANDROID_DEVICE_READY_TIMEOUT);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the name of the directory on the device in which the apk will be pushed before install.
+     *
+     * @param path is the directory on the device in which the apk will be pushed before install.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_INSTALL_PATH
+     */
+    public AndroidMobileOptions setAndroidInstallPath(String path) {
```
mykola-mokhnach: Does it make sense to use java Path instead of raw String?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the timeout to wait for an apk to install to the device.
+     *
+     * @param ms is the number of milliseconds to wait for an apk to install to the device.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_INSTALL_TIMEOUT
+     */
+    public AndroidMobileOptions setAndroidInstallTimeout(Duration ms) {
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the timeout to wait for an apk to install to the device.
+     *
+     * @return Duration to wait for an apk to install to the device.
+     * @see AndroidMobileCapabilityType#ANDROID_INSTALL_TIMEOUT
+     */
+    public Duration getAndroidInstallTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.ANDROID_INSTALL_TIMEOUT);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the name of the directory on the device in which the screenshot will be put.
+     *
+     * @param path is the directory on the device in which the screenshot will be put.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_SCREENSHOT_PATH
+     */
+    public AndroidMobileOptions setAndroidScreenshotPath(String path) {
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the Timeout used to wait for the appWaitActivity to launch.
+     *
+     * @param ms is the number of milliseconds to wait for the appWaitActivity to launch.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#APP_WAIT_DURATION
+     */
+    public AndroidMobileOptions setAppWaitDuration(Duration ms) {
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the timeout used to wait for the appWaitActivity to launch.
+     *
+     * @return Duration to wait for the appWaitActivity to launch.
+     * @see AndroidMobileCapabilityType#APP_WAIT_DURATION
+     */
+    public Duration getAppWaitDuration() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.APP_WAIT_DURATION);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the amount of time to wait for Webview context to become active.
+     *
+     * @param ms is the number of milliseconds to wait for Webview context to become active.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#AUTO_WEBVIEW_TIMEOUT
+     */
+    public AndroidMobileOptions setAutoWebviewTimeout(Duration ms) {
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the amount of time to wait for Webview context to become active.
+     *
+     * @return Duration to wait for Webview context to become active.
+     * @see AndroidMobileCapabilityType#AUTO_WEBVIEW_TIMEOUT
+     */
+    public Duration getAutoWebviewTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.AUTO_WEBVIEW_TIMEOUT);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set the wait time for an avd to launch and connect to ADB.
+     *
+     * @param ms is the number of milliseconds to wait for an avd to launch and connect to ADB.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#AVD_LAUNCH_TIMEOUT
+     */
+    public AndroidMobileOptions setAvdLaunchTimeout(Duration ms) {
+        return amend(AndroidMobileCapabilityType.AVD_LAUNCH_TIMEOUT, ms);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the wait time for an avd to launch and connect to ADB.
+     *
+     * @return Duration to wait for an avd to launch and connect to ADB.
+     * @see AndroidMobileCapabilityType#AVD_LAUNCH_TIMEOUT
+     */
+    public Duration getAvdLaunchTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.AVD_LAUNCH_TIMEOUT);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set the wait time for an avd to finish its boot animations.
+     *
+     * @param ms is the number of milliseconds to wait for an avd to finish its boot animations.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#AVD_READY_TIMEOUT
+     */
+    public AndroidMobileOptions setAvdReadyTimeout(Duration ms) {
+        return amend(AndroidMobileCapabilityType.AVD_READY_TIMEOUT, ms);
```
mykola-mokhnach: same as above

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the the wait time for an avd to finish its boot animations.
+     *
+     * @return Duration to wait for an avd to finish its boot animations.
+     * @see AndroidMobileCapabilityType#AVD_READY_TIMEOUT
+     */
+    public Duration getAvdReadyTimeout() {
+        return (Duration) getCapability(AndroidMobileCapabilityType.AVD_READY_TIMEOUT);
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+        return amend(AndroidMobileCapabilityType.CHROMEDRIVER_ARGS, args);
+    }
+
+    /**
+     * Get the arguments to be passed to the chromedriver binary when it's run by Appium..
+     *
+     * @return list of String arguments to be passed to the chromedriver binary when it's run by Appium.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_ARGS
+     */
+    public Object getChromedriverArgs() {
```
mykola-mokhnach: why Object?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the file location which maps Chromedriver versions to the minimum Chrome version that it supports.
+     *
+     * @param path is the absolute path to a file mapping Chromedriver versions to Chrome versions.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_CHROME_MAPPING_FILE
+     */
+    public AndroidMobileOptions setChromedriverChromeMappingFile(String path) {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the path to webdriver executable.
+     *
+     * @param path is the absolute local path to webdriver executable.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE
+     */
+    public AndroidMobileOptions setChromedriverExecutable(String path) {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the directory to look for Chromedriver executables in, for automatic discovery of compatible Chromedrivers.
+     *
+     * @param dir is the absolute path of the directory to automatically discover compatible Chromedriver executable.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE_DIR
+     */
+    public AndroidMobileOptions setChromedriverExecutableDir(String dir) {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set the valid ports for Appium to use for communication with Chromedrivers.
+     *
+     * @param ports is the list of ports for Appium to use with Chromedrivers.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#BUILD_TOOLS_VERSION
+     */
+    public AndroidMobileOptions setChromedriverPorts(List<Integer> ports) {
+        return amend(AndroidMobileCapabilityType.BUILD_TOOLS_VERSION, ports);
```
mykola-mokhnach: BUILD_TOOLS_VERSION -> copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the valid ports for Appium to use for communication with Chromedrivers.
+     *
+     * @return list of Integers for the valid ports for Appium to use for communication with Chromedrivers.
+     * @see AndroidMobileCapabilityType#BUILD_TOOLS_VERSION
+     */
+    public Object getChromedriverPorts() {
+        return getCapability(AndroidMobileCapabilityType.BUILD_TOOLS_VERSION);
```
mykola-mokhnach: BUILD_TOOLS_VERSION?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the app to bypass automatic Chromedriver configuration and use the version that comes downloaded with Appium.
+     *
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_DISABLE_BUILD_CHECK
```
mykola-mokhnach: copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the chromeOptions capability for ChromeDriver.
+     *
+     * @return ChromeOptions set for use with ChromeDriver.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE_DIR
+     */
+    public Object getChromeOptions() {
+        return getCapability(AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE_DIR);
```
mykola-mokhnach: copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set the chromeOptions capability for ChromeDriver.
+     *
+     * @param opts is the chromeoptions for ChromeDriver.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE_DIR
+     */
+    public AndroidMobileOptions setChromeOptions(ChromeOptions opts) {
+        return amend(AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE_DIR, opts);
```
mykola-mokhnach: copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get whether to bypass automatic Chromedriver configuration and use the version that comes downloaded with Appium.
+     *
+     * @return true if using the version of Chromedriver that comes with Appium.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_DISABLE_BUILD_CHECK
+     */
+    public boolean doesChromedriverUseSystemExecutable() {
+        return (boolean) getCapability(AndroidMobileCapabilityType.CHROMEDRIVER_DISABLE_BUILD_CHECK);
```
mykola-mokhnach: copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set whether to bypass automatic Chromedriver configuration and use the version that comes downloaded with Appium.
+     *
+     * @param bool when true uses the version of Chromedriver that comes with Appium.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_DISABLE_BUILD_CHECK
+     */
+    public AndroidMobileOptions setChromedriverUseSystemExecutable(boolean bool) {
+        return amend(AndroidMobileCapabilityType.CHROMEDRIVER_DISABLE_BUILD_CHECK, bool);
```
mykola-mokhnach: copypaste?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the valid ports for Appium to use for communication with Chromedrivers.
+     *
+     * @param ports is the list of ports for Appium to use with Chromedrivers.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#BUILD_TOOLS_VERSION
+     */
+    public AndroidMobileOptions setChromedriverPorts(List<Integer> ports) {
```
mykola-mokhnach: is it really a list?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the path to custom keystore.
+     *
+     * @param path is the path to the custom keystore.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#KEYSTORE_PATH
+     */
+    public AndroidMobileOptions setKeystorePath(String path) {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the additional intent arguments that will be used to start activity.
+     *
+     * @param speed is the additional intent arguments that will be used to start activity.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#OPTIONAL_INTENT_ARGUMENTS
+     */
+    public AndroidMobileOptions setOptionalIntentArguments(String speed) {
```
mykola-mokhnach: speed?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the port used to connect to the driver.
+     *
+     * @param port is the port used to connect to the driver.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#SYSTEM_PORT
+     */
+    public AndroidMobileOptions setSystemPort(String port) {
```
mykola-mokhnach: perhaps int?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the port used to connect to the driver.
+     *
+     * @return String of the port used to connect to the driver.
+     * @see AndroidMobileCapabilityType#SYSTEM_PORT
+     */
+    public String getSystemPort() {
+        return (String) getCapability(AndroidMobileCapabilityType.SYSTEM_PORT);
```
mykola-mokhnach: int ?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the app to enable Unicode input.
+     *
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#UNICODE_KEYBOARD
+     */
+    public AndroidMobileOptions setUnicodeKeyboard() {
```
mykola-mokhnach: I would not include this cap at all. This only has effect for UiAutomator1 backend, which is considered obsolete

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+
+    /**
+     * Set the list of packages to uninstall before installing apks.
+     *
+     * @param packages is the list of packages to uninstall before installing apks.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#UNINSTALL_OTHER_PACKAGES
+     */
+    public AndroidMobileOptions setUninstallOtherPackages(List<String> packages) {
+        return amend(AndroidMobileCapabilityType.UNINSTALL_OTHER_PACKAGES, packages);
```
mykola-mokhnach: I assume this has to be a comma separated string: https://github.com/appium/appium-android-driver/blob/e2a7e4b5407b845ce7f21b8e4b0be0c2cc0fd01a/lib/desired-caps.js#L190

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the list of packages to uninstall before installing apks.
+     *
+     * @return list of Strings for packages to uninstall before installing apks.
+     * @see AndroidMobileCapabilityType#UNINSTALL_OTHER_PACKAGES
+     */
+    public Object getUninstallOtherPackages() {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Set the TCP port for communication with the webview.
+     *
+     * @param port is the port for communicating with the webview.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#WEBVIEW_DEVTOOLS_PORT
+     */
+    public AndroidMobileOptions setWebviewDevtoolsPort(String port) {
```
mykola-mokhnach: int ?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1716 @@
...
+    }
+
+    /**
+     * Get the TCP port for communication with the webview.
+     *
+     * @return String of the TCP port for communication with the webview.
+     * @see AndroidMobileCapabilityType#WEBVIEW_DEVTOOLS_PORT
+     */
+    public String getWebviewDevtoolsPort() {
+        return (String) getCapability(AndroidMobileCapabilityType.WEBVIEW_DEVTOOLS_PORT);
```
mykola-mokhnach: int ?

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,515 @@
...
+
+    /**
+     * Set the absolute local path for the location of the App.
+     * The  or remote http URL to a {@code .ipa} file (IOS),
+     *
+     * @param path is a String representing the location of the App
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#APP
+     */
+    public T setApp(String path) {
```
mykola-mokhnach: should it be of java's Path type?

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    /**
+     * Set the timeout for new commands.
+     *
+     * @param seconds is the number of seconds to timeout before seeing a new command.
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#NEW_COMMAND_TIMEOUT
+     */
+    public T setNewCommandTimeout(Duration seconds) {
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,515 @@
...
+    }
+
+    /**
+     * Get the timeout for new commands.
+     *
+     * @return timeout before seeing a new command in seconds.
+     * @see MobileCapabilityType#NEW_COMMAND_TIMEOUT
+     */
+    public Duration getNewCommandTimeout() {
+        return (Duration) getCapability(MobileCapabilityType.NEW_COMMAND_TIMEOUT);
```
mykola-mokhnach: ⬆️

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -0,0 +1,515 @@
...
+
+    /**
+     * Set the location of the app(s) to install before running a test.
+     *
+     * @param apps is the apps to install.
+     * @return this MobileOptions, for chaining.
+     * @see MobileCapabilityType#OTHER_APPS
+     */
+    public T setOtherApps(List<String> apps) {
+        return amend(MobileCapabilityType.OTHER_APPS, apps);
```
mykola-mokhnach: this should be a comma separated string

mykola-mokhnach:
What is the purpose of src/test/java/io/appium/java_client/IntentExample.apk ?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: Ah, but capabilities doesn't, I need to convert it. 
664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 664223287e7973d5c78c320bc331d084d4c91239

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: e38f7e1ca3c2cb82220c67dc5923a73fa418fcba

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: I'm not sure what you mean. If we set it as an Integer, it needs to be returned as an Integer.
Updated: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: It returns correctly as a List, and I can't cast the result into one. Is there a better way here? Java typing isn't my strong suit. :)

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
mykola-mokhnach: integer is fine 👍

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
mykola-mokhnach: if the `incoming` argument type is `List<String>` then I would expect this one to be the same

mykola-mokhnach: > It returns correctly as a List, and I can't cast the result into one

Why it is not possible to cast?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: Yes, [according to the docs](https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/caps.md#android-only) it is a list and not a comma separated string.
Though, this example would indicate that it isn't only a List of Integers:
> e.g. `[8000, [9000, 9005]]`
c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: Makes sense. Can we add that knowledge to the documentation and/or CapabilityType file?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: Oh, that's an interesting file that I didn't know about. :)
We should update the documentation then, since it lists this as an option:
`e.g. "io.appium.example", ["io.appium.example1", "io.appium.example2"], '*'`

titusfortner: https://github.com/appium/appium/pull/14133
1453a1a68bbe5b472568b94e8c4830430ce494d1

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: 1453a1a68bbe5b472568b94e8c4830430ce494d1

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: c5fbb3dc67a0891a002bbca235595fe13a8c3351

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: https://github.com/appium/appium-android-driver/blob/e2a7e4b5407b845ce7f21b8e4b0be0c2cc0fd01a/lib/desired-caps.js#L91

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: 1df2efd08b737ca1c942d244dc17ae2cd92b9d97

titusfortner:
> What is the purpose of src/test/java/io/appium/java_client/IntentExample.apk ?

I have no idea. I'm assuming that something I did when running the build/tests changed it. I can revert the change, but if tests are passing, I'm assuming it doesn't matter?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: ffd26c45f4e724b828427a9cbbefa33a806d3a6f

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
titusfortner: Well, IntelliJ was complaining before, and it isn't now. So I'll update :)

titusfortner:
All the tests are passing finally. Once I figure out what this is actually supposed to be: https://github.com/appium/appium/pull/14133 I'll do a rebase to tidy up the commits.

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+
+    /**
+     * Get the Timeout in milliseconds used to wait for adb command execution.
+     *
+     * @return Duration to wait for adb command execution.
+     * @see AndroidMobileCapabilityType#ADB_EXEC_TIMEOUT
+     */
+    public Duration getAdbExecTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.ADB_EXEC_TIMEOUT);
+        if (duration.getClass().isAssignableFrom(Long.class)) {
```
mykola-mokhnach, line 66: return Duration.ofMillis(Long.valueOf("" + duration));

mykola-mokhnach: same below.
Also, this could be extracted to a helper method to avoid unnecessary duplication

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get whether installation of a test package which has android:testOnly="true" in the manifest is allowed.
+     *
+     * @return true if able to install a test package marked testOnly.
+     * @see AndroidMobileCapabilityType#ALLOW_TEST_PACKAGES
+     */
+    public boolean doesAllowTestPackages() {
+        return (boolean) getCapability(AndroidMobileCapabilityType.ALLOW_TEST_PACKAGES);
```
mykola-mokhnach, line 123: Above we return `Integer` (object type) for a number value, but here we do `boolean` (primitive type). I assume we should follow the common pattern everywhere and return either all primitive (in such case NPE will be thrown if the corresponding value is null) or object types

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the timeout to wait for a device to become ready after booting.
+     *
+     * @param duration is the number of seconds to wait for a device to become ready after booting.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_DEVICE_READY_TIMEOUT
+     */
+    public AndroidMobileOptions setAndroidDeviceReadyTimeout(Duration duration) {
```
mykola-mokhnach, line 175: It looks like this cap is not used anywhere in appium code: https://github.com/appium/appium/search?q=androidDeviceReadyTimeout&type=Code

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the port used to connect to the ADB server.
+     *
+     * @param port is the port to connect to the ADB server.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ADB_PORT
+     */
+    public AndroidMobileOptions setAdbPort(Integer port) {
```
mykola-mokhnach, line 81: `int port`. It makes no sense to allow setting it to null

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the timeout to wait for a device to become ready after booting.
+     *
+     * @return Duration to wait for a device to become ready after booting.
+     * @see AndroidMobileCapabilityType#ANDROID_DEVICE_READY_TIMEOUT
+     */
+    public Duration getAndroidDeviceReadyTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.ANDROID_DEVICE_READY_TIMEOUT);
```
mykola-mokhnach, line 186: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the name of the directory on the device in which the apk will be pushed before install.
+     *
+     * @param path is the directory on the device in which the apk will be pushed before install.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_INSTALL_PATH
+     */
+    public AndroidMobileOptions setAndroidInstallPath(String path) {
```
mykola-mokhnach, line 223: this cap is not used anywhere in the code: https://github.com/appium/appium/search?q=androidInstallPath&unscoped_q=androidInstallPath&type=Code

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+        return Duration.ofMillis(value);
+    }
+
+    /**
+     * Set the app to allow for correct handling of orientation on landscape-oriented devices.
+     *
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_NATURAL_ORIENTATION
+     */
+    public AndroidMobileOptions setAndroidNaturalOrientation() {
```
mykola-mokhnach, line 270: this capability only works for the obsolete UIAutomator1 automation backend

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the name of the directory on the device in which the screenshot will be put.
+     *
+     * @param path is the directory on the device in which the screenshot will be put.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#ANDROID_SCREENSHOT_PATH
+     */
+    public AndroidMobileOptions setAndroidScreenshotPath(String path) {
```
mykola-mokhnach, line 302: this capability only works for the obsolete UIAutomator1 automation backend

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the activity name for the Android activity you want to launch from your package.
+     *
+     * @param activity is the name of the activity to launch from the package.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#APP_ACTIVITY
+     */
+    public AndroidMobileOptions setAppActivity(String activity) {
```
mykola-mokhnach, line 323: I think it makes sense to add a link to the documentation, because users quite often have difficulties understanding what activity name concept is: https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/android/activity-startup.md

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the Java package of the Android app you want to run.
+     *
+     * @param pkg is the Android app you want to run.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#APP_PACKAGE
+     */
+    public AndroidMobileOptions setAppPackage(String pkg) {
```
mykola-mokhnach, line 344: ⬆️ same as for activity name

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the activity name/names, comma separated, for the Android activity you want to wait for.
+     *
+     * @param activity is the comma separated activity names for activities to wait for.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#APP_WAIT_ACTIVITY
+     */
+    public AndroidMobileOptions setAppWaitActivity(String activity) {
```
mykola-mokhnach, line 365: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Set the Timeout used to wait for the appWaitActivity to launch.
+     *
+     * @param duration is the number of milliseconds to wait for the appWaitActivity to launch.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#APP_WAIT_DURATION
+     */
+    public AndroidMobileOptions setAppWaitDuration(Duration duration) {
```
mykola-mokhnach, line 386: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the timeout used to wait for the appWaitActivity to launch.
+     *
+     * @return Duration to wait for the appWaitActivity to launch.
+     * @see AndroidMobileCapabilityType#APP_WAIT_DURATION
+     */
+    public Duration getAppWaitDuration() {
+        Object duration = getCapability(AndroidMobileCapabilityType.APP_WAIT_DURATION);
```
mykola-mokhnach, line 397: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+     * @see AndroidMobileCapabilityType#APP_WAIT_ACTIVITY
+     */
+    public String getAppWaitActivity() {
+        return (String) getCapability(AndroidMobileCapabilityType.APP_WAIT_ACTIVITY);
+    }
+
+    /**
+     * Set the Timeout used to wait for the appWaitActivity to launch.
+     *
+     * @param duration is the number of milliseconds to wait for the appWaitActivity to launch.
```
mykola-mokhnach, line 382: should be unit-agnostic now

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the amount of time to wait for Webview context to become active.
+     *
+     * @return Duration to wait for Webview context to become active.
+     * @see AndroidMobileCapabilityType#AUTO_WEBVIEW_TIMEOUT
+     */
+    public Duration getAutoWebviewTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.AUTO_WEBVIEW_TIMEOUT);
```
mykola-mokhnach, line 507: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the wait time for an avd to launch and connect to ADB.
+     *
+     * @return Duration to wait for an avd to launch and connect to ADB.
+     * @see AndroidMobileCapabilityType#AVD_LAUNCH_TIMEOUT
+     */
+    public Duration getAvdLaunchTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.AVD_LAUNCH_TIMEOUT);
```
mykola-mokhnach, line 576: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+        }
+        Integer integer = (Integer) duration;
+        Long value = Long.valueOf(integer);
+        return Duration.ofMillis(value);
+    }
+
+    /**
+     * Set the wait time for an avd to finish its boot animations.
+     *
+     * @param duration is the number of milliseconds to wait for an avd to finish its boot animations.
```
mykola-mokhnach, line 588: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the the wait time for an avd to finish its boot animations.
+     *
+     * @return Duration to wait for an avd to finish its boot animations.
+     * @see AndroidMobileCapabilityType#AVD_READY_TIMEOUT
+     */
+    public Duration getAvdReadyTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.AVD_READY_TIMEOUT);
```
mykola-mokhnach, line 603: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+     * @param path is the absolute local path to webdriver executable.
+     * @return this MobileOptions, for chaining.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE
+     */
+    public AndroidMobileOptions setChromedriverExecutable(String path) {
+        return amend(AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE, path);
+    }
+
+    /**
+     * Get the path to webdriver executable.
```
mykola-mokhnach, line 718: I would also add the link to https://github.com/appium/appium/blob/master/docs/en/writing-running-appium/web/chromedriver.md

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+     * @return String of the path to webdriver executable.
+     * @see AndroidMobileCapabilityType#CHROMEDRIVER_EXECUTABLE
+     */
+    public String getChromedriverExecutable() {
+        return (String) getCapability(AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE);
+    }
+
+    /**
+     * Set the directory to look for Chromedriver executables in, for automatic discovery of compatible Chromedrivers.
+     *
```
mykola-mokhnach, line 729: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+        return amend(AndroidMobileCapabilityType.CHROME_OPTIONS, opts);
+    }
+
+    /**
+     * Get the chromeOptions capability for ChromeDriver.
+     *
+     * @return ChromeOptions set for use with ChromeDriver.
+     * @see AndroidMobileCapabilityType#CHROME_OPTIONS
+     */
+    public Object getChromeOptions() {
```
mykola-mokhnach, line 838: why Object?

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+        return getCapability(ChromeOptions.CAPABILITY);
+    }
+
+    /**
+     * Set the timeout for device to become ready.
+     *
+     * @param duration is the number of seconds to wait for the device to become ready.
```
mykola-mokhnach, line 845: ⬆️

src/main/java/io/appium/java_client/android/AndroidMobileOptions.java
```diff
@@ -0,0 +1,1733 @@
...
+    }
+
+    /**
+     * Get the timeout for device to become ready.
+     *
+     * @return Duration to wait for device to become ready.
+     * @see AndroidMobileCapabilityType#DEVICE_READY_TIMEOUT
+     */
+    public Duration getDeviceReadyTimeout() {
+        Object duration = getCapability(AndroidMobileCapabilityType.DEVICE_READY_TIMEOUT);
```
mykola-mokhnach, line 860: ⬆️

mykola-mokhnach:
@titusfortner I think it makes sense to split this PR into smaller parts, for example:
- basic classes implementation without concrete caps
- android caps
- ios caps
...

Then it would be much easier to review it

titusfortner:
@mykola-mokhnach  thanks for your painstaking reviews of this; closing to open a new PR.

PR #1324, opened 2020-03-29
change capabilities to use w3c compliant prefixes

titusfortner:
## Change list

* updated mobile capabilities to use w3c compliant prefix `appium:`
 
## Types of changes

## Details

This makes java-client compliant with required w3c syntax, which will likely be necessary when users update to Selenium 4.

The tests like `startingAndroidAppWithCapabilitiesOnlyTest` fail with this PR, but I would consider these more unit tests than user facing changes because no one should be checking the input to Capabilities with the output from the server.

That said, if we have #1319 we could override `getCapability` to add `appium:` if we need the inputs to match the outputs.

I also do not know if or how settings might need to be adjusted to ensure compatible versions of things are used.

### DIFF
--- a/src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
+++ b/src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
@@ -32,13 +32,13 @@ public interface AndroidMobileCapabilityType extends CapabilityType {
      * instead of {@code MainActivity}). By default this capability is received from the package
      * manifest (action: android.intent.action.MAIN , category: android.intent.category.LAUNCHER)
      */
-    String APP_ACTIVITY = "appActivity";
+    String APP_ACTIVITY = "appium:appActivity";
 
     /**
      * Java package of the Android app you want to run. By default this capability is received
      * from the package manifest ({@literal @}package attribute value)
      */
-    String APP_PACKAGE = "appPackage";
+    String APP_PACKAGE = "appium:appPackage";
 
     /**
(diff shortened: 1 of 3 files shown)


titusfortner:
(this is draft because I'm not sure how you want the tests handled, and if this should wait for #1319)

src/main/java/io/appium/java_client/remote/AndroidMobileCapabilityType.java
```diff
@@ -32,13 +32,13 @@
      * instead of {@code MainActivity}). By default this capability is received from the package
      * manifest (action: android.intent.action.MAIN , category: android.intent.category.LAUNCHER)
      */
-    String APP_ACTIVITY = "appActivity";
+    String APP_ACTIVITY = "appium:appActivity";
```
mykola-mokhnach: it makes sense to have a separate constant defined for `appium:`

mykola-mokhnach:
I think we cannot merge this change yet, because it will break mjsonwp tests. Actually, it's the same reason we don't incorporate the alpha version of selenium 4 lib - users want to have stable releases rather than alpha snapshots.

titusfortner:
Nevermind, you are automatically doing what is needed here:
https://github.com/appium/java-client/blob/50e231831c7a0d0a4bfb47a2f74242af6de8cff2/src/main/java/io/appium/java_client/remote/NewAppiumSessionPayload.java#L390-L392

and I just didn't missed it.

PR #1325, opened 2020-03-29
fix: remove appiumVersion from MobileCapabilityType

enriquegh:
## Change list

- Removed appiumVersion from MobileCapabilityType
 
## Types of changes

## Details

The capability `appiumVersion` is a Sauce specific capability.
This should not be included in the Java bindings since Appium does not do anything with this capability and should be handled either manually by the user or via some other third party tool.

This fixes this being included on the Appium Java client.

### DIFF
--- a/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
+++ b/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
@@ -65,10 +65,6 @@ public interface MobileCapabilityType extends CapabilityType {
      */
     String UDID = "udid";
 
-    /**
-     * Sauce-specific.
-     */
-    String APPIUM_VERSION = "appiumVersion";
 
     /**
      * Language to set for iOS (XCUITest driver only) and Android.


mykola-mokhnach:
@enriquegh Could you please sign the CLA?

enriquegh:
Done! @mykola-mokhnach

src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
```diff
@@ -65,10 +65,6 @@
      */
     String UDID = "udid";
 
-    /**
-     * Sauce-specific.
-     */
-    String APPIUM_VERSION = "appiumVersion";
```
mykola-mokhnach: I'd recommend marking the constant as `@Deprecated` first before we remove it completely

enriquegh: @mykola-mokhnach how about marking it as `@Deprecated` and add a filtering rule on `NewAppiumSessionPayload.java` to not include it (or any constants marked deprecated) when adding the `appium:` prefix?

I haven't found a way to do this but I can keep researching.

The goal here is to avoid the Java client sending `appium:appiumVersion` across the wire.

mykola-mokhnach: > The goal here is to avoid the Java client sending appium:appiumVersion across the wire.
 Write

One cannot simply deny using it :)
put Deprecated annotation and add an explanation in the docstring on why this cap should not be used.

enriquegh: @mykola-mokhnach 
that wouldn't be enough unfortunately.

So the MobileCapabilityType is used for two things.
1. The obvious one of using it to aid you when setting a capability. So something like:
```java
caps.setCapability(MobileCapabilityType.APPIUM_VERSION, "1.17.0")
```
For that, this can be deprecated and it will show up on the user's IDE nicely.
2. In `NewAppiumSessionPayload` we grab all "Appium" commands and add them to a list:
https://github.com/appium/java-client/blob/b914237a58054593612031a568bc5fc8e71cb79f/src/main/java/io/appium/java_client/remote/NewAppiumSessionPayload.java#L87-L91

we then go through the capabilities the user is trying to send and if any matches one of our capabilities, we add `appium:` to it:
https://github.com/appium/java-client/blob/b914237a58054593612031a568bc5fc8e71cb79f/src/main/java/io/appium/java_client/remote/NewAppiumSessionPayload.java#L390-L391

This means that it doesn't matter if I do
```
caps.setCapability(MobileCapabilityType.APPIUM_VERSION, "1.17.0")
```
or
```
caps.setCapability("appiumVersion", "1.17.0")
```
the client will send it as `appium:appiumVersion` no matter what.

This makes it an issue since that is technically not correct. The Appium server doesn't recognize it as a proper cap and Sauce can't read it in since it has a `appium:` prefix.

What I am trying to solve is usage 2.
I want `appiumVersion` not to be included in the addition of the appium prefix.

enriquegh:
This is not quite what I wanted to do...
It added a previous commit in my PR.

I will try to fix it

PR #1332, opened 2020-04-03
chore: Change getDeviceTime to call the `mobile` implementation

mykola-mokhnach:
## Types of changes

## Details

Selenium webdriver implementation ignores parameters passed into GET requests. This PR replaces the call with parameters with the corresponding `mobile:` endpoint implementation to make it working properly. See https://github.com/appium/appium/issues/14141


markbearden:
Note that the main Appium documentation for the system_time endpoint, and the Appium-hosted copy of the Java client's getDeviceTime(), do not yet indicate any support for the "format" parameter that this change is adding.  (I am wondering from where came the expectation that the client could ask the Appium server to format the time response, when that appears to have never been supported/documented for the Appium server?)  It is not explicitly indicated either here, or at appium/appium#14141, what specification is being complied by making this change.

PR #1333, opened 2020-04-07
refactor: Put test files into test resources

mykola-mokhnach:
## Change list

Previously test resources were stored together with test sources. This is not the way it should be done.
 
## Types of changes


src/test/java/io/appium/java_client/TestResources.java
```diff
@@ -0,0 +1,31 @@
+package io.appium.java_client;
+
+import java.nio.file.Path;
+
+import static io.appium.java_client.TestUtils.resourcePathToLocalPath;
+
+public class TestResources {
```
SrinivasanTarget, line 7: 👍

Issue #1336, opened 2020-04-21
Wrong JSON for chromeOptions


parunruaivo:
## Description

When trying to disable the w3c using the following lines

```
val chromeOptions = new ChromeOptions();
chromeOptions.setExperimentalOption("w3c", false);
desiredCapabilities.setCapability(ChromeOptions.CAPABILITY, chromeOptions);
```

When parsing the capabilities in the class `AppiumServiceBuilder` the chromeOptions is parsed with  `String.valueOf(value)` in the method `parseCapabilitiesIfWindows`, resulting in `\"goog:chromeOptions\": {args=[], extensions=[], w3c=false}`, then when executing the command line in the method `executeAsync() ` of the class `OsProcess ` the `org.apache.commons.exec.CommandLine cl;` is as below

```
[C:\Program Files\nodejs\node.exe, C:\Program Files\nodejs\node_modules\appium\build\lib\main.js, --port, 39775, --address, 127.0.0.1, --chromedriver-executable, D:\TESTRIGOR\Vagrant\temp\81\chromedriver.exe, --default-capabilities, {\"acceptSslCerts\": true, \"autoGrantPermissions\": true, \"automationName\": \"UIAutomator2\", \"avd\": \"android-26-pixel-xl\", \"avdArgs\": \"-no-audio -writable-system -ports 0,5563\", \"avdLaunchTimeout\": 180000, \"avdReadyTimeout\": 180000, \"browserName\": \"chrome\", \"clearSystemFiles\": true, \"deviceName\": \"android-26-pixel-xl\", \"disableAndroidWatchers\": true, \"goog:chromeOptions\": {args=[], extensions=[], w3c=false}, \"nativeWebScreenshot\": true, \"newCommandTimeout\": 900, \"platformName\": \"Android\", \"platformVersion\": \"8.0.0\", \"uiautomator2ServerInstallTimeout\": 120000}]
```
Having a wrong JSON because of the `args=[], extensions=[], w3c=false`

## Workaround
I created the class 
```
@FieldDefaults(level = PRIVATE, makeFinal = true)
@RequiredArgsConstructor
public class W3cOption {
	boolean value;

	public Map<String, Object> asMap() {
		return ImmutableMap.of("w3c", value);
	}

	@Override
	public String toString() {
		if (Platform.getCurrent().is(Platform.WINDOWS)) {
			return "{\\\"w3c\\\": " + value + "}";
		}
		return "{\"w3c\": " + value + "}";
	}
}
```

[...]

PR #1337, opened 2020-04-22
fix: Properly translate desiredCapabilities into a command line argument

mykola-mokhnach:
## Types of changes

## Details

Addresses https://github.com/appium/java-client/issues/1336

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
@@ -23,8 +23,8 @@ import static org.openqa.selenium.remote.CapabilityType.PLATFORM_NAME;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
-import io.appium.java_client.remote.AndroidMobileCapabilityType;
-import io.appium.java_client.remote.MobileCapabilityType;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 import io.appium.java_client.service.local.flags.ServerArgument;
 
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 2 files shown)


mykola-mokhnach:
@SrinivasanTarget Could you please rerun the E2E tests? I've just patched the UIA2 driver

SrinivasanTarget:
> @SrinivasanTarget Could you please rerun the E2E tests? I've just patched the UIA2 driver

Retriggered now

root-intruder:
@mykola-mokhnach I also recently stumbled across this. With your fix, it seems that no arguments are escaped with `\"` which means that it always has to be manually escaped for every capability.

I see why a fix was needed, but that makes the situation worse for all windows users using mostly common simple capabilities such as "app" or "udid".

Is there any mechanism to work against this problem, without now quoting each arguments I pass to the capabilities.

Thanks!

root-intruder:
@mykola-mokhnach thanks for the answer! How about overloading `public AppiumServiceBuilder withCapabilities(DesiredCapabilities capabilities)"` as `public AppiumServiceBuilder withCapabilities(DesiredCapabilities capabilities, boolean quoteAllCapabilities)`

and then quote everything in `private String capabilitiesToCmdlineArg()` https://github.com/appium/java-client/blob/bcca20276191f98d9293974b6430cb97cb5e465c/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java#L299 if quoteAllCapabilities is set?!

Then still the user could decide if he want's to quote everything. 

Alternatively it would also be possible to make quoting everything the default on windows systems and only disable it if the user sets the flag. This way, most win users wouldn't have to escape by hand and special cases like #1336 can still be addressed.

Cheers!

mykola-mokhnach:
The overload proposal seems reasonable to me. Feel free to create a pull request

root-intruder:
@mykola-mokhnach would you be fine with the described second option: quoting everything on win systems and the user needs to opt out and then manually quote in cases as described in #1336?

I'd do a pull request in the next days!

mykola-mokhnach:
> would you be fine with the described second option: quoting everything on win systems and the user needs to opt out and then manually quote in cases as described in #1336?

I would rather stick to the first option, since changing the default behaviour is a breaking change. Clients must explicitly opt in for "automatic" quoting if they want to

root-intruder:
you're absolutely right...didn't think of existing users

PR #1354, opened 2020-05-27
fix: Increase the timeout for graceful service termination

mykola-mokhnach:
## Change list

Selenium lib hardcodes the graceful shutdown timeout to 2 seconds (zero on Windows). This is not enough for Appium to properly shutdown all the drivers and clean up the leftovers.

 
## Types of changes


src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -187,10 +191,47 @@ public void stop() {
...
+     * @return The exit code of the process or zero if the process was not running.
+     */
+    private int destroyProcess(Duration timeout) {
+        if (!process.isRunning()) {
+            return 0;
         }
+
+        // This all magic is necessary, because Selenium does not publicly expose
+        // process killing timeouts. By default a process is killed forcibly if
+        // it does not exit after two seconds, which is in most cases not enough for
```
SrinivasanTarget, line 207: >a process is killed forcibly if it does not exist after two seconds

Interesting.

PR #1356, opened 2020-05-30
As map

titusfortner:
## Change list
* Added an asMap method for MobileOptions class to override the one in MutableCapabilities

## Types of changes

## Details

The new `MobileOptions` classes can force w3c compliance without concerns of backward compatibility since they have not yet been released. Current `asMap()` implementation is not w3c compliant because it needs to prepend `appium:`. Essentially I'm looking to be able to merge together Capabilities, the same way [Selenium does for browsers](https://github.com/SeleniumHQ/selenium/blob/master/java/client/src/org/openqa/selenium/firefox/FirefoxOptions.java#L279)

The way I'm reading `getW3C()` in `NewAppiumSessionPayload` it should still do the correct thing for all inputs. The only place that seems to get a capability without using the helper is for `FORCE_MJSONWP` which shouldn't apply to the new options classes, so I don't think it should be a concern.

Note: I can't figure out how to set up my machine for running the entire test suite locally and I can't find any documentation on it, so I can only confirm that the new mobile options classes are working correctly. Any assistance on this would be appreciated.


src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -512,4 +518,20 @@ protected T amend(String optionName, Object value) {
...
         return (T) this;
     }
+
+    @Override
+    public Map<String, Object> asMap() {
+        Map<String, Object> toReturn = new TreeMap<>();
+
+        Map<String, Object> capsMap = super.asMap();
+        capsMap.forEach((key, value) -> {
+            if (key.equals("platformName")) {
```
mykola-mokhnach, line 528: I think it is not only the platform name. W3C defines a bunch of keys, which must be supplied without prefixes and Selenium has this list somewhere in constants

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -512,4 +518,20 @@ protected T amend(String optionName, Object value) {
...
+                toReturn.put(key, value);
+            } else {
+                toReturn.put(APPIUM_PREFIX + key, value);
```
mykola-mokhnach, line 531: should we also check if cap name is already prefixed?

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -512,4 +518,20 @@ protected T amend(String optionName, Object value) {
         setCapability(optionName, value);
         return (T) this;
     }
+
+    @Override
+    public Map<String, Object> asMap() {
+        Map<String, Object> toReturn = new TreeMap<>();
+
+        Map<String, Object> capsMap = super.asMap();
+        capsMap.forEach((key, value) -> {
```
mykola-mokhnach, line 527: I think we can build toReturn without having the intermediate `capsMap` variable, e.g.

```
Map<String, Object> toReturn = super.asMap().entrySet().stream().map ....
```

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Hmm; `platformName` is the only one we specify in `MobileOptions`, but you are probably right here. Looks like `NewAppiumSessionPayload` uses `AcceptedW3CCapabilityKeys` which is probably the right fix?

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: I pretty much just copied it from Selenium code, which I think is what happens when you use Selenium 4 with current Appium code? And Selenium browser options I think did more things with the intermediate step, so maybe it can be replaced with your suggestion here, I can try.

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: Yeah, I guess if we are allowing people to do `setCapability()` we should have some protection for that.

src/main/java/io/appium/java_client/remote/MobileOptions.java
titusfortner: So, I can't figure out how to get the stream to work. This is what I'm working with right now, which removes the redundant `capsMap`, but still uses `toReturn` and `forEach` directly:

```java
        super.asMap().forEach((key, value) -> {
            AcceptedW3CCapabilityKeys acceptedW3CCapabilityKeys = new AcceptedW3CCapabilityKeys();
            if (acceptedW3CCapabilityKeys.test(key) || key.startsWith(APPIUM_PREFIX)) {
                toReturn.put(key, value);
            } else {
                toReturn.put(APPIUM_PREFIX + key, value);
            }
        });
```

titusfortner: Ok, got a stream figured out. Let me know what you think.

src/main/java/io/appium/java_client/remote/MobileOptions.java
```diff
@@ -512,4 +518,14 @@ protected T amend(String optionName, Object value) {
         setCapability(optionName, value);
         return (T) this;
     }
+
+    @Override
+    public Map<String, Object> asMap() {
+        AcceptedW3CCapabilityKeys acceptedW3CCapabilityKeys = new AcceptedW3CCapabilityKeys();
+
+        return unmodifiableMap(super.asMap().entrySet().stream().collect(Collectors.toMap(
+                entry -> (acceptedW3CCapabilityKeys.test(entry.getKey()) || entry.getKey().startsWith(APPIUM_PREFIX)) ?
```
mykola-mokhnach, line 527: I would move this handler into a separate method for better readability. Also, in Appium server code we don't add prefix if a capability name already contains `:` character, not just the `appium:` prefix. Should we consider to do the same here?

PR #1370, opened 2020-07-08
fix: Automatically retrieve the host and port in ListensToLogcatMessages

wulfihm:
## Change list

Related to #1352
 
## Types of changes

[ ] No changes in production code.
[x] Bugfix (non-breaking change which fixes an issue)
[ ] New feature (non-breaking change which adds functionality)
[ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details

Automatically retrieve the host and port instead of assuming they are the defaults.

### DIFF
--- a/src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
+++ b/src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
@@ -33,13 +33,13 @@ import java.util.function.Consumer;
 public interface ListensToLogcatMessages extends ExecutesMethod {
     StringWebSocketClient getLogcatClient();
 
+
     /**
      * Start logcat messages broadcast via web socket.
-     * This method assumes that Appium server is running on localhost and
-     * is assigned to the default port (4723).
      */
     default void startLogcatBroadcast() {
-        startLogcatBroadcast("localhost", DEFAULT_APPIUM_PORT);
+        URL remoteAdress = ((RemoteWebDriver) this).getRemoteAddress();
+        this.startLogcatBroadcast(remoteAdress.getHost(), remoteAdress.getPort());
     }
 
     /**
--- a/src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
+++ b/src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
@@ -21,11 +21,13 @@ import static org.openqa.selenium.remote.DriverCommand.EXECUTE_SCRIPT;
 
 import com.google.common.collect.ImmutableMap;
 
+import io.appium.java_client.AppiumDriver;
 import io.appium.java_client.ExecutesMethod;
 import io.appium.java_client.ws.StringWebSocketClient;
 import org.openqa.selenium.remote.RemoteWebDriver;
 
 import java.net.URI;
+import java.net.URL;
 import java.net.URISyntaxException;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
```diff
@@ -33,13 +33,13 @@
...
 
+
     /**
      * Start logcat messages broadcast via web socket.
-     * This method assumes that Appium server is running on localhost and
-     * is assigned to the default port (4723).
      */
     default void startLogcatBroadcast() {
-        startLogcatBroadcast("localhost", DEFAULT_APPIUM_PORT);
+        URL remoteAdress = ((RemoteWebDriver) this).getRemoteAddress();
```
mykola-mokhnach, line 41: This might break the backward compatibility consider created another overriden method there, which allows to set the host and port in arguments

src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
```diff
@@ -33,13 +33,13 @@
 public interface ListensToLogcatMessages extends ExecutesMethod {
     StringWebSocketClient getLogcatClient();
 
+
```
mykola-mokhnach, line 36: no need for an extra empty line here

src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
wulfihm: I have trouble understanding what you mean. There is still the `default void startLogcatBroadcast(String host, int port)` if you want to set it manually.

mykola-mokhnach: yes, I see now. Then such change would be a breaking one

src/main/java/io/appium/java_client/android/ListensToLogcatMessages.java
```diff
@@ -35,11 +37,10 @@
 
     /**
      * Start logcat messages broadcast via web socket.
-     * This method assumes that Appium server is running on localhost and
-     * is assigned to the default port (4723).
      */
     default void startLogcatBroadcast() {
-        startLogcatBroadcast("localhost", DEFAULT_APPIUM_PORT);
+        URL remoteAddress = ((AppiumDriver) this).getRemoteAddress();
+        this.startLogcatBroadcast(remoteAddress.getHost(), remoteAddress.getPort());
```
mykola-mokhnach, line 43: This won't work if the client is connected to a hub. The hub cannot proxy/forward web socket connections

wulfihm: A hub?
I looked through the code and the only class that implements the `ListensToLogcatMessages` is `AndroidDriver<T extends WebElement>`. This class on the other hand extends `AppiumDriver`. Every constructor of the AppiumDriver class writes its address to the `remoteAdress` variable, which `getRemoteAddress` returns.
So I don't really see the possibilitiy of it not working. Am I missing something?

mykola-mokhnach: This endpoint is special from the perspective it uses web socket communication rather than http ones. `remoteAddress` may be pointing to the address of a selenium hub, where no web sockets are running.

PR #1454, opened 2021-03-23
fix: bring back ability to automatically quote desired capabilities on win

root-intruder:
## Change list
- Bring back automatic quote escaping for desired capabilities command line arguments on windows
 
## Types of changes

## Details
Escaped quotes are necessary for starting appium from the command line on windows systems (see http://appium.io/docs/en/writing-running-appium/default-capabilities-arg/). 

This capability to automatically quote on windows systems was completely removed with commit #1337. Since #1336 discovered a special case where automatic quote escaping on windows was not possible. That means, that windows users now need to manually escape the JSON such as:

```
DesiredCapabilities cap = new DesiredCapabilities();
cap.setCapability("\"platformName\"", "\"Android\"");

AppiumServiceBuilder builder = new AppiumServiceBuilder();
builder.withCapabilities(cap);
```
For a normal user this additional escape `"\"platformName\""` is cumbersome and not intuitive at all. It also makes it hard to write java code running on windows and unix/linux systems. 

From my understanding, the escaping should be done automatically by the java-client. However, this is a breaking change, thus I propose a switch to be able to enable quote escaping. If desired, I could also implement quote escaping as default behaviour with a switch to disable it. But I think this decision should be on your side.

Cheers!

### DIFF
--- a/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
@@ -25,6 +25,8 @@ import com.google.common.collect.ImmutableMap;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import io.appium.java_client.remote.AndroidMobileCapabilityType;
+import io.appium.java_client.remote.MobileCapabilityType;
 import io.appium.java_client.service.local.flags.ServerArgument;
 
 import org.apache.commons.io.IOUtils;
@@ -32,6 +34,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.SystemUtils;
 import org.apache.commons.validator.routines.InetAddressValidator;
 import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.Platform;
 import org.openqa.selenium.os.ExecutableFinder;
 import org.openqa.selenium.remote.BrowserType;
 import org.openqa.selenium.remote.DesiredCapabilities;
@@ -78,6 +81,7 @@ public final class AppiumServiceBuilder
     private File node;
     private String ipAddress = BROADCAST_IP_ADDRESS;
     private DesiredCapabilities capabilities;
+    private boolean autoQuoteCapabilitiesOnWindows = false;
     private static final Function<File, String> APPIUM_JS_NOT_EXIST_ERROR = (fullPath) -> String.format(
             "The main Appium script does not exist at '%s'", fullPath.getAbsolutePath());
     private static final Function<File, String> NODE_JS_NOT_EXIST_ERROR = (fullPath) ->
@@ -86,6 +90,8 @@ public final class AppiumServiceBuilder
     // The first starting is slow sometimes on some environment
     private long startupTimeout = 120;
     private TimeUnit timeUnit = TimeUnit.SECONDS;
+    private static final List<String> PATH_CAPABILITIES = ImmutableList.of(AndroidMobileCapabilityType.KEYSTORE_PATH,
+            AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE, MobileCapabilityType.APP);
 
     public AppiumServiceBuilder() {
         usingPort(DEFAULT_APPIUM_PORT);
@@ -240,6 +246,21 @@ public final class AppiumServiceBuilder
     }
 
     /**
+     * Adds a desired capabilities.
+     *
+     * @param capabilities is an instance of {@link DesiredCapabilities}.
+     * @param autoQuoteCapabilitiesOnWindows automatically escape quote all
+     *                                       capabilities when calling appium.
+     *                                       This is required on windows systems only.
+     * @return the self-reference.
+     */
+    public AppiumServiceBuilder withCapabilities(DesiredCapabilities capabilities,
+                                                 boolean autoQuoteCapabilitiesOnWindows) {
+        this.autoQuoteCapabilitiesOnWindows = autoQuoteCapabilitiesOnWindows;
+        return withCapabilities(capabilities);
+    }
+
+    /**
      * Sets an executable appium.js.
      *
      * @param appiumJS an executable appium.js (1.4.x and lower) or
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,48 @@ private void loadPathToMainScript() {
...
+            Set<? extends Map.Entry<String, ?>> entries = capabilitiesMap.entrySet();
+
+            for (Map.Entry<String, ?> entry : entries) {
+                Object value = entry.getValue();
+
+                if (value == null) {
+                    continue;
+                }
+
+                if (String.class.isAssignableFrom(value.getClass())) {
```
mykola-mokhnach, line 334: value instanceof String should be enough

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,48 @@ private void loadPathToMainScript() {
...
+                    if (pathCapabilities.contains(entry.getKey())) {
+                        value = "\\\"" + String.valueOf(value).replace("\\", "/") + "\\\"";
```
mykola-mokhnach, line 336: we could safely case value to String inside of this block

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,48 @@ private void loadPathToMainScript() {
...
+                    } else {
+                        value = "\\\"" + value + "\\\"";
+                    }
+                } else {
+                    value = String.valueOf(value);
+                }
+
+                String key = "\\\"" + entry.getKey() + "\\\"";
+                if (StringUtils.isBlank(result)) {
+                    result = key + ": " + value;
```
mykola-mokhnach, line 346: I usually prefer a ternary operator

mykola-mokhnach: same above

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: This was the code before commit #1337 so I decided to simply copy and paste this function to not break anyting.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: This was the code before commit #1337 so I decided to simply copy and paste this function to not break anyting.

root-intruder: Should I still implement your suggestions?

root-intruder: considering ternary operator, I disagree...usually makes code less readable.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: resolved in newest push

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: resolved in the newest push

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,49 @@ private void loadPathToMainScript() {
         this.appiumJS = findMainScript();
     }
 
+    private String parseCapabilitiesIfWindows() {
+        String result = StringUtils.EMPTY;
```
mykola-mokhnach, line 319: Please replace it with StringBuilder

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
mykola-mokhnach: I don't have a strong opinion on that, just a matter of taste

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: done in latest push

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,50 @@ private void loadPathToMainScript() {
...
+                    }
+                } else {
+                    value = String.valueOf(value);
+                }
+
+                String key = "\\\"" + entry.getKey() + "\\\"";
+                if (result.length() > 0) {
+                    result.append(", ");
+                }
+                result.append(key);
```
mykola-mokhnach, line 349: this method returns self instance, so all calls could be changed like result.append().append()....

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,50 @@ private void loadPathToMainScript() {
...
+                        value = "\\\"" + valueString.replace("\\", "/") + "\\\"";
+                    } else {
+                        value = "\\\"" + valueString + "\\\"";
+                    }
+                } else {
+                    value = String.valueOf(value);
+                }
+
+                String key = "\\\"" + entry.getKey() + "\\\"";
+                if (result.length() > 0) {
```
mykola-mokhnach, line 346: !result.isEmpty() could be more readable

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: you're right looks cleaner...pushed

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,48 @@ private void loadPathToMainScript() {
         this.appiumJS = findMainScript();
     }
 
+    private String parseCapabilitiesIfWindows() {
+        StringBuilder result = new StringBuilder();
+        final List<String> pathCapabilities = ImmutableList.of(AndroidMobileCapabilityType.KEYSTORE_PATH,
+                AndroidMobileCapabilityType.CHROMEDRIVER_EXECUTABLE, MobileCapabilityType.APP);
+
+        if (capabilities != null) {
```
mykola-mokhnach, line 323: this guard could be simplified to 
```
if (capabilities == null) {
  return "{}";
}
```

and moved to the beginning of the method

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +315,48 @@ private void loadPathToMainScript() {
         this.appiumJS = findMainScript();
     }
 
+    private String parseCapabilitiesIfWindows() {
+        StringBuilder result = new StringBuilder();
+        final List<String> pathCapabilities = ImmutableList.of(AndroidMobileCapabilityType.KEYSTORE_PATH,
```
mykola-mokhnach, line 320: it makes sense to move this list to class-level constants

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: I see no real advantage in your suggestion. The guard `if (capabilities == null) {` skips all methods code and directly jumps to the return at the end. Expressing it in the original reduces code duplication by not writing the return twice.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
root-intruder: done.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
mykola-mokhnach: this improves readability. Longer code blocks are hard to scan through

root-intruder: done

mykola-mokhnach (approved):
Thanks, I like it

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -296,7 +317,46 @@ private void loadPathToMainScript() {
         this.appiumJS = findMainScript();
     }
 
+    private String parseCapabilitiesIfWindows() {
```
mykola-mokhnach, line 320: This method name does not do any check for the current platform, so I find it confusing it has `ifWindows` in its name

root-intruder: rename to ...for Windows?!

mykola-mokhnach: how about `capabilitiesToQuotedCmdlineArg` ?

root-intruder: sounds good aswell...changed

PR #1506, opened 2021-08-21
feat: allow to add custom command dynamically

SrinivasanTarget:
## Change list

Appium 2.0 will allow custom plugins to add their own endpoint. This feature helps add custom commands dynamically. 
 
## Types of changes

## Details

```
driver.addCommand(HttpMethod.GET, "/sessions", "getSessions");
driver.execute("getSessions");
```

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -281,7 +281,6 @@ public class AppiumDriver<T extends WebElement>
         } else if (method == HttpMethod.DELETE) {
             MobileCommand.commandRepository.put(name, MobileCommand.deleteC(url));
         }
-        new AppiumCommandExecutor(MobileCommand.commandRepository, remoteAddress).defineCommand(name);
     }
 
     @Override
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -50,6 +50,7 @@ import org.openqa.selenium.remote.http.HttpClient;
 import org.openqa.selenium.remote.http.HttpMethod;
 
 import java.net.URL;
+import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -306,7 +307,9 @@ public class AppiumDriver<T extends WebElement>
                 MobileCommand.commandRepository.put(methodName, MobileCommand.deleteC(url));
                 break;
             default:
-                throw new WebDriverException("Unsupported HTTP Method: " + httpMethod);
+                throw new WebDriverException(String.format("Unsupported HTTP Method: %s. Only %s methods are supported",
+                        httpMethod,
+                        Arrays.toString(HttpMethod.values())));
         }
     }
 
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -290,17 +290,23 @@ public class AppiumDriver<T extends WebElement>
     /**
      * This method is used to add custom appium commands in Appium 2.0.
      *
-     * @param method the available {@link HttpMethod}.
+     * @param httpMethod the available {@link HttpMethod}.
      * @param url The url to URL template as https://www.w3.org/TR/webdriver/#endpoints.
-     * @param name The name of custom appium command.
+     * @param methodName The name of custom appium command.
      */
-    public void addCommand(HttpMethod method, String url, String name) {
-        if (method == HttpMethod.POST) {
-            MobileCommand.commandRepository.put(name, MobileCommand.postC(url));
-        } else if (method == HttpMethod.GET) {
-            MobileCommand.commandRepository.put(name, MobileCommand.getC(url));
-        } else if (method == HttpMethod.DELETE) {
-            MobileCommand.commandRepository.put(name, MobileCommand.deleteC(url));
+    public void addCommand(HttpMethod httpMethod, String url, String methodName) {
+        switch (httpMethod) {
+            case GET:
+                MobileCommand.commandRepository.put(methodName, MobileCommand.getC(url));
+                break;
+            case POST:
+                MobileCommand.commandRepository.put(methodName, MobileCommand.postC(url));
+                break;
+            case DELETE:
+                MobileCommand.commandRepository.put(methodName, MobileCommand.deleteC(url));
+                break;
+            default:
+                throw new WebDriverException("Unsupported HTTP Method: " + httpMethod);
         }
     }
 
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -16,45 +16,30 @@
 
 package io.appium.java_client;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static io.appium.java_client.remote.MobileCapabilityType.PLATFORM_NAME;
-import static org.apache.commons.lang3.StringUtils.containsIgnoreCase;
-import static org.apache.commons.lang3.StringUtils.isBlank;
-
 import com.google.common.collect.ImmutableMap;
-
 import io.appium.java_client.internal.CapabilityHelpers;
 import io.appium.java_client.internal.JsonToMobileElementConverter;
 import io.appium.java_client.remote.AppiumCommandExecutor;
 import io.appium.java_client.remote.MobileCapabilityType;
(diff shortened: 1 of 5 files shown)


src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -34,13 +35,22 @@
...
 import org.openqa.selenium.support.ui.ExpectedConditions;
 import org.openqa.selenium.support.ui.WebDriverWait;
 
 import java.time.Duration;
 
 public class IOSDriverTest extends AppIOSTest {
 
+    @Test
+    public void addCustomCommandTest() {
+        driver.addCommand(HttpMethod.GET, "/sessions", "getSessions");
```
KazuCocoa, line 49: Do you have examples of placeholders?
e.g. session id, element id etc
https://github.com/appium/ruby_lib_core/blob/0f93a52bbdc44bf916c9b974fe9fd09d48e5ff39/lib/appium_lib_core/common/base/driver.rb#L107-L145
https://github.com/appium/python-client/blob/e48085d27968a93d9cfb5d86964ac84198b52214/test/unit/webdriver/webdriver_test.py#L285-L309

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -288,6 +287,23 @@ public void rotate(ScreenOrientation orientation) {
...
+     * @param name The name of custom appium command.
+     */
+    public void addCommand(HttpMethod method, String url, String name) {
+        if (method == HttpMethod.POST) {
+            MobileCommand.commandRepository.put(name, MobileCommand.postC(url));
+        } else if (method == HttpMethod.GET) {
+            MobileCommand.commandRepository.put(name, MobileCommand.getC(url));
+        } else if (method == HttpMethod.DELETE) {
+            MobileCommand.commandRepository.put(name, MobileCommand.deleteC(url));
+        }
```
KazuCocoa, line 304: Probably it is helpful to print log or raises an error if no matched case

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
SrinivasanTarget: Sure @KazuCocoa will add it.

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -288,6 +287,23 @@ public void rotate(ScreenOrientation orientation) {
...
 
+    /**
+     * This method is used to add custom appium commands in Appium 2.0.
+     *
+     * @param method the available {@link HttpMethod}.
+     * @param url The url to URL template as https://www.w3.org/TR/webdriver/#endpoints.
+     * @param name The name of custom appium command.
+     */
+    public void addCommand(HttpMethod method, String url, String name) {
+        if (method == HttpMethod.POST) {
```
mykola-mokhnach, line 298: maybe `switch` would look more readable here

SrinivasanTarget:
@jlipps fyi

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -34,13 +35,43 @@
...
+        final Response getSessions = driver.execute("getSessions");
+        assertNotNull(getSessions.getSessionId());
+    }
+
+    @Test
+    public void addCustomCommandWithSessionIdTest() {
+        driver.addCommand(HttpMethod.POST, "/session/" + driver.getSessionId() + "/appium/app/launch", "launchApplication");
```
KazuCocoa, line 58: ah, so, users should define a path with values that can change per session/per element etc every time before calling the name in Java...?
(Just a confirmation. seems reasonable in java so far.)

KazuCocoa (approved):

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
SrinivasanTarget: Yes Kazu, you are right.

KazuCocoa: 👍

saikrishna321 (approved):

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -288,6 +287,29 @@ public void rotate(ScreenOrientation orientation) {
...
+                MobileCommand.commandRepository.put(methodName, MobileCommand.getC(url));
+                break;
+            case POST:
+                MobileCommand.commandRepository.put(methodName, MobileCommand.postC(url));
+                break;
+            case DELETE:
+                MobileCommand.commandRepository.put(methodName, MobileCommand.deleteC(url));
+                break;
+            default:
+                throw new WebDriverException("Unsupported HTTP Method: " + httpMethod);
```
mykola-mokhnach, line 309: It would be also useful to print which methods are actually supported

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -241,6 +241,9 @@ public Response execute(Command command) throws WebDriverException {
                 }
             });
         }
+        if (getAdditionalCommands().containsKey(command.getName())) {
+            super.defineCommand(command.getName(), getAdditionalCommands().get(command.getName()));
```
mykola-mokhnach, line 245: should we verify if the defined command does not override an existing one?

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -34,13 +35,43 @@
...
+        final Response launchApplication = driver.execute("launchApplication");
+        assertNotNull(launchApplication.getSessionId());
+    }
+
+    @Test
+    public void addCustomCommandWithElementIdTest() {
```
mykola-mokhnach, line 64: 👍

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -34,13 +35,43 @@
...
+        IOSElement intA = driver.findElementById("IntegerA");
+        intA.clear();
+        driver.addCommand(HttpMethod.POST, "/session/" + driver.getSessionId() + "/appium/element/" + intA.getId() + "/value", "setNewValue");
+        Map<String, String> parameters = new HashMap<>();
```
mykola-mokhnach, line 68: nit: you could use ImmutableMap.of() notation to make it shorter

src/test/java/io/appium/java_client/ios/IOSDriverTest.java
```diff
@@ -34,13 +35,43 @@
...
+        driver.addCommand(HttpMethod.POST, "/session/" + driver.getSessionId() + "/appium/app/launch", "launchApplication");
+        final Response launchApplication = driver.execute("launchApplication");
+        assertNotNull(launchApplication.getSessionId());
+    }
+
+    @Test
+    public void addCustomCommandWithElementIdTest() {
+        IOSElement intA = driver.findElementById("IntegerA");
+        intA.clear();
+        driver.addCommand(HttpMethod.POST, "/session/" + driver.getSessionId() + "/appium/element/" + intA.getId() + "/value", "setNewValue");
```
mykola-mokhnach, line 67: nit: I usually prefer using String.format, as it is easier to read

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
SrinivasanTarget: At server side of plugin, we allow to override the behaviour of an Appium command though the path of command remain same. Keep the client side in line with server, I would like to allow client plugins also to override the existing command in same path with a different behaviour.

mykola-mokhnach (approved):

jlipps (commented):
I think it would help to have an example of how an actual client plugin might be used by end users. This PR seems to help plugin builders, without necessarily providing a recommended format. For example, would it be something like this? (pseudo java):

```java
CustomPlugin plugin = new CustomPlugin(driver);
driver.findElement(...); // do stuff with driver
plugin.myCustomCommand(...); // do stuff with plugin
```

Is there no way in Java to do some metaprogramming? Ideally we'd be able to use multiple plugins, something like this:

```java
CustomPlugin1.installTo(driver);
CustomPlugin2.installTo(driver);
driver.findElement(...); // do stuff with driver
driver.customCommand1(...); // do stuff with plugin 1
driver.customCommand2(...); // do stuff with plugin 2
```

I don't know Java well enough to know whether this is possible.

PR #1538, opened 2021-10-21
refactor: unify locator factories naming and `toString` implementations

valfirst:
## Change list

Keep the unified naming of methods and `toString` implementations.
 
## Types of changes

## Details
Unify locator factory-methods naming and `toString` implementations to follow the same approach as it's done in Selenium.


saikrishna321:
I like the idea to add Deprecate tag and also rename to AppiumBy

PR #1556, opened 2021-10-27
chore: switch to javac compiler

mykola-mokhnach:
## Change list

There is no need to use ECJ anymore after we get rid from selenium package override
 
## Types of changes


MrGizzle (commented):
Undo

PR #1566, opened 2021-10-29
chore: Include more tests into the pipeline

mykola-mokhnach:
## Change list

Some tests exist but never get executed by the CI.
 
## Types of changes


azure-pipelines.yml
```diff
@@ -61,3 +60,16 @@ jobs:
...
+    - task: Gradle@2
+      inputs:
+        gradleWrapperFile: 'gradlew'
+        gradleOptions: '-Xmx3072m'
+        javaHomeOption: 'JDKVersion'
+        jdkVersionOption: "$(JDK_VERSION)"
+        jdkArchitectureOption: 'x64'
+        publishJUnitResults: true
+        tasks: 'build'
+        options: 'miscTest -x checkstyleTest -x test -x signMavenJavaPublication'
```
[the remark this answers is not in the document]
valfirst: Created PR to check the assumption: https://github.com/appium/java-client/pull/1568, all builds are passed.

src/main/java/io/appium/java_client/remote/options/BaseMapOptionData.java
```diff
@@ -33,6 +35,11 @@ public BaseMapOptionData(Map<String, Object> options) {
         this.options = options;
     }
 
+    public BaseMapOptionData(String json) {
+        //noinspection unchecked
+        this((Map<String, Object>) new Gson().fromJson(json, Map.class));
```
valfirst, line 40: I guess it would be better to convert `Gson` object to a static field and reuse it here and in methods `toJson()` and `toString()`

src/test/java/io/appium/java_client/drivers/options/OptionsBuildingTest.java
```diff
@@ -0,0 +1,146 @@
...
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+@SuppressWarnings("ConstantConditions")
+public class OptionsBuildingTest {
+    @Test
+    public void canBuildXcuiTestOptions() throws MalformedURLException {
+        XCUITestOptions options = new XCUITestOptions();
+        assertEquals(options.getPlatformName(), Platform.IOS);
```
valfirst, line 50: `asssertEquals` accepts `expected` as the first parameter and `actual` as the second parameter

valfirst: and it seems it's relevant to all assertions in this class

src/main/java/io/appium/java_client/remote/options/BaseMapOptionData.java
mykola-mokhnach: done

src/test/java/io/appium/java_client/drivers/options/OptionsBuildingTest.java
mykola-mokhnach: fixed

SrinivasanTarget (approved):

PR #1591, opened 2021-11-16
feat: Add fileFieldName default value of ScreenRecordingUploadOptions

jaden-k:
## Change list
Added fileFieldName default value(file) of ScreenRecordingUploadOptions with some basic tests
 
## Types of changes

## Details
In v7.4.0 earlier versions,STOP_RECORDING_SCREEN command is not working since appium-server v1.21.0.
One of the ways to solve this problem is upgrading java-client and adding the fileFieldName option in test code.
It looks like being provided fileFieldName default value is the better one.

### DIFF
--- a/src/main/java/io/appium/java_client/screenrecording/ScreenRecordingUploadOptions.java
+++ b/src/main/java/io/appium/java_client/screenrecording/ScreenRecordingUploadOptions.java
@@ -32,6 +32,13 @@ public class ScreenRecordingUploadOptions {
     private Map<String, String> headers;
     private Map<String, Object> formFields;
 
+    /**
+     * Creates new instance with default fileFieldName.
+     */
+    public ScreenRecordingUploadOptions() {
+        withFileFieldName("file");
+    }
+
     public static ScreenRecordingUploadOptions uploadOptions() {
         return new ScreenRecordingUploadOptions();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/io/appium/java_client/screenrecording/ScreenRecordingUploadOptions.java
```diff
@@ -32,6 +32,13 @@
     private Map<String, String> headers;
     private Map<String, Object> formFields;
 
+    /**
+     * Creates new instance with default fileFieldName.
```
mykola-mokhnach, line 36: the default value for this option is already set to `file`: https://github.com/appium/appium-support/blob/04416e36df6e1c43f8fd00da7747372b03abb3c5/lib/net.js#L33

jaden-k: Right~ but, 'Content-Length' header is added by [this line](https://github.com/appium/appium-support/blob/44344b6ab96205538f4fb450bed54b67729a1a87/lib/net.js#L164).
Probably, it would occur [a problem](https://discuss.appium.io/t/need-support-uploading-of-screen-recording-to-opened-s3/33254) again that fileFieldName option is set to default value before [this line](https://github.com/appium/appium-support/blob/44344b6ab96205538f4fb450bed54b67729a1a87/lib/net.js#L164).
So, I made this PR.
Please let me know if you have any better ideas.

mykola-mokhnach: then it would make sense to change the way headers are set. I'll create a PR for it. You could anyway provide  a vlue to `withFileFieldName` in order to workaround the issue with the current implemenetation

mykola-mokhnach: https://github.com/appium/appium-support/pull/233 should do the trick

jaden-k:
uncalled-for~
appium-support(appium server) provide this feature

PR #1606, opened 2021-12-02
build: relax constraints of Selenium dependencies versions

valfirst:
## Change list

In a favor of flexibility the constraints of selenium dependencies versions are relaxed.
 
## Types of changes

## Details

https://docs.gradle.org/current/userguide/dependency_downgrade_and_exclude.html: "conflicting strict dependencies will trigger a build failure that you have to resolve".

An attempt to use `java-client:8.0.0-beta` in conjunction with Selenium 4.1.0 leads to build failures, like:
```
   > Could not resolve org.seleniumhq.selenium:selenium-remote-driver:{strictly 4.0.0}.
     Required by:
         project :xxxxxxx > io.appium:java-client:8.0.0-beta
```
In a favor of flexibility the constraints of selenium dependencies versions are relaxed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,17 +33,20 @@ dependencies {
 
     api ('org.seleniumhq.selenium:selenium-api') {
         version {
-            strictly "${seleniumVersion}"
+            strictly '[4.0, 5.0)'
+            prefer "${seleniumVersion}"
         }
     }
     api ('org.seleniumhq.selenium:selenium-remote-driver') {
         version {
-            strictly "${seleniumVersion}"
+            strictly '[4.0, 5.0)'
+            prefer "${seleniumVersion}"
         }
     }
     implementation ('org.seleniumhq.selenium:selenium-support') {
         version {
-            strictly "${seleniumVersion}"
+            strictly '[4.0, 5.0)'
+            prefer "${seleniumVersion}"
         }
     }
     implementation 'com.google.code.gson:gson:2.8.9'
@@ -63,7 +66,8 @@ dependencies {
     }
     testImplementation ('org.seleniumhq.selenium:selenium-chrome-driver') {
         version {
-            strictly "${seleniumVersion}"
+            strictly '[4.0, 5.0)'
+            prefer "${seleniumVersion}"
         }
     }
 }


SrinivasanTarget (approved):

mykola-mokhnach:
We could try that, although there were already issues about several breaking selenium API updates because of incompatible changes in their API. 

Basically the question is how big is the probability that the next selenium version (minor,  major or patch) is going to be incompatible to the current Appium client. And from my perspective such probability could be pretty high, since we don't only use public APIs, but rather perform many reflective calls to private/protected stuff and hack selenium lib here and there.

mykola-mokhnach:
Maybe for now we could declare a strict version range like in https://docs.gradle.org/current/userguide/single_versions.html (Example 2):

```
version {
           strictly '[4, 5)'
        }
```

valfirst:
@mykola-mokhnach thank you for your comments, I really appreciate them, I've updated the PR following your hints

build.gradle
```diff
@@ -33,17 +33,20 @@ dependencies {
...
         version {
-            strictly "${seleniumVersion}"
+            strictly "[${seleniumVersion}, 5.0)"
+            prefer "${seleniumVersion}"
         }
     }
     api ('org.seleniumhq.selenium:selenium-remote-driver') {
         version {
-            strictly "${seleniumVersion}"
+            strictly "[${seleniumVersion}, 5.0)"
```
mykola-mokhnach, line 42: not sure if we have to put 4.0 there instead...
`seleniumVersion` in such case would be the version we tested with and verified working, so it might be higher than 4.0, e.g. 4.1, although it does not mean that 4.0 should necessarily stop working

build.gradle
```diff
@@ -33,17 +33,20 @@ dependencies {
 
     api ('org.seleniumhq.selenium:selenium-api') {
         version {
-            strictly "${seleniumVersion}"
+            strictly "4.0, 5.0)"
```
mykola-mokhnach, line 36: missing bracket here

mykola-mokhnach (approved):

Issue #1660, opened 2022-03-14
AppiumDriverLocalService is not able to start - checkStatus is failing


robinGupta11392:
## Description

Hi All, I have observed that in Appium 8 we are facing issue while trying to start the Appium server programatically. As per the debug analysis, I could see that the Appium server is starting fine but the checkStatus call is failing because the URL is not formatted properly in case of a basepath. I am using Appium v1.22.0 CLI where I am providing the BasePath flag as "/wd/hub". I noticed that while starting the server, client is passing that as server argument but when it's trying to check the status it's always getting 404 because the code is trying to ping "/wd/status" instead of "/wd/hub/status". I looked up the code then realised that current addSufix code may need a fix to prevent the removal of "/hub" from the endpoint.

## Environment

* Java client build version or git revision if you use some snapshot: v8.0.0
* Appium server version or git revision if you use some snapshot: v1.22.0
* Desktop OS/version used to run Appium if necessary: Mac
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: N/A
* Mobile platform/version under test: N/A
* Real device or emulator/simulator: N/A

## Details

Noticed that while starting the server, client is passing that as server argument but when it's trying to check the status it's always getting 404 because the code is trying to ping "/wd/status" instead of "/wd/hub/status". I looked up the code then realised that current addSufix code may need a fix to prevent the removal of "/hub" from the endpoint.

## Code To Reproduce Issue

AppiumDriverLocalService service = buildService(new AppiumServiceBuilder()
        .withIPAddress("127.0.0.1")
        .usingPort(Integer.parseInt("4723"))
        .withArgument(GeneralServerFlag.BASEPATH, "/wd/hub")
        .withAppiumJS(new File(APPIUM_JS_FILEPATH)) // replaced with appium CLI path (main.js)
        .withArgument(GeneralServerFlag.LOG_LEVEL, "debug")
    );
service.start()

## Exception Stacktraces

https://gist.github.com/robinGupta11392/9b8780adda59ffe283b4f1fd6f428a93#file-gistfile1-txt

robinGupta11392:
As a workaround, I am using ".withArgument(GeneralServerFlag.BASEPATH, "/")" and its able to startup now. But with this I had to change the Appium endpoint configuration to "http://127.0.0.1:4723/" and have to remove "/wd/hub" from everywhere. Do note that it don't work with Blank String ("") as the basepath from the java_client lib but on command line it works.

[...]

PR #1661, opened 2022-03-17
fix: #1660 - correction for ping method to get proper status URL in case of a basepath provided

robinGupta11392:
## Change list

Fixed the logic of appending basepath to the Hub URL which fixes Issue #1660 - https://github.com/appium/java-client/issues/1660.

AppiumDriverLocalService was not able to start with Appium CLI versions < 2 as the ping() method was not able to get the status URL properly. Fixed the logic of appending basepath to sanitise the user input and appending basepath with "/" in the end, if not provided by user.
 
## Types of changes

## Details

Test Execution logs - https://gist.github.com/robinGupta11392/e8c37c7e10e03cd00a4d44d398402738
Issue fixed - https://github.com/appium/java-client/issues/1660
Changed Classes:
 - ServerBuilderTest
 - AppiumDriverLocalService
 
Method Added:
 - sanitizeBasePath
 
Method Modified:
 - withBasePath

Test Added:
 - checkAbilityToStartServiceUsingBasePath


src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -94,10 +94,23 @@ public static AppiumDriverLocalService buildService(AppiumServiceBuilder builder
...
         return this;
     }
 
+    @SneakyThrows private static String sanitizeBasePath(String basePath) {
+        if (null == basePath) {
+            LOG.warn("Base Path cannot be NULL -- ignoring the basepath configuration");
+            return null;
+        }
+        basePath = basePath.trim();
+        if (basePath.isBlank() || basePath.isEmpty()) {
```
valfirst, line 107: `String#isBlank` is available starting from Java 11 only.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -94,10 +94,23 @@ public static AppiumDriverLocalService buildService(AppiumServiceBuilder builder
...
+            LOG.warn("Base Path cannot be Empty or Blank -- ignoring the basepath configuration");
+            return null;
+        }
+        return (basePath.endsWith("/") ? basePath : basePath + "/");
```
valfirst, line 111: redundant parenthesis
```suggestion
        return basePath.endsWith("/") ? basePath : basePath + "/";
```

valfirst (changes requested):

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -94,10 +94,23 @@ public static AppiumDriverLocalService buildService(AppiumServiceBuilder builder
     }
 
     public AppiumDriverLocalService withBasePath(String basePath) {
-        this.basePath = basePath;
+        this.basePath = sanitizeBasePath(basePath);
         return this;
     }
 
+    @SneakyThrows private static String sanitizeBasePath(String basePath) {
+        if (null == basePath) {
```
mykola-mokhnach, line 102: Consider using checkNotNull helper from guava lib. An exception must be thrown if this argument is set to null

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
mykola-mokhnach: Apache commons lib also provides isBlank method

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
```diff
@@ -94,10 +94,23 @@ public static AppiumDriverLocalService buildService(AppiumServiceBuilder builder
     }
 
     public AppiumDriverLocalService withBasePath(String basePath) {
-        this.basePath = basePath;
+        this.basePath = sanitizeBasePath(basePath);
         return this;
     }
 
+    @SneakyThrows private static String sanitizeBasePath(String basePath) {
```
mykola-mokhnach, line 101: what does this method throw?

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: I believe the default value for basePath is null in case the user doesn't provide a basepath at all.. I referred to the getUrl method in which its checking for null. we should not throw an exception in this case (in my opinion) rather a warning message should be good enough and appium can start with no additional basepath

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: I will remove the SneakyThrows, it came by mistakenly copy paste. thanks for highlighting.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: But yeah, we can highlight to user by throwing an exception that they have provided a null value in basepath, so that they can make an informed decision to either provide a valid value or no value at all. I will make the change, thanks.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: quick question on this - do we have any log/documentation which informs user that they need to provide a basepath in order to use Appium CLI v < v2? 
I believe we can pass a blank/empty string to it and it still works fine.

robinGupta11392: Just noticed that Blank String as argument values are not added to the appium command line while starting "main.js", will throw an exception if user tries to configure basepath as blank.

Below code which doesn't add value to the command line in case of blank

 ```
@Override
    protected ImmutableList<String> createArgs() {
...
Set<Map.Entry<String, String>> entries = serverArguments.entrySet();
        for (Map.Entry<String, String> entry : entries) {
...
if (!StringUtils.isBlank(value)) {
                argList.add(value);
            }
...
```

File: src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java (hunk 3 above)
Comment (mykola-mokhnach) [reply src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java:102]: the only place this API is called is `withBasePath` builder method. And there basePath must not be null. In general we are quite careful about explicitly returning null from any method and try to avoid that as much as possible.

File: src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java (hunk 1 above)
Comment (mykola-mokhnach) [reply src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java:107]: > quick question on this - do we have any log/documentation which informs user that they need to provide a basepath in order to use Appium CLI v < v2?
> I believe we can pass a blank/empty string to it and it still works fine.

https://github.com/appium/java-client/blob/master/docs/v7-to-v8-migration-guide.md#appiumdriverlocalservice

File: src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java [hunk 5]
```diff
@@ -94,10 +94,23 @@ public static AppiumDriverLocalService buildService(AppiumServiceBuilder builder
...
+        if (null == basePath) {
+            LOG.warn("Base Path cannot be NULL -- ignoring the basepath configuration");
+            return null;
+        }
+        basePath = basePath.trim();
+        if (basePath.isBlank() || basePath.isEmpty()) {
+            LOG.warn("Base Path cannot be Empty or Blank -- ignoring the basepath configuration");
```
mykola-mokhnach, line 108: I usually avoid writing logs in the library code unless really necessary. They might be annoying and hard to disable.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
mykola-mokhnach: In general I'd add something like `checkArgument(isBlank(basePath), "basePath must not be empty or null")`

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: Fixed this in the new commit, please review.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: Fixed this by moving the validations to "withArgument" method and checking for null, blank and empty values. Introduced a InvalidBasePathException to throw in case of any validations fail.

src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java
robinGupta11392: Removed the LOG in the new commit.

robinGupta11392:
https://gist.github.com/robinGupta11392/19cb22749170119d791cbeb8e1a714b3 --> ServerBuilderTest execution logs with Appium V2 Beta 25 post the fix. It seems there is one Log test failing right from the moment i cloned it - maybe you guys can have a look for reference.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -217,13 +218,25 @@ public AppiumServiceBuilder withArgument(ServerArgument argument, String value)
...
                 serverArguments.put(argName, value);
                 break;
         }
         return this;
     }
 
+    private static String sanitizeBasePath(String basePath) {
+        basePath = checkNotNull(basePath).trim();
+        if (StringUtils.isBlank(basePath) || basePath.isEmpty()) {
+            throw new InvalidBasePathException(
```
mykola-mokhnach, line 234: IMHO it's an overkill to have a separate exception there. the standard InvalidArgumentException thrown by checkArgument would be totally fine

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -217,13 +218,25 @@ public AppiumServiceBuilder withArgument(ServerArgument argument, String value)
...
             default:
                 serverArguments.put(argName, value);
                 break;
         }
         return this;
     }
 
+    private static String sanitizeBasePath(String basePath) {
+        basePath = checkNotNull(basePath).trim();
+        if (StringUtils.isBlank(basePath) || basePath.isEmpty()) {
```
mykola-mokhnach, line 233: isBlank also returns true for empty strings

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +319,56 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
     }
+
+    @Test
+    public void checkAbilityToStartServiceUsingValidBasePathWithMultiplePathParams() {
+        String baseUrl = String.format("http://%s:%d/", BROADCAST_IP_ADDRESS, DEFAULT_APPIUM_PORT);
+        String basePath = "wd/hub";
+        service = new AppiumServiceBuilder().withArgument(BASEPATH, basePath).build();
+        service.start();
+        assertTrue(service.isRunning());
+        assertEquals(baseUrl + basePath + "/",service.getUrl().toString());
```
mykola-mokhnach, line 330: Please format the stuff everywhere, so there are spaces after comma

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -217,13 +218,25 @@ public AppiumServiceBuilder withArgument(ServerArgument argument, String value)
             case "-g":
                 withLogFile(new File(value));
                 break;
+            case "--base-path":
+                serverArguments.put(argName,sanitizeBasePath(value));
```
valfirst, line 222:
```suggestion
                serverArguments.put(argName, sanitizeBasePath(value));
```

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
valfirst, line 233: no need to check for blank value after trim operation
```suggestion
        basePath = checkNotNull(basePath).trim();
        if (basePath.isEmpty()) {
```

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +319,56 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
+        assertTrue(service.isRunning());
+        assertEquals(baseUrl + basePath.substring(1) ,service.getUrl().toString());
+    }
+
+    @Test
+    public void checkAbilityToValidateBasePathForEmptyBasePath() {
+        try {
+            service = new AppiumServiceBuilder().withArgument(BASEPATH, "").build();
+            fail("Base path was not validated for Blank or Empty string");
+        } catch (Exception e) {
```
valfirst, line 348: use `org.junit.Assert#assertThrows(java.lang.Class<T>, org.junit.function.ThrowingRunnable)` instead

valfirst (changes requested):

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
robinGupta11392: Noted on this, take care in the latest commit.

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Noted on this, take care in the latest commit.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
robinGupta11392: Noted on this, take care in the latest commit.

robinGupta11392: Noted on this, taken care in the latest commit.

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
robinGupta11392: Noted on this, taken care in the latest commit.

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -54,6 +57,9 @@
     private static final Path PATH_T0_TEST_MAIN_JS = ROOT_TEST_PATH
             .resolve("service").resolve("local").resolve("main.js");
 
+    private static final String INVALID_BASE_PATH_ERROR_MESSAGE =
```
mykola-mokhnach, line 60: it is usually ok to just verify the exception class. Tests become too flaky because string messages could be changed too often

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -54,6 +40,17 @@
...
+import lombok.SneakyThrows;
+import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.SystemUtils;
+import org.apache.commons.validator.routines.InetAddressValidator;
+import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.Platform;
+import org.openqa.selenium.os.ExecutableFinder;
+import org.openqa.selenium.remote.Browser;
+import org.openqa.selenium.remote.service.DriverService;
```
valfirst, line 53: a lot of unnecessary changes in the import order

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -217,13 +214,25 @@ public AppiumServiceBuilder withArgument(ServerArgument argument, String value)
...
         }
         return this;
     }
 
+    private static String sanitizeBasePath(String basePath) {
+        basePath = checkNotNull(basePath).trim();
+        if (basePath.isEmpty()) {
+            throw new IllegalArgumentException(
+                "Given base path is not valid - blank or empty values are not allowed for base path");
+        }
```
valfirst, line 232:
```suggestion
        Preconditions.checkArgument(!basePath.isEmpty(), "Given base path is not valid - blank or empty values are not allowed for base path");
```

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -18,22 +21,22 @@
...
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.OutputStream;
 import java.nio.file.Path;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.List;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
```
valfirst, line 39: unnecessary changes in the imports order

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +318,47 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
+        service = new AppiumServiceBuilder().withArgument(BASEPATH, basePath).build();
+        service.start();
+        assertTrue(service.isRunning());
+        assertEquals(baseUrl + basePath.substring(1), service.getUrl().toString());
+    }
+
+    @Test
+    public void checkAbilityToValidateBasePathForEmptyBasePath() {
+        Exception exception = assertThrows(IllegalArgumentException.class, () -> {
+            service = new AppiumServiceBuilder().withArgument(BASEPATH, "").build();
```
valfirst, line 345: no need to init `service` variable here

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +318,47 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
+        });
+        assertEquals(INVALID_BASE_PATH_ERROR_MESSAGE, exception.getMessage());
+    }
+
+    @Test
+    public void checkAbilityToValidateBasePathForBlankBasePath() {
+        Exception exception = assertThrows(IllegalArgumentException.class, () -> {
+            service = new AppiumServiceBuilder().withArgument(BASEPATH, "   ").build();
```
valfirst, line 353: no need to init `service` variable here

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +318,47 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
+        });
+        assertEquals(INVALID_BASE_PATH_ERROR_MESSAGE, exception.getMessage());
+    }
+
+    @Test
+    public void checkAbilityToValidateBasePathForNullBasePath() {
+        assertThrows(NullPointerException.class, () -> {
+            service = new AppiumServiceBuilder().withArgument(BASEPATH, null).build();
```
valfirst, line 361: no need to init `service` variable here

valfirst (changes requested):

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
robinGupta11392: Import order got changed as per alphanumeric order after I formatted the code as per the check-style recommended, should I change it back?

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Updated in the new commit, thanks for highlighting

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Updated in the new commit, thanks for highlighting

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Updated in the new commit, thanks for highlighting

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Updated in the new commit, thanks for highlighting

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
robinGupta11392: Updated in the new commit

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Updated in the new commit, thanks for highlighting

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
```diff
@@ -312,4 +316,45 @@ public void checkAbilityToStartServiceWithLogFileUsingShortFlag() {
...
+        String basePath = "/wd/";
+        service = new AppiumServiceBuilder().withArgument(BASEPATH, basePath).build();
+        service.start();
+        assertTrue(service.isRunning());
+        assertEquals(baseUrl + basePath.substring(1), service.getUrl().toString());
+    }
+
+    @Test
+    public void checkAbilityToValidateBasePathForEmptyBasePath() {
+        assertThrows(IllegalArgumentException.class, () -> {
```
mykola-mokhnach, line 342: FYI: you could add expected exception signature to Test annotations: https://www.guru99.com/junit-exception-test.html

mykola-mokhnach (approved):

src/test/java/io/appium/java_client/service/local/ServerBuilderTest.java
robinGupta11392: Noted with thanks.

valfirst (approved):

PR #1735, opened 2022-08-05
feat: add creating a driver with ClientConfig

KazuCocoa:
## Change list

Current Java client does not have a way to generates a http client with `ClientConfig clientConfig`.
 When we give `URL remoteAddress` to the AppiumDriver, it generates a default `clientConfig` as below to create a http client instance. So, in case users want to configure proxy etc to the http client, they need to define their own http factory method for now.
 
```
                        ClientConfig.defaultConfig()
                                .baseUrl(Require.nonNull("Server URL", ofNullable(service)
                                        .map(DriverService::getUrl)
                                        .orElse(addressOfRemoteServer)))
                                .readTimeout(DEFAULT_READ_TIMEOUT),
```

It seems like recent selenium v4 java client allows users to configure the http client (with the default netty client) by giving clientConfig like:

```
ClientConfig config = ClientConfig.defaultConfig()
        .connectionTimeout(Duration.ofMinutes(5))
        .readTimeout(Duration.ofMinutes(3));

WebDriver driver = RemoteWebDriver.builder()
        .oneOf(browserOptions)
        .setCapability("sauce:options", sauceOptions)
        .address("https://ondemand.us-west-1.saucelabs.com/wd/hub")
        .config(config)
        .build();
```
https://docs.saucelabs.com/web-apps/automated-testing/selenium/selenium4/

Current out method does not provide the same syntax, but it would be worth giving users to configure the http client with `ClientConfig` without defining HTTP factory method. Many examples in the internet explain factory method with okhttp, but selenium v4 no longer supports okhttp by default. So this ClientConfig method helps users to configure client proxy etc than defining their own factory.

In case a user wants to define their own http client, they can use existing method to start a driver with `HttpClient.Factory`

## Types of changes


src/main/java/io/appium/java_client/android/AndroidDriver.java
```diff
@@ -182,6 +183,27 @@ public AndroidDriver(HttpClient.Factory httpClientFactory, Capabilities capabili
...
 
+    /**
+     * Creates a new instance based on the given ClientConfig and {@code capabilities}.
+     * The HTTP client is default client generated by {@link HttpCommandExecutor#getDefaultClientFactory}.
+     *
+     * Example:
+     *
+     *  ClientConfig clientConfig = ClientConfig.defaultConfig()
+     *    .baseUri(URI.create("WebDriver URL"))
+     *    .proxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress("your.proxy.ip", 8080)))
```
mykola-mokhnach, line 194: We anyway don't support proxies, so this documentation might be confusing for users

mykola-mokhnach: same below

src/main/java/io/appium/java_client/windows/WindowsDriver.java
```diff
@@ -79,6 +80,28 @@ public WindowsDriver(HttpClient.Factory httpClientFactory, Capabilities capabili
...
+     * Creates a new instance based on the given ClientConfig and {@code capabilities}.
+     * The HTTP client is default client generated by {@link HttpCommandExecutor#getDefaultClientFactory}.
+     *
+     * Example:
+     *
+     *  ClientConfig clientConfig = ClientConfig.defaultConfig()
+     *    .baseUri(URI.create("WebDriver URL"))
+     *    .proxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress("your.proxy.ip", 8080)))
+     *    .readTimeout(Duration.ofMinutes(5));
+     *    options = WindowsOptions options = new WindowsOptions();
```
mykola-mokhnach, line 93: extra leading spaces here
Consider formatting the code section though: https://reflectoring.io/howto-format-code-snippets-in-javadoc/

KazuCocoa: changed to use pre

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -62,24 +62,35 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
...
+                clientConfig != null
+                        ? clientConfig
+                        : ClientConfig.defaultConfig()
+                            .baseUrl(Require.nonNull("Server URL", ofNullable(service)
+                                    .map(DriverService::getUrl)
+                                    .orElse(addressOfRemoteServer)))
+                            .readTimeout(DEFAULT_READ_TIMEOUT),
+                httpClientFactory != null
+                        ? httpClientFactory
+                        : getDefaultClientFactory()
```
valfirst, line 77:
```suggestion
                ofNullable(httpClientFactory).orElseGet(this::getDefaultClientFactory)
```

KazuCocoa: 👍 InteiJ suggested changing `this` to `HttpCommandExecutor` as static, so I followed it in bd780ce

mykola-mokhnach (approved):

SrinivasanTarget (approved):

valfirst (approved):

PR #1747, opened 2022-08-25
feat: add directConnect feature

KazuCocoa:
## Change list

I'd like to add https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments for Java client. Other clients such as webdriverio, Ruby and Python already have them for long. If the response had directConnect capabilities, the appium client attempt to send requests to the directConnect one instead of the original URL.

- directConnectProtocol
- directConnectHost
- directConnectPort
- directConnectPath

or `appium:` prefixed ones.

## Types of changes

## Details

The main change is `setDirectConnect` in src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
To allow users to turn it on/off, I've added `AppiumClientConfig` and give the instance to the driver.

The example code is like...
```
     * AppiumClientConfig clientConfig = AppiumClientConfig.defaultConfig()
     *     .baseUri(URI.create("WebDriver URL"))
     *     .readTimeout(Duration.ofMinutes(5));
     * WindowsOptions options = new WindowsOptions();
     * WindowsDriver driver = new WindowsDriver(clientConfig, options);
```

By inheriting `ClientConfig`, we can set a custom User Agent as well. This PR includes the way to set the UA like `appium/8.3.0 (selenium/4.5.0 (java mac))`

The naming and implementation methods in AppiumClientConfig follow ClientConfig as possible.
`ImeHandler` removal in the test code was selenium 4.5.0 removed it.


src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -88,46 +88,94 @@ public AppiumDriver(ClientConfig clientConfig, Capabilities capabilities) {
         this(new AppiumCommandExecutor(MobileCommand.commandRepository, clientConfig), capabilities);
     }
 
+    public AppiumDriver(ClientConfig clientConfig, Capabilities capabilities, AppiumClientConfig appiumClientConfig) {
```
mykola-mokhnach, line 91: There are too many permutations. Lets only have one constructor with all possible arguments and add other if this is really needed/requested

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+        return response;
     }
 
     public void refreshAdditionalCommands() {
         getAdditionalCommands().forEach(this::defineCommand);
     }
 
+    @SuppressWarnings("unchecked")
+    private void setDirectConnect(Response response) throws MalformedURLException {
+        if (!this.appiumClientConfig.isDirectConnectEnabled()) {
```
mykola-mokhnach, line 227: I would rather move this check to the place where this method is called for better visibility

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+            return;
+        }
+
+        Map<String, ?> responseValue = (Map<String, ?>) response.getValue();
+
+        String directConnectProtocol = getDirectConnectValue(responseValue, "directConnectProtocol");
```
mykola-mokhnach, line 233: I assume such feature could be a huge security hole in case of non-https connections. Please convince me that I'm wrong.

Also, is it specced officially somewhere?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+                || directConnectPath == null || directConnectPort == null) {
+            return;
+        }
+        setClientWithURL(new URL(directConnectProtocol + "://" + directConnectHost + ":"
+                + directConnectPort + directConnectPath));
+    }
+
+    private String getDirectConnectValue(Map<String, ?> responseValue, String key) {
+        String directConnectPath = String.valueOf(responseValue.get("appium:" + key));
+        return directConnectPath == null ? String.valueOf(responseValue.get(key)) : directConnectPath;
```
mykola-mokhnach, line 248: according to the code above none of direct connect attributes could be null

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+        if (directConnectProtocol == null || directConnectHost == null
+                || directConnectPath == null || directConnectPort == null) {
+            return;
+        }
+        setClientWithURL(new URL(directConnectProtocol + "://" + directConnectHost + ":"
+                + directConnectPort + directConnectPath));
+    }
+
+    private String getDirectConnectValue(Map<String, ?> responseValue, String key) {
+        String directConnectPath = String.valueOf(responseValue.get("appium:" + key));
```
mykola-mokhnach, line 247: can we use a constant instead?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+
+        if (directConnectProtocol == null || directConnectHost == null
+                || directConnectPath == null || directConnectPort == null) {
+            return;
+        }
+        setClientWithURL(new URL(directConnectProtocol + "://" + directConnectHost + ":"
+                + directConnectPort + directConnectPath));
+    }
+
+    private String getDirectConnectValue(Map<String, ?> responseValue, String key) {
```
mykola-mokhnach, line 246: I assume the result of this method is nullable. Please mark it with appropriate annotation

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
         setCommandCodec(new AppiumW3CHttpCommandCodec());
         refreshAdditionalCommands();
         setResponseCodec(dialect.getResponseCodec());
-        return result.createResponse();
+        Response response = result.createResponse();
+
```
mykola-mokhnach, line 215: there is no need for extra empty lines

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -147,6 +186,12 @@ protected HttpClient getClient() {
         return getPrivateFieldValue(HttpCommandExecutor.class, "client", HttpClient.class);
     }
 
+    protected void setClientWithURL(URL serverUrl) {
```
mykola-mokhnach, line 189: how about `overrideClientUrl`

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -60,10 +62,13 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
...
     private final Optional<DriverService> serviceOptional;
 
+    private final AppiumClientConfig appiumClientConfig;
+
     private AppiumCommandExecutor(Map<String, CommandInfo> additionalCommands, DriverService service,
                                   URL addressOfRemoteServer,
                                   HttpClient.Factory httpClientFactory,
-                                  ClientConfig clientConfig) {
+                                  ClientConfig clientConfig,
+                                  AppiumClientConfig appiumClientConfig) {
```
valfirst, line 71: Does it make sense to inherit `ClientConfig` in `AppiumClientConfig` and then use only `AppiumClientConfig`?

Old methods with `ClientConfig` can be marked as `@Deprecated`

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -147,6 +186,12 @@ protected HttpClient getClient() {
         return getPrivateFieldValue(HttpCommandExecutor.class, "client", HttpClient.class);
     }
 
+    protected void setClientWithURL(URL serverUrl) {
+        HttpClient.Factory httpClientFactory = getPrivateFieldValue(HttpCommandExecutor.class,
+                "httpClientFactory", HttpClient.Factory.class);
+        setPrivateFieldValue(HttpCommandExecutor.class, "client", httpClientFactory.createClient(serverUrl));
```
mykola-mokhnach, line 192: I'm not quite sure we should create a new client here. Can we simply replace the url for the existing one?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
SrinivasanTarget: +1.  We have not released a new version post introducing client config so we don't have to deprecate it. We can have AppiumClientConfig inherit ClientConfig and use it across.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: It could be. We can restrict it to accept only HTTPS, but basically, the response depends on the server side's implementation.

Maybe not explicitly. Ordinally this was experimentally added as https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments in some clients like https://github.com/appium/python-client/pull/338 a couple of years ago. So far, Java and C# clients haven't had this. It seems like this was used than a similar experimental execute_driver, btw.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +211,43 @@ private Response createSession(Command command) throws IOException {
...
+
+        return response;
     }
 
     public void refreshAdditionalCommands() {
         getAdditionalCommands().forEach(this::defineCommand);
     }
 
+    @SuppressWarnings("unchecked")
+    private void setDirectConnect(Response response) throws MalformedURLException {
```
mykola-mokhnach, line 226: we must wrap this exception and provide a proper error message to the user if the server did not return correct values to construct the url

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: also, do we only allow `appium` prefix? I thought it could be any vendor prefix

mykola-mokhnach:
I assume a vanilla redirect (code 302) would work better for such purpose than all this logic with response parsing and multiple values

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: Maybe possible. I just followed https://github.com/webdriverio/webdriverio/blob/6bbe1cf371f06e517ac62a4e02724c8b610803cb/packages/webdriver/src/utils.ts#L292

Afaik, ordinally these caps did not have any vendor prefix.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: Perhaps could be. The ClientConfig is in selenium, so I defined only Appium thing separately

KazuCocoa:
I agree with your intention basically, but I think the behavior was a bit different by this.:
Redirect changes the url after calling the default webdriver URL by getting 302, so basically, every request goes through the default webdrier URL (and redirects to another URL).
https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments idea was a new session request went thought the webdriver URL and created a session against behind a middle server (managed distributed hosts/devices like selenium grid), and then following request in the session would reach directConnect one directly instead of the original webdrievr URL to reduce HTTP request's latency.

mykola-mokhnach:
> I agree with your intention basically, but I think the behavior was a bit different by this.:
> Redirect changes the url after calling the default webdriver URL by getting 302, so basically, every request goes through the default webdrier URL (and redirects to another URL).
> https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments idea was a new session request went thought the webdriver URL and created a session against behind a middle server (managed distributed hosts/devices like selenium grid), and then following request in the session would reach directConnect one directly instead of the original webdrievr URL to reduce HTTP request's latency.

Thanks for the explanation. In such case redirect is probably won't work, although I still don't see an actual need to provide 4 different values if we still need to combine them into a single url. It would probably too late to change the current implementation as it is already present in other clients

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: I don't see any difference as we anyway must access a private field here

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: lets start by only accepting https as protocol

KazuCocoa:
thanks.
Yea, I came up with the same thought after a while... > a single url

KazuCocoa:
Changed to draft to update reviews

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,89 @@
...
+import java.time.Duration;
+
+/**
+ * Manage Appium Client configurations.
+ */
+
+public class AppiumClientConfig {
+    private boolean directConnect;
+
+    private ClientConfig clientConfig;
```
mykola-mokhnach, line 30: not sure what would be better here - composition or inheritance

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+    private static final Duration DEFAULT_READ_TIMEOUT = Duration.ofMinutes(10);
+
+    public AppiumClientConfig() {
+        this.clientConfig = ClientConfig.defaultConfig().readTimeout(DEFAULT_READ_TIMEOUT);
```
mykola-mokhnach, line 35: can we assign these variables upon definition?

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,89 @@
...
+        this.directConnect = false;
+    }
+
+    public static class DirectConnect {
```
mykola-mokhnach, line 39: what is the purpose of this static class? variables could be also defined in the main one (don't forget to add "final" though)

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Whether enable directConnect feature described in
+     * <a href="https://appiumpro.com/editions/86-connecting-directly-to-appium-hosts-in-distributed-environments">
+     *     Connecting Directly to Appium Hosts in Distributed Environments</a>.
+     *
+     * @param directConnect if enable the directConnect feature
+     * @return A self reference
+     */
+    public AppiumClientConfig directConnect(boolean directConnect) {
```
mykola-mokhnach, line 62: lets use a common approach and have `set` prefix for all builder setter methods

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,89 @@
...
+     * Set the given {@link ClientConfig} for the HTTP client instance.
+     * @param clientConfig to keep it in AppiumClientConfig instance
+     * @return A self reference
+     */
+    public AppiumClientConfig setHttpClientConfig(ClientConfig clientConfig) {
+        this.clientConfig = clientConfig;
+        return this;
+    }
+
+    public ClientConfig getHttpClientConfig() {
```
mykola-mokhnach, line 86: missing docstring?

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -84,48 +83,22 @@ public AppiumDriver(HttpCommandExecutor executor, Capabilities capabilities) {
...
-
-    public AppiumDriver(HttpClient.Factory httpClientFactory, Capabilities capabilities) {
-        this(AppiumDriverLocalService.buildDefaultService(), httpClientFactory,
-                capabilities);
-    }
-
-    public AppiumDriver(Capabilities capabilities) {
-        this(AppiumDriverLocalService.buildDefaultService(), capabilities);
+    protected AppiumDriver(
+            @Nullable URL remoteAddress,
```
mykola-mokhnach, line 87: I'm not quite sure about nullability of these arguments

src/main/java/io/appium/java_client/gecko/GeckoDriver.java
```diff
@@ -84,20 +133,24 @@ public GeckoDriver(HttpClient.Factory httpClientFactory, Capabilities capabiliti
...
+     * GeckoDriver driver = new GeckoDriver(options, appiumClientConfig);
      *
      * </pre>
      *
-     * @param clientConfig take a look at {@link ClientConfig}
      * @param capabilities take a look at {@link Capabilities}
+     * @param appiumClientConfig take a look at {@link AppiumClientConfig}
      *
      */
-    public GeckoDriver(ClientConfig clientConfig, Capabilities capabilities) {
```
mykola-mokhnach, line 96: Please do not change the original sequence of arguments, because this would be a breaking change then

mykola-mokhnach: Also AppiumClientConfig must be inherited from ClientConfig in order to make this contructor backward compatible

src/main/java/io/appium/java_client/mac/Mac2Driver.java
```diff
@@ -95,21 +138,24 @@ public Mac2Driver(HttpClient.Factory httpClientFactory, Capabilities capabilitie
...
      *
-     * @param clientConfig take a look at {@link ClientConfig}
      * @param capabilities take a look at {@link Capabilities}
+     * @param appiumClientConfig take a look at {@link AppiumClientConfig}
      *
      */
-    public Mac2Driver(ClientConfig clientConfig, Capabilities capabilities) {
-        super(clientConfig, ensurePlatformAndAutomationNames(
-                capabilities, PLATFORM_NAME, AUTOMATION_NAME));
+    public Mac2Driver(Capabilities capabilities, AppiumClientConfig appiumClientConfig) {
```
mykola-mokhnach, line 152: same as above

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -60,45 +65,97 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
...
+            @Nullable URL addressOfRemoteServer,
+            @Nullable HttpClient.Factory httpClientFactory,
+            @Nullable AppiumClientConfig appiumClientConfig) {
+
         super(additionalCommands,
-                ofNullable(clientConfig).orElse(
-                        ClientConfig.defaultConfig()
-                                .baseUrl(Require.nonNull("Server URL", ofNullable(service)
+                ofNullable(appiumClientConfig)
+                        .orElse(AppiumClientConfig.defaultConfig())
```
mykola-mokhnach, line 91: Please use orElseGet if the argument is not a constant.
By using orElse an unnecessary instance is created even though it might not be used.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -147,6 +204,18 @@ protected HttpClient getClient() {
...
     }
 
+    /**
+     * Override the http client in the HttpCommandExecutor class with a new http client instance with the given URL.
+     * Use the same http client factory and client config if the constructor got them.
+     *
+     * @param serverUrl to set the URL as the new client's base url.
+     */
+    protected void overrideServerUrl(URL serverUrl) {
+        setPrivateFieldValue(HttpCommandExecutor.class, "client",
```
mykola-mokhnach, line 208: so it is not possible to replace URL for an exiting client without replacing the client instance itself?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +235,58 @@ private Response createSession(Command command) throws IOException {
...
+                AppiumClientConfig.DirectConnect.DIRECT_CONNECT_HOST);
+        String directConnectPort = getDirectConnectValue(responseValue,
+                AppiumClientConfig.DirectConnect.DIRECT_CONNECT_PORT);
+
+        if (directConnectProtocol == null || directConnectHost == null
+                || directConnectPath == null || directConnectPort == null) {
+            return;
+        }
+
+        if (!directConnectProtocol.contentEquals("https")) {
```
mykola-mokhnach, line 261: should we use `equals` instead?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +235,58 @@ private Response createSession(Command command) throws IOException {
...
+            throw new WebDriverException(
```
mykola-mokhnach, line 262: I would rather use sessionNotCreated exception in such case

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +235,58 @@ private Response createSession(Command command) throws IOException {
...
+                    String.format("The protocol must be https. %s was given.", directConnectProtocol));
+        }
+
+        URL newUrl;
+        String newUrlCandidate = String.format("%s://%s:%s%s",
+                directConnectProtocol, directConnectHost, directConnectPort, directConnectPath);
+        try {
+            newUrl = new URL(newUrlCandidate);
+        } catch (MalformedURLException e) {
+            throw new WebDriverException(String.format("directConnect generated invalid URL as %s", newUrlCandidate));
```
mykola-mokhnach, line 272: same here

mykola-mokhnach: I would rather say:
```
The remote server XXX returned an invalid value used to build the direct connection URL: %s.
```

src/main/java/io/appium/java_client/AppiumClientConfig.java
KazuCocoa: I wonder if here should follow `ClientConfig`, that had no `set`:

e.g.
```
  public ClientConfig baseUrl(URL baseUrl) {
    try {
      return baseUri(Require.nonNull("Base URL", baseUrl).toURI());
    } catch (URISyntaxException e) {
      throw new RuntimeException(e);
    }
  }

  public URI baseUri() {
    return baseUri;
  }
```

Still not decided yet to inherit, or just keep the clientconfig as a variable so that users can define their subclass of clientconfig as their clientconfig.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: I haven't dug into netty client itself yet, but maybe we can find a way to replace url only for netty client, but not for client defined by user via factory,
https://github.com/SeleniumHQ/selenium/blob/9656da70e574a77213cba49ff00e11a2d0f1f393/java/src/org/openqa/selenium/remote/http/netty/NettyClient.java
https://github.com/SeleniumHQ/selenium/blob/9656da70e574a77213cba49ff00e11a2d0f1f393/java/src/org/openqa/selenium/remote/http/netty/NettyHttpHandler.java#L56

or maybe replacing ClientConfig instance might work
https://github.com/SeleniumHQ/selenium/blob/9656da70e574a77213cba49ff00e11a2d0f1f393/java/src/org/openqa/selenium/remote/http/RemoteCall.java

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: Btw, how can we inherit private constructor? (ClientConfig)
https://github.com/SeleniumHQ/selenium/blob/9b012df78dbbfd84d8fef06308f40fbf6a29455b/java/src/org/openqa/selenium/remote/http/ClientConfig.java#L42

Maybe I forgot something in Java, or probably haven't tried to extend private constructor's class in my past...

valfirst: I'm afraid it's not possible in general OOP paradigm, reflection might be overhead here, it seems only composition can be used

mykola-mokhnach: I believe we can ask Selenium fellows to help us making it public. cc @titusfortner @diemol

titusfortner: I'm not sure why this design pattern was chosen for this class. Make a PR to change the constructor to protected and we'll get it released.

KazuCocoa: Thanks, I have created https://github.com/SeleniumHQ/selenium/pull/10985

KazuCocoa:
I'll wait for the constructor change. Then, continue updating this pr

src/main/java/io/appium/java_client/gecko/GeckoDriver.java
KazuCocoa: Yes, the clientConfig version has not been released yet, so I modified it. But this will stay later.

KazuCocoa:
still working on this, keeping  draft

gradle.properties
```diff
@@ -7,4 +7,6 @@ signing.secretKeyRingFile=PathToYourKeyRingFile
 ossrhUsername=your-jira-id
 ossrhPassword=your-jira-password
 
-selenium.version=4.4.0
+selenium.version=4.5.0
+# Please increment the value in a release
+appiumClient.version=8.2.0
```
KazuCocoa, line 12: cc @SrinivasanTarget 
I've moved this appiumClient version definition here to read the value for User Agent (in src/main/java/io/appium/java_client/AddAppiumUserAgent.java). You will need to increment here in a release

KazuCocoa (commented):
updated

KazuCocoa:
iOS tests in the CI was passed before the commit ([c00697d](https://github.com/appium/java-client/pull/1747/commits/c00697d747273b3341b8bd6ed97dfc070eac4843)) https://dev.azure.com/srinivasansekar/java-client/_build/results?buildId=1417&view=results , so it is kind of CI env flakiness.
Not sure the conventional commits's result, but I guess this was because some of commit did not follow the rule..

SrinivasanTarget (approved):
LGTM

.github/workflows/gradle.yml
```diff
@@ -27,4 +27,4 @@ jobs:
         java-version: ${{ matrix.java }}
         cache: 'gradle'
     - name: Build with Gradle
-      run: ./gradlew clean build -x test -x checkstyleTest
+      run: ./gradlew clean build -x unitTest -x checkstyleTest
```
mykola-mokhnach, line 30: why do we need this change?

src/main/java/io/appium/java_client/AddAppiumUserAgent.java
```diff
@@ -0,0 +1,57 @@
...
+import org.openqa.selenium.remote.http.Filter;
+import org.openqa.selenium.remote.http.HttpHandler;
+
+import java.util.Locale;
+
+/**
+ * Manage Appium Client configurations.
+ */
+
+public class AddAppiumUserAgent implements Filter {
```
mykola-mokhnach, line 32: it would make sense to extract this functionality into a separate PR. It has nothing to do with direct connect stuff.

Also, I would rather use nouns for class names, for example `UserAgentAdder`

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    /**
+     * Return the instance of {@link AppiumClientConfig} with a default config.
+     * @return the instance of {@link AppiumClientConfig}.
+     */
+    public static AppiumClientConfig defaultConfig() {
+        return new AppiumClientConfig(
+                null,
+                Duration.ofSeconds(10),
```
mykola-mokhnach, line 78: magic value?

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,240 @@
...
+                clientConfig.baseUri(),
+                clientConfig.connectionTimeout(),
+                clientConfig.readTimeout(),
+                clientConfig.filter(),
+                clientConfig.proxy(),
+                clientConfig.credentials(),
+                directConnect);
+    }
+
+
```
mykola-mokhnach, line 210: extra line break

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,240 @@
...
+            URI baseUri,
+            Duration connectionTimeout,
+            Duration readTimeout,
+            Filter filters,
+            Proxy proxy,
+            Credentials credentials,
+            boolean directConnect) {
+        super(baseUri, connectionTimeout, readTimeout, filters, proxy, credentials);
+
+        this.directConnect = Require.nonNull("Direct Connect", directConnect);
```
mykola-mokhnach, line 68: `boolean` values cannot ever be null

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,240 @@
...
+                clientConfig.proxy(),
+                clientConfig.credentials(),
+                directConnect);
+    }
+
+    @Override
+    public AppiumClientConfig withFilter(Filter filter) {
+        // ClientConfig returns a new instance
+        ClientConfig clientConfig = super.withFilter(filter);
+        return new AppiumClientConfig(
```
mykola-mokhnach, line 157: I see some code duplication here. Can we have a helper method to reduce it?

.github/workflows/gradle.yml
KazuCocoa: Because... `test` found zero tests. I haven't dug deeply in this gradle's filtering test logic, but just kicked `test` and reported zero tests found. So tests under `io.appium.java_client.internal` (should run as unit test) haven't kicked via this test command.

KazuCocoa:
Updated. https://github.com/appium/java-client/pull/1779 includes UA stuff as another PR

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -38,67 +39,132 @@
...
 
-    private AppiumCommandExecutor(Map<String, CommandInfo> additionalCommands, DriverService service,
-                                  URL addressOfRemoteServer,
-                                  HttpClient.Factory httpClientFactory,
-                                  ClientConfig clientConfig) {
+    private final HttpClient.Factory httpClientFactory;
+
+    private final AppiumClientConfig appiumClientConfig;
+
+    private static class DirectConnect {
```
mykola-mokhnach, line 68: I would say it is ok to have this class public. Also, you could use [lombok](https://projectlombok.org/) for this purpose, which greatly reduces the amount of bilerplate code for POJOs

mykola-mokhnach: also having it public simplifies unit testing

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +246,43 @@ private Response createSession(Command command) throws IOException {
...
+
+        DirectConnect directConnect = new DirectConnect(responseValue);
+
+        if (!directConnect.isValid()) {
+            return;
+        }
+
+        if (!directConnect.protocol.equals("https")) {
+            throw new SessionNotCreatedException(
+                    String.format("The protocol must be https. %s was given.", directConnect.protocol));
```
mykola-mokhnach, line 273: I think the error message could be improved, for example:

The '%s' protocol is the direct connection url %s returned by the remote server is not acceptable. The only supported protocol is 'https'

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -166,13 +246,43 @@ private Response createSession(Command command) throws IOException {
...
+        }
+
+        URL newUrl;
+        try {
+            newUrl = directConnect.getUrl();
+        } catch (MalformedURLException e) {
+            throw new SessionNotCreatedException(e.getMessage());
```
mykola-mokhnach, line 280: same as above. We need to be quite specific about the source of the error

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -38,67 +39,132 @@
...
+        }
+
+        @Nullable
+        private String getDirectConnectValue(Map<String, ?> responseValue, String key) {
+            String directConnectPath = String.valueOf(responseValue.get(APPIUM_PREFIX + key));
+            return directConnectPath == null ? String.valueOf(responseValue.get(key)) : directConnectPath;
+        }
+
+        private boolean isValid() {
+            return !(this.protocol == null || this.path == null || this.host == null || this.port == null);
```
mykola-mokhnach, line 93: could also be `!Stream.of(protocol, ...).anyMatch((x) => x == null)`

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
KazuCocoa: Moved to a new public class file. I'll try the lambok later as another pr after the UA change (So far, I've left a TODO comment)

src/main/java/io/appium/java_client/remote/DirectConnect.java
```diff
@@ -0,0 +1,73 @@
...
+import static io.appium.java_client.internal.CapabilityHelpers.APPIUM_PREFIX;
+
+// TODO: simplify with lombok as another PR
+public class DirectConnect {
+    private static final String DIRECT_CONNECT_PROTOCOL = "directConnectProtocol";
+    private static final String DIRECT_CONNECT_PATH = "directConnectPath";
+    private static final String DIRECT_CONNECT_HOST = "directConnectHost";
+    private static final String DIRECT_CONNECT_PORT = "directConnectPort";
+
+    public final String protocol;
```
mykola-mokhnach, line 19: please create public accessors rather than making properties public (e.g. protocol -> getProtocol()). Later we could simplify that code with lombok

src/test/java/io/appium/java_client/internal/DirectConnectTest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.net.MalformedURLException;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class DirectConnectTest {
+
+    @Test
+    void hasValidDirectConnectValuesWithoutAppiumPrefix() throws MalformedURLException {
```
mykola-mokhnach, line 15: 👍

src/test/java/io/appium/java_client/internal/DirectConnectTest.java
```diff
@@ -0,0 +1,56 @@
...
+        responseValue.put("appium:directConnectPath", "/path/to");
+        responseValue.put("appium:directConnectHost", "host");
+        responseValue.put("appium:directConnectPort", "port");
+        DirectConnect directConnect = new DirectConnect(responseValue);
+        assertTrue(directConnect.isValid());
+        assertThrowsExactly(MalformedURLException.class, directConnect::getUrl);
+    }
+
+    @Test
+    void hasValidDirectConnectInvalid() {
```
mykola-mokhnach, line 51: this test name does not make much sense to me

mykola-mokhnach (approved):

SrinivasanTarget (approved):

src/main/java/io/appium/java_client/AppiumClientConfig.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Override
+    public AppiumClientConfig connectionTimeout(Duration timeout) {
+        ClientConfig clientConfig = super.connectionTimeout(timeout);
+        return buildAppiumClientConfig(clientConfig, directConnect);
+    }
+
+    @Override
+    public AppiumClientConfig readTimeout(Duration timeout) {
+        ClientConfig clientConfig = super.connectionTimeout(timeout);
```
tksjh, line 133: you should call super.readTimeout(timeout)

KazuCocoa: Thank you, fixed in https://github.com/appium/java-client/pull/1959

PR #1753, opened 2022-08-27
test: Verified disabled tests working and enabled them

sceiler:
- Enable Android tests that were disabled and verified them working locally on Emulator 28
- Rename AndroidConnectionTest methods to have more descriptive names and disable the ones relying on #17422
- Refactor tests to work

## Change list

Enabling Android tests and made slight changes where applicable to make tests run successfully.
 
## Types of changes

## Details

As part of the Open Source Test Automation Summit we went through disabled Android tests and verified that disabled tests work on a local Emulator with the same API level as used by the CI/CD server. The tests where then enabled and failing ones were fixed and/or got a comment that indicates that enabling them would depend on an issue getting fixed first.


mykola-mokhnach:
It looks like some tests still need tuning for the CI.

sceiler:
Yep I'm going through it asap. The pullFile test fails for example because that file doesn't exist on the Emulator. Probably less flaky if we create the file on the Emulator and then pull it instead of assuming that that file already exist.

PR #1787, opened 2022-10-28
fix: Override getScreenshotAs to support the legacy base64 encoding

mykola-mokhnach:
## Change list

Addresses https://github.com/SeleniumHQ/selenium/issues/11168
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -45,15 +45,9 @@ import org.openqa.selenium.remote.html5.RemoteLocationContext;
 import org.openqa.selenium.remote.http.HttpClient;
 import org.openqa.selenium.remote.http.HttpMethod;
 
-import java.io.File;
-import java.io.IOException;
 import java.lang.reflect.Field;
 import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.Arrays;
-import java.util.Base64;
 import java.util.Collections;
 import java.util.Map;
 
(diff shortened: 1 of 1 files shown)


SrinivasanTarget (approved):

src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -270,4 +277,48 @@ public Response execute(String driverCommand, Map<String, ?> parameters) {
...
+        if (outputType == OutputType.BYTES) {
+            return (X) screenshotRawBytes;
+        } else if (outputType == OutputType.BASE64) {
+            //noinspection unchecked
+            return (X) Base64.getEncoder().encodeToString(screenshotRawBytes);
+        } else if (outputType == OutputType.FILE) {
+            //noinspection unchecked
+            return (X) saveToFile(screenshotRawBytes);
+        }
+        throw new IllegalArgumentException(String.format("Unsupported %s instance", OutputType.class.getName()));
```
valfirst, line 322: what about making it simpler?:

```suggestion
        return super.getScreenshotAs(new OutputType<>() {
            @Override
            public X convertFromBase64Png(String base64Png) {
            // TODO: Eventually we should not override this method.
            // TODO: Although, we have a legacy burden,
            // TODO: so it's impossible to do it the other way as of Oct 29 2022.
            // TODO: See https://github.com/SeleniumHQ/selenium/issues/11168
                return outputType.convertFromBase64Png(base64Png.replaceAll("\r?\n", ""));
            }

            @Override
            public X convertFromPngBytes(byte[] png) {
                return outputType.convertFromPngBytes(png);
            }
        });
```

mykola-mokhnach: good idea. Changed

valfirst (approved):

krishtoautomate:
Possible to create maven build of appium java client with this patch?

sergey-lysenko:
This only fixed fullscreen screentshots, while calling method of WebElement falls back to default implementation and fails the test.

I guess the only option now (untill overriding is the only option) is to create a routine in tests which will make a full screenshot, then crop needed part based on WebElement's dimentions.

mykola-mokhnach:
> This only fixed fullscreen screentshots, while calling method of WebElement falls back to default implementation and fails the test.
> 
> I guess the only option now (untill overriding is the only option) is to create a routine in tests which will make a full screenshot, then crop needed part based on WebElement's dimentions.

This patch is ugly and it anyway would make sense to apply it on Selenium side instead. Or simply use the newest version of an appium driver, where it generates a compatible base64-encoding.

PR #1790, opened 2022-11-02
feat: Add alternative proxy implementation

mykola-mokhnach:
## Change list

Basically same as https://github.com/SeleniumHQ/selenium/pull/11201
 
## Types of changes


src/main/java/io/appium/java_client/proxy/Interceptor.java
```diff
@@ -0,0 +1,123 @@
...
+
+import java.lang.reflect.Method;
+import java.util.Collection;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.WeakHashMap;
+import java.util.concurrent.Callable;
+import java.util.logging.Level;
+import java.util.logging.Logger;
```
valfirst, line 33: since it's not a part of Selenium, what about using SLF4J (it's already in the dependencies): `org.slf4j.Logger` and `org.slf4j.LoggerFactory` ?

mykola-mokhnach: changed

valfirst (approved):

PR #1791, opened 2022-11-03
chore: remove unnecessary annotation

KazuCocoa:
## Change list

The accessors annotation can be removed when it has no arguments.
```
> Task :compileJava
/Users/kazu/GitHub/java-client/src/main/java/io/appium/java_client/remote/DirectConnect.java:16: warning: Accessors on its own does nothing. Set at least one parameter
@Accessors
^
```
 
## Types of changes

## Details

It seems like the annotation is used only when we'd like to set non-default options
https://github.com/projectlombok/lombok/blob/560ded50127b12bd8168f8ae2bada47253f15bc8/src/core/lombok/core/AnnotationValues.java#L570

(i have tested without this annotation)

### DIFF
--- a/src/main/java/io/appium/java_client/remote/DirectConnect.java
+++ b/src/main/java/io/appium/java_client/remote/DirectConnect.java
@@ -13,7 +13,6 @@ import java.util.stream.Stream;
 
 import static io.appium.java_client.internal.CapabilityHelpers.APPIUM_PREFIX;
 
-@Accessors
 public class DirectConnect {
     private static final String DIRECT_CONNECT_PROTOCOL = "directConnectProtocol";
     private static final String DIRECT_CONNECT_PATH = "directConnectPath";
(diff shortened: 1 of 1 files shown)


valfirst (approved):

src/main/java/io/appium/java_client/remote/DirectConnect.java
```diff
@@ -13,7 +29,6 @@
 
 import static io.appium.java_client.internal.CapabilityHelpers.APPIUM_PREFIX;
 
-@Accessors
```
mykola-mokhnach, line 16: Is the appropriate import sill used?

mykola-mokhnach (approved):

src/main/java/io/appium/java_client/remote/DirectConnect.java
SrinivasanTarget: Nope its unused, it should be removed.

PR #1793, opened 2022-11-04
chore: Force unified imports order

### LINKED ISSUE #1722: Code Styles not in sync
## Description

Throughout the project Code Formatter is not in sync

## Details

Please provide more details, if necessary.

## Code To Reproduce Issue [ Good To Have ]

The below class uses static imports in first place
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/android/AndroidDriver.java

However in the below class ,static imports are used in last place
https://github.com/appium/java-client/blob/master/src/main/java/io/appium/java_client/ExecuteCDPCommand.java

## Exception Stacktraces
N/A

## Link To Appium Logs
N/A

valfirst:
## Change list

Closes #1722: 
- Use unified imports order for all classes
- Configure checkstyle rules for imports
 
## Types of changes

## Details
The imports order reflects the default order from IntellJ IDEA (so that developers don't need to configure their IDEs) except for star-imports, they should be avoided.


SrinivasanTarget (approved):

google-style.xml
```diff
@@ -41,9 +41,16 @@
             <property name="allowByTailComment" value="true"/>
             <property name="allowNonPrintableEscapes" value="true"/>
         </module>
-        <module name="AvoidStarImport">
-            <property name="severity" value="warning"/>
-        </module>
+        <module name="AvoidStarImport"/>
```
mykola-mokhnach, line 44: is it still the same set of rules that Google uses?
does it create warnings for unused/star imports?

valfirst: > is it still the same set of rules that Google uses?

no, Google relies on the very first version of this config

> does it create warnings for unused/star imports?

it produces violations for all star(*) imports, unused and redundant imports are checked by 2 modules below:
```xml
<module name="RedundantImport"/>
<module name="UnusedImports"/>
```

mykola-mokhnach: then we probably should not have `google` in its name as it is now a custom one

mykola-mokhnach (approved):

valfirst:
> it would be cool we we could reference the original Google style from a remote repository and override it locally, but I assume it is not possible

yes, unfortunately that's not possible

PR #1813, opened 2022-12-03
feat: Add a possibility to connect to a running session

mykola-mokhnach:
## Change list

I am still not quite sure what the purpose of such feature, but people have been asking for it. 
The main issue is that Selenium API does not have a concept of session state, so we cannot reliably share it across multiple client instances. But we still can send commands, which is probably enough for debugging and other weird purposes.

The concept is simple, - all drivers got an additional public constructor, which accepts the remote session url (e.g. server url + "/session/id" prefix) and platform/automation name where necessary. The created "generic" driver only knows about the actual session id. Otherwise it is always assumed it is a valid Appium session using the W3C protocol.

## Types of changes


KazuCocoa (approved):
seems reasonable to me to set platfornName and automationName as no definition in w3c to get current session info

SrinivasanTarget (approved):

valfirst (approved):

PR #1849, opened 2023-01-25
perf: Improve performance of proxy `Interceptor` logging

valfirst:
## Change list

Improve performance of proxy `Interceptor` logging
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/proxy/Interceptor.java
+++ b/src/main/java/io/appium/java_client/proxy/Interceptor.java
@@ -72,10 +72,10 @@ public class Interceptor {
             } catch (NotImplementedException e) {
                 // ignore
             } catch (Exception e) {
-                logger.error(
-                        String.format("Got an unexpected error in beforeCall listener of %s.%s method",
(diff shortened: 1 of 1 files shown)


SrinivasanTarget (approved):

mykola-mokhnach:
Could you please add some links to actual performance comparisons/documentation?

valfirst:
That's a recommendation from SLF4J: https://www.slf4j.org/faq.html#logging_performance

mykola-mokhnach:
> That's a recommendation from SLF4J: https://www.slf4j.org/faq.html#logging_performance

If I read it right the tutorial above advices something like `logger.debug("The new entry is {}.", entry);`

valfirst:
> If I read it right the tutorial above advices something like logger.debug("The new entry is {}.", entry);

Right, that's a major improvement: replace `String.format` with parametrised message.

Also PR uses fluent logging approach (https://www.slf4j.org/manual.html#fluent), it allows to calculate logged values lazily (only if logging for the particular level is enabled).

Issue #1872, opened 2023-03-26
Java 8.3.0 is not compatible with selenium 4.8.2 


Idoserovitz:
## Description

selenium java version [4.8.2](https://github.com/SeleniumHQ/selenium/blob/trunk/java/CHANGELOG) was released no long ago and appium java-client 8.3.0 cannot invoke appium server programmatically(manually it works well). I get `SurefireExecutionException: Exception in provider` error when executing the test and I can see in the logs that appium server wasn’t able to launch.

## Environment

* Java client build version or git revision if you use some snapshot: 8.3.0
* Appium server version or git revision if you use some snapshot: 2.0.0-beta.57
* Desktop OS/version used to run Appium if necessary: MacOS Ventura, m1 machine
* Node.js version (unless using Appium.app|exe) or Appium CLI or Appium.app|exe: 16.16.0
* Mobile platform/version under test: android api 30, iOS 15.5
* Real device or emulator/simulator: emulator and simulator on both platforms

## Details

I use maven, testng and surefire-plugin setup. Maven automatically updates selenium's version.

## Exception Stacktraces
SurefireExecutionException: Exception in provider

Idoserovitz:
Solved it by going to `$HOME/.m2/repository/io/appium/java-client/8.3.0` and adjusting the file: `java-client-8.3.0.pom`.
Changing all selenium related dependencies' version tag from: `<version>[4.7.0, 5.0)</version>` to `<version>[4.7.0, 4.8.2)</version>` to eliminate incompatible 4.8.2 selenium version.

Still would like to have a time estimation for a compatible version, Thanks.

mykola-mokhnach:
Please provide the full stacktrace

Idoserovitz:
@mykola-mokhnach 
https://gist.github.com/Idoserovitz/58c2b2e41ede8c950d7ffb4bc3eeaffd

Note that "CustomTestListenerAdapter" is a class I created which implements test failure post action and triggered by testng. It should not cause such an issue.
And again, appium server never started.

mykola-mokhnach:
I don't observe any mentions of appium client in the stacktrace. Are you sure your issue is somehow related to it?

amnajamshaid:
> I've just merged #1878
> 
> It will be published with the next java client patch

How long would it take? Would it be deployed till tomorrow?

amnajamshaid:
[...]

PR #1880, opened 2023-04-03
chore: Bump the minimum selenium client version to 4.8.2

mykola-mokhnach:
## Change list

Due to the [breaking change](https://github.com/appium/java-client/issues/1872#issuecomment-1494233769) in the recent Selenium client release we need to update the service builder class to make it great again.
 
## Types of changes

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.7.0
+selenium.version=4.8.2
 # Please increment the value in a release
 appiumClient.version=8.3.0
--- a/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
@@ -406,7 +406,7 @@ public final class AppiumServiceBuilder
     @Override
     public AppiumDriverLocalService build() {
         File driverExecutable = ReflectionHelpers.getPrivateFieldValue(
-                AppiumServiceBuilder.class, this, "exe", File.class
+                DriverService.Builder.class, this, "exe", File.class
         );
         if (driverExecutable == null) {
             usingDriverExecutable(findDefaultExecutable());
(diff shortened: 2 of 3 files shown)


src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -401,6 +403,17 @@ protected ImmutableList<String> createArgs() {
         return new ImmutableList.Builder<String>().addAll(argList).build();
     }
 
+    @Override
+    public AppiumDriverLocalService build() {
+        File driverExecutable = ReflectionHelpers.getPrivateFieldValue(
+                AppiumServiceBuilder.class, this, "exe", File.class
```
nir-tal-talkspace, line 409: will this work on mac/Linux also?

valfirst (approved):

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
mykola-mokhnach: yes, why shouldn't it?

gradle.properties
```diff
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.7.0
+selenium.version=4.8.2
```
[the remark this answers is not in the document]
mykola-mokhnach: 4.8.2 is the lowest range limit. The client accepts [4.8.2, 5.0)

valfirst (changes requested):
the tests are failing

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
valfirst: `AppiumServiceBuilder.class` -> `org.openqa.selenium.remote.service.DriverService.Builder.class` - it seems this should fix the access since the field `exe` is private

mykola-mokhnach: already done ;)

valfirst (approved):

debasisj:
has this been released ? also given the latest selenum version is 4.8.3- assume the  fix will work for this as well?

PR #1883, opened 2023-04-13
refactor: Replace non-W3C API  calls with corresponding extension calls in app management

mykola-mokhnach:
## Change list

Eventually we should avoid creating wrappers for non-W3C-compatible endpoints as these endpoints are mostly legacy from mobile-specific commands. We should only provide `mobile:` extension for these.

I don't want to remove existing wrappers though as this could create extra headaches for clients. These should work same way as before, but call corresponding extensions under the hood.

I would say it makes sense to introduce similar changes to other clients as well (e.g. Python/Dotnet/Ruby)

cc @KazuCocoa @jlipps @dor-bl

### DIFF
--- a/src/main/java/io/appium/java_client/InteractsWithApps.java
+++ b/src/main/java/io/appium/java_client/InteractsWithApps.java
@@ -20,21 +20,17 @@ import com.google.common.collect.ImmutableMap;
 import io.appium.java_client.appmanagement.ApplicationState;
 import io.appium.java_client.appmanagement.BaseActivateApplicationOptions;
 import io.appium.java_client.appmanagement.BaseInstallApplicationOptions;
+import io.appium.java_client.appmanagement.BaseOptions;
 import io.appium.java_client.appmanagement.BaseRemoveApplicationOptions;
 import io.appium.java_client.appmanagement.BaseTerminateApplicationOptions;
 
 import javax.annotation.Nullable;
 import java.time.Duration;
-import java.util.AbstractMap;
+import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


valfirst (approved):

KazuCocoa (approved):

src/main/java/io/appium/java_client/InteractsWithApps.java
```diff
@@ -171,11 +170,10 @@ default boolean terminateApp(String bundleId) {
...
-                new String[]{"bundleId", "options"};
-        Object[] values = options == null ? new Object[]{bundleId} :
-                new Object[]{bundleId, options.build()};
-        return CommandExecutionHelper.execute(this,
-                new AbstractMap.SimpleEntry<>(TERMINATE_APP, prepareArguments(parameters, values)));
+        Map<String, Object> args = new HashMap<>();
+        args.put("bundleId", bundleId);
+        args.put("appId", bundleId);
+        Optional.ofNullable(options).map(BaseOptions::build).ifPresent(args::putAll);
+        return CommandExecutionHelper.executeScript(this, "mobile: terminateApp", args);
```
KazuCocoa, line 177: btw, does this include for mac2 driver as well?  Then, mac2 driver needs `macos` instead of `mobile`
https://github.com/appium/appium-mac2-driver#macos-activateapp

mykola-mokhnach: no, only AndroidDriver and IOSDriver implement this interface

PR #1892, opened 2023-04-21
refactor: Finish replacing iOS extensions with their mobile alternatives

mykola-mokhnach:


src/main/java/io/appium/java_client/LocksDevice.java
```diff
@@ -41,15 +45,31 @@ default void lockDevice() {
...
 
     /**
      * Unlock the device if it is locked. This method will return silently if the device
      * is not locked.
      */
     default void unlockDevice() {
-        CommandExecutionHelper.execute(this, unlockDeviceCommand());
+        try {
+            //noinspection ConstantConditions
+            if (!(Boolean) CommandExecutionHelper.executeScript(this, "mobile: isLocked")) {
```
valfirst, line 65:
```suggestion
            if (!isDeviceLocked()) {
```

valfirst: got it, thanks!

valfirst (approved):

PR #1908, opened 2023-05-09
chore: Bump minimum Selenium client version to 4.9.1

valfirst:
Selenium 4.9.1 introduced 2 breaking changes that cause compilation errors:
- https://github.com/SeleniumHQ/selenium/commit/66e51be38cbd492c11605397e1b50c1cb786a579
- https://github.com/SeleniumHQ/selenium/commit/3d8c6fa7a950ce129f4bd8fb347c71a2cdd4bf07

## Change list

Make Appium Java client compatible with the newest Selenium client
 
## Types of changes

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.8.2
+selenium.version=4.9.1
 # Please increment the value in a release
 appiumClient.version=8.4.0
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -290,8 +290,8 @@ public class AppiumDriver extends RemoteWebDriver implements
         }
 
         @SuppressWarnings("unchecked") Map<String, Object> rawCapabilities = (Map<String, Object>) responseValue;
-        // A workaround for Selenium API enforcing some legacy capability values
-        rawCapabilities.remove(CapabilityType.PLATFORM);
+        // TODO: remove this workaround for Selenium API enforcing some legacy capability values in major version
+        rawCapabilities.remove("platform");
         if (rawCapabilities.containsKey(CapabilityType.BROWSER_NAME)
                 && isBlank((String) rawCapabilities.get(CapabilityType.BROWSER_NAME))) {
             rawCapabilities.remove(CapabilityType.BROWSER_NAME);
--- a/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
+++ b/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
@@ -404,14 +404,13 @@ public final class AppiumServiceBuilder
     }
 
     @Override
-    public AppiumDriverLocalService build() {
+    protected void loadSystemProperties() {
         File driverExecutable = ReflectionHelpers.getPrivateFieldValue(
                 DriverService.Builder.class, this, "exe", File.class
         );
         if (driverExecutable == null) {
             usingDriverExecutable(findDefaultExecutable());
         }
-        return super.build();
     }
 
     /**


src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -291,7 +291,6 @@ protected void startSession(Capabilities capabilities) {
 
         @SuppressWarnings("unchecked") Map<String, Object> rawCapabilities = (Map<String, Object>) responseValue;
         // A workaround for Selenium API enforcing some legacy capability values
-        rawCapabilities.remove(CapabilityType.PLATFORM);
```
mykola-mokhnach, line 294: is the workaround not needed anymore?

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
```diff
@@ -404,14 +404,13 @@ protected ImmutableList<String> createArgs() {
     }
 
     @Override
-    public AppiumDriverLocalService build() {
```
mykola-mokhnach, line 407: was the `build` method removed?

gradle.properties
```diff
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.8.2
+selenium.version=4.9.1
```
mykola-mokhnach, line 3: I assume this would also be compatible to 4.9.0
We put a minimum compatible version here

src/main/java/io/appium/java_client/AppiumDriver.java
valfirst: `CapabilityType.PLATFORM` is removed from Selenium, so I guess Selenium doesn't enforce any capability named "platform"

gradle.properties
valfirst: `loadSystemProperties` is introduced in `4.9.1`.

actually we can do a trick, it's possible to make changes in the way to keep Appium client compatible with 4.8.2, 4.8.3, 4.9.0, 4.9.1: no `@Override` should be put on `loadSystemProperties`, that's a bit hacky way, does it worth?

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
mykola-mokhnach: I would say it is ok to have loadSystemProperties.
Although I assume the "original" build method returns is typed by the class, which is probably not what we want here. That is why it still makes sense to override it in order to better reflect the returned instance type

gradle.properties
mykola-mokhnach: no, better not to. This might open new ways to f#^%k it up.

src/main/java/io/appium/java_client/AppiumDriver.java
mykola-mokhnach: also do not forget to remove the obsolete comment above after you verify the capability is not there anymore

src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java
valfirst: > Although I assume the "original" build method return is typed by the class, which is probably not what we want here (the parent one). That is why it still makes sense to override it in order to better reflect the returned instance type

Actually it uses generics and the returned type is specified by Builder declaration:
https://github.com/appium/java-client/blob/f2fb595a81cc8cfeca22fd7ff80847f277c2a2a9/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java#L58-L59
so it works perfectly for Appium and the result of `build` method is of `AppiumDriverLocalService` type.

mykola-mokhnach (approved):

kuldeep04:
@valfirst `The following artifacts could not be resolved: io.appium:java-client:jar:9c18b2f (absent): Could not find artifact io.appium:java-client:jar:9c18b2f in release-http-unblocker`

above error is showing with the latest commit.

```
        <dependency>
            <groupId>io.appium</groupId>
            <artifactId>java-client</artifactId>
            <version>9c18b2f</version>
        </dependency>
```

PR #1958, opened 2023-07-11
Fix for AUTO_WEBVIEW in MobileCapabilityType

FiniH:
Changed value for AUTO_WEBVIEW from 'autoWebview' to 'autoWebView'.

## Change list

Please provide briefly described change list which are you going to propose. 

- Changing the value for MobileCapabilityType.AUTO_WEBVIEW from 'autoWebview' to 'autoWebView'.
 
## Types of changes

## Details

I realized while using the MobileCapabilityType.AUTO_WEBVIEW that the value for the type is 'autoWebview' and not 'autoWebView'. The latter seems to be the correct value though as my app test actually crashed on me until I used the 'autoWebView' string instead of the type itself. This is why I propose this change to be in the project itself to I can use the type in the future and not depend on the 'autoWebView' capability using the string anymore.

### DIFF
--- a/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
+++ b/src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
@@ -85,7 +85,7 @@ public interface MobileCapabilityType extends CapabilityType {
     /**
      * Move directly into Webview context. Default false.
      */
-    String AUTO_WEBVIEW = "autoWebview";
+    String AUTO_WEBVIEW = "autoWebView";
 
     /**
      * Don't reset app state before this session. See


src/main/java/io/appium/java_client/remote/MobileCapabilityType.java
```diff
@@ -85,7 +85,7 @@ public interface MobileCapabilityType extends CapabilityType {
     /**
      * Move directly into Webview context. Default false.
      */
-    String AUTO_WEBVIEW = "autoWebview";
+    String AUTO_WEBVIEW = "autoWebView";
```
valfirst, line 88: It doesn't seem to be a valid fix

mykola-mokhnach:
We should probably deprecate this whole interface for good.
One should use options in order to create a new session

mykola-mokhnach:
The class has been marked for deprecation

PR #1969, opened 2023-07-28
Caching Mechanism for ProxyListenerContainer

Auto81:
## Change list

Updating the ProxyListenerContainer to use a LinkedHashMap caching mechanism, in place of a List. This is to improve performance as the amount of listeners grows.

I have defaulted the cache to 5000 decorated pages, but this can be changed if needed through a system property
 
## Types of changes

## Details

Original [Discussion](https://github.com/appium/appium/discussions/18860)

### DIFF
--- a/src/main/java/io/appium/java_client/proxy/ProxyListenersContainer.java
+++ b/src/main/java/io/appium/java_client/proxy/ProxyListenersContainer.java
@@ -18,10 +18,9 @@ package io.appium.java_client.proxy;
 
 import java.lang.ref.WeakReference;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.LinkedHashMap;
+import java.util.Map;
 import java.util.concurrent.Semaphore;
 
 class ProxyListenersContainer {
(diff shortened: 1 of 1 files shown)


src/main/java/io/appium/java_client/proxy/ProxyListenersContainer.java
```diff
@@ -39,25 +37,17 @@ public static synchronized ProxyListenersContainer getInstance() {
...
-            return key;
-        }
-
-        public V getValue() {
-            return value;
-        }
-    }
+    // As of (this version?) we are now using a LinkedHashMap as a caching mechanism. This can be overridden by a System
+    // Property value "appium.override.listener.limit": default 5000.
+    private final LinkedHashMap<Object, HashSet<MethodCallListener>>
```
mykola-mokhnach, line 42: I am not sure we want to have LRU-like structure here. We don't want older listeners to be dropped based on the amount of newer ones. We only want them to be dropped if the amount of references to them is zero (e.g. ready for garbage collect)

src/main/java/io/appium/java_client/proxy/ProxyListenersContainer.java
```diff
@@ -76,27 +66,7 @@ public <T> T setListeners(T proxyInstance, Collection<MethodCallListener> listen
...
-                    pair.getValue().clear();
-                    pair.getValue().addAll(listeners);
-                    wasInstancePresent = true;
-                }
-                i++;
-            }
-            if (!wasInstancePresent) {
-                this.listeners.add(new Pair<>(new WeakReference<>(proxyInstance), new HashSet<>(listeners)));
-            }
+            this.listeners.put(String.valueOf(proxyInstance.getClass()), new HashSet<>(listeners));
```
mykola-mokhnach, line 69: I am not convinced that class itself could be the key here. listeners must be connected to instances

mykola-mokhnach: Also, why it is necessary to perform `String.valueOf` on the value? What would be the example output?

Auto81: > 

Removed String.valueOf

Auto81:
> @Auto81 Do you have any particular updates on this issue after checking the memory leak? Do you think it is still something that could be addressed from the java client perspective?

Have just updated the discussion at https://github.com/appium/appium/discussions/18860

PR #1981, opened 2023-08-15
fix: Use weak references to elements inside of interceptor objects

mykola-mokhnach:
## Change list

This is a quick and poorly tested approach to address https://github.com/appium/appium/discussions/18860#discussioncomment-6731006
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
@@ -155,8 +155,8 @@ class AppiumElementLocator implements CacheableLocator {
 
         SearchContext searchContext = getSearchContext()
                 .orElseThrow(() -> new IllegalStateException(
-                        String.format("The element %s is not locatable anymore " +
-                                "because its context has been garbage collected", by)
+                        String.format("The element %s is not locatable anymore "
+                                + "because its context has been garbage collected", by)
                 ));
 
         By bySearching = getBy(this.by, searchContext);
@@ -181,8 +181,8 @@ class AppiumElementLocator implements CacheableLocator {
 
         SearchContext searchContext = getSearchContext()
                 .orElseThrow(() -> new IllegalStateException(
-                    String.format("Elements %s are not locatable anymore " +
-                            "because their context has been garbage collected", by)
+                    String.format("Elements %s are not locatable anymore "
+                            + "because their context has been garbage collected", by)
                 ));
 
         List<WebElement> result;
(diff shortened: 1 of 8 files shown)


src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -46,10 +48,36 @@ class AppiumElementLocator implements CacheableLocator {
...
+     * using @AndroidFindBy (-s), @iOSFindBy (-s) and @FindBy (-s) annotation
+     * sets
+     *
+     * @param searchContextReference     The context reference to use when finding the element
+     * @param by                         a By locator strategy
+     * @param shouldCache                is the flag that signalizes that elements which
+     *                                   are found once should be cached
+     * @param duration                   timeout parameter for the element to be found
+     */
+    public AppiumElementLocator(
```
valfirst, line 68:
```suggestion
    AppiumElementLocator(
```
does it make sense to make this constructor package-private?

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
```diff
@@ -61,15 +89,29 @@ class AppiumElementLocator implements CacheableLocator {
...
     }
 
+    private Optional<SearchContext> getSearchContext() {
+        if (searchContext != null) {
+            return Optional.of(searchContext);
+        }
+        if (searchContextReference != null) {
+            return Optional.ofNullable(searchContextReference.get());
+        }
+        return Optional.empty();
```
valfirst, line 112:
```suggestion
        return Optional.ofNullable(searchContextReference).map(WeakReference::get);
```

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
valfirst: The same quesiton for all new API-s using `WeakReference`-s

valfirst (changes requested):

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
mykola-mokhnach: changed

src/main/java/io/appium/java_client/pagefactory/AppiumElementLocator.java
mykola-mokhnach: changed

valfirst (approved):

PR #2036, opened 2023-10-10
refactor!: Update selenium dependencies to v4.14

mykola-mokhnach:
BREAKING CHANGE: The library is now only compatible to Java11 and above
BREAKING CHANGE: The previously deprecated MobileBy class has been removed. Use AppiumBy instead
BREAKING CHANGE: The previously deprecated launchApp, resetApp and closeApp methods have been removed. Use extension methods instead
BREAKING CHANGE: The previously deprecated WindowsBy class and related location strategies have been removed
BREAKING CHANGE: The previously deprecated ByAll class has been removed in favour of the Selenium one
BREAKING CHANGE: The previously deprecated AndroidMobileCapabilityType interface has been removed. Use driver options instead
BREAKING CHANGE: The previously deprecated IOSMobileCapabilityType interface has been removed. Use driver options instead
BREAKING CHANGE: The previously deprecated MobileCapabilityType interface has been removed. Use driver options instead
BREAKING CHANGE: The previously deprecated MobileOptions class has been removed. Use driver options instead
BREAKING CHANGE: The previously deprecated YouiEngineCapabilityType interface has been removed. Use driver options instead
BREAKING CHANGE: Removed several misspelled methods. Use properly spelled alternatives instead
BREAKING CHANGE: Removed startActivity method from AndroidDriver. Use 'mobile: startActivity' extension method instead
BREAKING CHANGE: Removed the previously deprecated APPIUM constant from the AutomationName interface
BREAKING CHANGE: Removed the previously deprecated PRE_LAUNCH value from the GeneralServerFlag enum
BREAKING CHANGE: Moved AppiumUserAgentFilter class to io.appium.java_client.internal.filters package

Removed dependencies to Apache Commons libraries


SrinivasanTarget (approved):

gradle.properties
```diff
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.9.1
+selenium.version=4.14.0
 # Please increment the value in a release
 appiumClient.version=8.6.0
```
valfirst, line 5:
```suggestion
appiumClient.version=9.0.0-SNAPSHOT
```
does it make sense to reflect the major version incremenet?

mykola-mokhnach: no, this should only be changed as the part of the release procudure

mykola-mokhnach:
Let me check failing tests before merging

mykola-mokhnach:
Looks like most of the test errors are happening because of connection reset. I did not observe  these in previous client version, so I'd assume it might something to do with the HTTP client change in Selenium lib. 
SO says we could enforce HTTP version to 1.1 (which also makes sense because the server anyway does not support HTTP2), but there is no way to do tune that in selenium client for now. I'll ask selenium maintainers to add this feature, just not sure if we need to wait until it is released (which might take a while) or continue with what we have now?

cc @valfirst @saikrishna321 @asolntsev

saikrishna321:
Let's enforce to 1.1 and go forward.

asolntsev:
@mykola-mokhnach Yes, it seems you should force http 1.1

I've tried to run Selenide tests against your branch, and **most** of tests worked correctly, but **few** tests failed with errors like this:
> Response code 500. Message: An unknown server-side error occurred while processing the command. Original error: Could not proxy command to the remote server. Original error: socket hang up

.github/workflows/gradle.yml
```diff
@@ -27,14 +31,13 @@ jobs:
...
         include:
-        # TODO: add new LTS Java ( 21 ) once it's released
-          - java: 8
-            platform: macos-latest
-            e2e-tests: android
           - java: 11
             # Need to use specific (not `-latest`) version of macOS to be sure the required version of Xcode/simulator is available
             platform: macos-12
             e2e-tests: ios
+          - java: 15
```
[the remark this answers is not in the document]
mykola-mokhnach: probably it does not

shreyansh791:
The previously deprecated MobileCapabilityType interface has been removed. Use driver options instead

Could you please explain what is the meaning of driver options here?

valfirst:
@shreyansh791 please check migration guide: https://github.com/appium/java-client/blob/master/docs/v8-to-v9-migration-guide.md

PR #2098, opened 2024-01-14
chore: Add non-W3C Location-management endpoints deprecated in Selenium client

valfirst:
## Change list

Add non-W3C Location-management endpoints deprecated in Selenium client: https://github.com/SeleniumHQ/selenium/pull/13372/files#diff-1af793ae2090a0cb691f1efbb3ee0d2b186c4027c5b441f6f3d55b3663e67ba1
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/Location.java
+++ b/src/main/java/io/appium/java_client/Location.java
@@ -16,6 +16,7 @@
 
 package io.appium.java_client;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 
@@ -26,10 +27,11 @@ import javax.annotation.Nullable;
  */
 @Getter
 @ToString
+@EqualsAndHashCode
 public class Location {
     private final double latitude;
     private final double longitude;
-    private final Double altitude;
+    @Nullable private final Double altitude;
 
     /**
      * Create {@link Location} with latitude, longitude and altitude values.
--- a/src/main/java/io/appium/java_client/Location.java
+++ b/src/main/java/io/appium/java_client/Location.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client;
+
+import lombok.Getter;
+import lombok.ToString;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents the physical location.
+ */
+@Getter
+@ToString
+public class Location {
+    private final double latitude;
+    private final double longitude;
+    private final Double altitude;
+
+    /**
+     * Create {@link Location} with latitude, longitude and altitude values.
+     *
+     * @param latitude latitude value.
+     * @param longitude longitude value.
+     * @param altitude altitude value (can be null).
+     */
+    public Location(double latitude, double longitude, @Nullable Double altitude) {
+        this.latitude = latitude;
+        this.longitude = longitude;
+        this.altitude = altitude;
+    }
+
+    /**
+     * Create {@link Location} with latitude and longitude values.
+     *
+     * @param latitude latitude value.
+     * @param longitude longitude value.
+     */
+    public Location(double latitude, double longitude) {
+        this(latitude, longitude, null);
+    }
+}
--- a/src/main/java/io/appium/java_client/android/AndroidDriver.java
+++ b/src/main/java/io/appium/java_client/android/AndroidDriver.java
@@ -251,6 +251,14 @@ public class AndroidDriver extends AppiumDriver implements
         return new AndroidBatteryInfo(CommandExecutionHelper.executeScript(this, "mobile: batteryInfo"));
     }
 
+    /**
+     * Provides the location context.
+     *
+     * @return instance of {@link RemoteLocationContext}
+     * @deprecated This method, {@link org.openqa.selenium.html5.LocationContext} and {@link RemoteLocationContext}
+     *     interface are deprecated, use {@link #getLocation()} and
+     *     {@link #setLocation(io.appium.java_client.Location)} instead.
+     */
     @Override
     @Deprecated(forRemoval = true)
     public RemoteLocationContext getLocationContext() {
--- a/src/main/java/io/appium/java_client/android/AndroidDriver.java
+++ b/src/main/java/io/appium/java_client/android/AndroidDriver.java
@@ -252,6 +252,7 @@ public class AndroidDriver extends AppiumDriver implements
     }
 
     @Override
+    @Deprecated(forRemoval = true)
     public RemoteLocationContext getLocationContext() {
         return locationContext;
     }
--- a/src/main/java/io/appium/java_client/remote/SupportsLocation.java
+++ b/src/main/java/io/appium/java_client/remote/SupportsLocation.java
@@ -16,19 +16,70 @@
 
 package io.appium.java_client.remote;
 
+import io.appium.java_client.CommandExecutionHelper;
+import io.appium.java_client.ExecutesMethod;
+import io.appium.java_client.MobileCommand;
 import org.openqa.selenium.WebDriver;
+import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.html5.Location;
 import org.openqa.selenium.html5.LocationContext;
 import org.openqa.selenium.remote.html5.RemoteLocationContext;
 
-public interface SupportsLocation extends WebDriver, LocationContext {
-    public RemoteLocationContext getLocationContext();
+import java.util.Map;
+import java.util.Optional;
 
+public interface SupportsLocation extends WebDriver, ExecutesMethod, LocationContext {
+
+    @Deprecated(forRemoval = true)
+    RemoteLocationContext getLocationContext();
+
+    /**
+     * Gets the physical location of the browser.
+     *
+     * @return A {@link Location} containing the location information. Returns null if the location is not available
+     * @deprecated This method and whole {@link LocationContext} interface are deprecated, use {@link #getLocation()}
+     *     instead.
+     */
+    @Deprecated(forRemoval = true)
     default Location location() {
         return getLocationContext().location();
     }
 
+    /**
+     * Gets the physical location.
+     *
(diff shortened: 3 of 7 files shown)


SrinivasanTarget (approved):

src/main/java/io/appium/java_client/Location.java
```diff
@@ -0,0 +1,57 @@
...
+     * @param altitude altitude value.
+     */
+    public Location(double latitude, double longitude, double altitude) {
+        super(latitude, longitude, altitude);
+        this.latitude = latitude;
+        this.longitude = longitude;
+        this.altitude = altitude;
+    }
+
+    public double getLatitude() {
```
mykola-mokhnach, line 41: Consider using lombok for data classes creation

src/main/java/io/appium/java_client/Location.java
```diff
@@ -0,0 +1,57 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.appium.java_client;
+
+/**
+ * Represents the physical location.
+ */
+public class Location extends org.openqa.selenium.html5.Location {
```
mykola-mokhnach, line 22: would this data model remain in selenium?

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,54 @@
...
+
+    /**
+     * Gets the physical location.
+     *
+     * @return A {@link Location} containing the location information. Returns null if the location is not available.
+     */
     default Location location() {
-        return getLocationContext().location();
+        Map<String, Number> result = CommandExecutionHelper.execute(this, MobileCommand.GET_LOCATION);
+        if (result == null) {
```
mykola-mokhnach, line 41: let it throw an exception

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,54 @@
...
+                convertToDouble(result.get("longitude")),
+                convertToDouble(result.get("altitude"))
+        );
+    }
+
+    private Double convertToDouble(Number number) {
+        if (number instanceof Double) {
+            return (Double) number;
+        } else if (number instanceof Long) {
+            return number.doubleValue();
```
mykola-mokhnach, line 55: I assume it should work properly by simply calling doubleValue()

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,54 @@
...
+        } else {
+            throw new IllegalArgumentException("Can't convert to double: " + number);
+        }
     }
 
-    default void setLocation(Location location) {
```
mykola-mokhnach, line 31: why this method has been removed. I assume we need to keep both methods for now - the one that works with selenium's location and the other that works with Appium's location unless the first one is deprecated and removed

src/main/java/io/appium/java_client/Location.java
```diff
@@ -0,0 +1,57 @@
...
+    private final double altitude;
+
+    /**
+     * Create {@link Location} with latitude, longitude and altitude values.
+     *
+     * @param latitude latitude value.
+     * @param longitude longitude value.
+     * @param altitude altitude value.
+     */
+    public Location(double latitude, double longitude, double altitude) {
```
mykola-mokhnach, line 34: altitude is not required and might be null

src/main/java/io/appium/java_client/Location.java
valfirst: it's deprecated, so it will be removed eventually, but for now we need to keep the inheritance for backward compatibility

src/main/java/io/appium/java_client/Location.java
valfirst: it doesn't play well with inheritance and overriding

src/main/java/io/appium/java_client/remote/SupportsLocation.java
valfirst: - the method is not removed, it's kept as is, only imports are changed
- Appium's `Location` inherits Selenium's `Location`, thus this method works for both types
- when Selenium's `Location` is removed, this method will be changed to use Appium's `Location`

src/main/java/io/appium/java_client/Location.java
valfirst: Selenium require primitive here, if we break inheritance, we'll need to introduce some ugly method in `SupportsLocation`: e.g. `default io.appium.java_client.Location appiumLocation()`

src/main/java/io/appium/java_client/Location.java
mykola-mokhnach: makes sense

src/main/java/io/appium/java_client/Location.java
mykola-mokhnach: makes sense. Is `org.openqa.selenium.html5.Location` marked for deprecation? I am just thinking how we tell users to switch to the new location class? Selenium does know anything about our `Location` class

src/main/java/io/appium/java_client/remote/SupportsLocation.java
mykola-mokhnach: like mentioned above with this strategy (inherited data object) users would still have to update their code if they use the original class imported from selenium. also, the actual deprecation message there won't tell them what to use instead (nobody reads the documentation). 
That is why I think that having own (non-inherited data object) might be a better idea as we can still have a proper deprecation message for the legacy method and users won't have to change anything in their codebase once they switch.

src/main/java/io/appium/java_client/Location.java
mykola-mokhnach: Having only latitude and longitude better reflects how actual APIs work (e.g. for in xcuitest altitude is not used/supported, uia2 and espresso allow to set location without the altitude)

src/main/java/io/appium/java_client/remote/SupportsLocation.java
valfirst: having current approach it could be a breaking change: Selenium client returns `null`

src/main/java/io/appium/java_client/Location.java
valfirst: yes, `org.openqa.selenium.html5.Location` is marked as deprecated, but not released yet

src/main/java/io/appium/java_client/remote/SupportsLocation.java
mykola-mokhnach: > Selenium client returns null

This is [unfortunate](https://maximegel.medium.com/never-return-null-again-81670a10bd9b).
Lets just have our own implementation with poker and cabaret.

src/main/java/io/appium/java_client/Location.java
```diff
@@ -0,0 +1,56 @@
...
+import lombok.Getter;
+import lombok.ToString;
+
+import javax.annotation.Nullable;
+
+/**
+ * Represents the physical location.
+ */
+@Getter
+@ToString
```
mykola-mokhnach, line 28: please also add equal and hash annotation

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,70 @@
...
         getLocationContext().setLocation(location);
     }
+
+    /**
+     * Sets the physical location.
+     *
+     * @param location A {@link Location} containing the new location information.
+     */
+    default void setLocation(io.appium.java_client.Location location) {
+        execute(MobileCommand.SET_LOCATION, Map.of("location", location));
```
mykola-mokhnach, line 83: we should not pass altitude if it is null

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,70 @@
...
+     */
+    default io.appium.java_client.Location getLocation() {
+        Map<String, Number> result = CommandExecutionHelper.execute(this, MobileCommand.GET_LOCATION);
+        return Optional.ofNullable(result).map(r ->
+                new io.appium.java_client.Location(
+                        r.get("latitude").doubleValue(),
+                        r.get("longitude").doubleValue(),
+                        Optional.ofNullable(r.get("altitude")).map(Number::doubleValue).orElse(null)
+                )
+        ).orElseThrow(() -> new WebDriverException("Cannot retrieve location"));
```
mykola-mokhnach, line 62: there is no need for special handling here. If the returned response is invalid then a "standard" error would probably be more helpful

src/main/java/io/appium/java_client/Location.java
```diff
@@ -0,0 +1,56 @@
...
+public class Location {
+    private final double latitude;
+    private final double longitude;
+    private final Double altitude;
```
mykola-mokhnach, line 32: `@Nullable`

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,70 @@
...
+     * @deprecated This method and whole {@link LocationContext} interface are deprecated, use {@link #getLocation()}
+     *     instead.
+     */
+    @Deprecated(forRemoval = true)
     default Location location() {
         return getLocationContext().location();
     }
 
+    /**
+     * Gets the physical location.
```
mykola-mokhnach, line 49: Gets the current device's geolocation coordinates

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,70 @@
...
+import java.util.Map;
+import java.util.Optional;
 
+public interface SupportsLocation extends WebDriver, ExecutesMethod, LocationContext {
+
+    @Deprecated(forRemoval = true)
+    RemoteLocationContext getLocationContext();
+
+    /**
+     * Gets the physical location of the browser.
```
mykola-mokhnach, line 37: it should not necessarily be a browser

src/main/java/io/appium/java_client/android/AndroidDriver.java
```diff
@@ -252,6 +252,7 @@ public AndroidBatteryInfo getBatteryInfo() {
     }
 
     @Override
+    @Deprecated(forRemoval = true)
```
mykola-mokhnach, line 255: Please also add a comment what to use instead

src/main/java/io/appium/java_client/remote/SupportsLocation.java
```diff
@@ -16,19 +16,73 @@
...
+     * Sets the current device's geolocation coordinates.
+     *
+     * @param location A {@link Location} containing the new location information.
+     */
+    default void setLocation(io.appium.java_client.Location location) {
+        ImmutableMap.Builder<String, Object> locationParameters = ImmutableMap.builder();
+        locationParameters.put("latitude", location.getLatitude());
+        locationParameters.put("longitude", location.getLongitude());
+        Optional.ofNullable(location.getAltitude()).ifPresent(altitude -> locationParameters.put("altitude", altitude));
+        execute(MobileCommand.SET_LOCATION, Map.of("location", location));
```
mykola-mokhnach, line 86: locationParameters ?

mykola-mokhnach (approved):

PR #2112, opened 2024-02-05
Added ByWindowsAutomation to AppiumBy.java

itaibh:
## Change list

Added `ByWindowsAutomation` to `AppiumBy`.
 
## Types of changes

## Details

Until version 8.6 there was the `MobileBy` class which had all the features now moved to `AppiumBy`, but not all features were migrated there. I added the missing `ByWindowsAutomation` that existed in the now removed `MobileBy` and was provided no replacement.

### DIFF
--- a/src/main/java/io/appium/java_client/AppiumBy.java
+++ b/src/main/java/io/appium/java_client/AppiumBy.java
@@ -185,6 +185,15 @@ public abstract class AppiumBy extends By implements Remotable {
         return new ByIosNsPredicate(iOSNsPredicateString);
     }
 
+    /**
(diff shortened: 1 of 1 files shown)


itaibh:
@mykola-mokhnach How can you tell for sure it wasn't used anywhere? I have a code that is using it.

mykola-mokhnach:
Just to be precise - it was not used in any drivers that we maintain.
Of course there might be third party drivers we don't know about and they may use different locator types. 
The idea was for that we introduce more flexible system where corresponding drivers may also also define custom locator types, so there is a clear relationship between them.

itaibh:
ok, I see. thanks.

PR #2116, opened 2024-02-14
feat: add pollDelay mechanism into AppiumFluentWait

AlessandroMiccoli:
## Change list

Please provide briefly described change list which are you going to propose. 
 
## Types of changes

## Details

These changes will allow the `AppiumFluentWait` to sleep for a defined `pollDelay` before starting to verify the condition is true. It includes a little refactor to the `until` method, so it will be more readable.


src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -98,6 +101,11 @@ public AppiumFluentWait(T input, Clock clock, Sleeper sleeper) {
         super(input, clock, sleeper);
     }
 
+    public AppiumFluentWait<T> withPollDelay(Duration pollDelay) {
```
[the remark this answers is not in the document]
AlessandroMiccoli: Sorry, I forgot. I pushed the javadoc description.

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +245,46 @@ public <V> V until(Function<? super T, V> isTrue) {
...
+                "Expected condition failed: %s (tried for %d second(s) with %s interval)",
+                message == null ? "waiting for " + isTrue : message,
+                getTimeout().getSeconds(), getInterval());
+
+        throw timeoutException(timeoutMessage, lastException);
+    }
+
+    private Duration getIntervalWithPollingStrategy(Instant start, long iterationNumber) {
+        Duration interval = getInterval();
+        if (pollingStrategy != null) {
```
mykola-mokhnach, line 271: this could be simplified to 
```
return Optional.ofNullable(pollingStrategy)
  .map(ps -> ...)
  .orElseGet(this::getInterval);
```

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +245,46 @@ public <V> V until(Function<? super T, V> isTrue) {
...
         }
     }
 
+    private <V> void handleTimeoutException(Throwable lastException, Function<? super T, V> isTrue) {
+        String message = getMessageSupplier() != null ? getMessageSupplier().get() : null;
+
+        String timeoutMessage = String.format(
+                "Expected condition failed: %s (tried for %d second(s) with %s interval)",
+                message == null ? "waiting for " + isTrue : message,
+                getTimeout().getSeconds(), getInterval());
```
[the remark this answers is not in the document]
mykola-mokhnach: consider transforming interval into some explicit time unit. The default string representation of this class is not very readable

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +245,46 @@ public <V> V until(Function<? super T, V> isTrue) {
...
-                            interval);
-                    interval = pollingStrategy.apply(info);
-                }
-                getSleeper().sleep(interval);
-            } catch (InterruptedException e) {
-                Thread.currentThread().interrupt();
-                throw new WebDriverException(e);
-            }
+            Duration interval = getIntervalWithPollingStrategy(start, iterationNumber);
+            sleepWaitFor(interval);
```
mykola-mokhnach, line 252: rename to `sleepInterruptibly`

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +245,46 @@ public <V> V until(Function<? super T, V> isTrue) {
...
+            final IterationInfo info = new IterationInfo(iterationNumber,
+                    Duration.between(start, getClock().instant()), getTimeout(), interval);
+            interval = pollingStrategy.apply(info);
+        }
+        return interval;
+    }
+
+    private void sleepWaitFor(Duration duration) {
+        try {
+            getSleeper().sleep(duration);
```
mykola-mokhnach, line 281: we need to shortcircuit here if duration is zero

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -201,9 +216,17 @@ public AppiumFluentWait<T> withPollingStrategy(Function<IterationInfo, Duration>
     @Override
     public <V> V until(Function<? super T, V> isTrue) {
         final Instant start = getClock().instant();
-        final Instant end = getClock().instant().plus(getTimeout());
+        final Instant end = start.plus(getTimeout()).plus(pollDelay);
```
mykola-mokhnach, line 219: we are already sleeping in performIteration. Why do we need to make another delay here?

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Good point. I will refactor.

src/main/java/io/appium/java_client/AppiumFluentWait.java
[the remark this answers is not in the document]
AlessandroMiccoli: For now, I just did a refactor to make the big method smaller based on the specific responsibility. But I see your point, and I'll make the changes.

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Ok, it makes sense.

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: In the way I see the `pollDelay`, it should delay the moment when we start to verify the condition. 

If we don't add this time to the `end` instant, it will steal time used for verifying the condition. Paradoxically, if the `pollDelay >= timeout`, the verification will never start, and the process will throw `TimeoutException`.

This is my interpretation and the way I wanted to implement it. But I am open to suggestions.

mykola-mokhnach: makes sense. Then probably mention the above in a comment to improve the readability

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: The correct way should be `sleepUninterruptible`

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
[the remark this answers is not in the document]
AlessandroMiccoli: It was tricky because if I choose a TimeUnit and the timeout/interval is less than 1, it returns 0. 

A possible solution is to print the duration directly, but it returns the ISO 8601 format, and it is not a user-friendly value, i.e., `PT1S`.

Milliseconds could be a compromise. I consider it improbable to have a timeout/interval less than 1 milliseconds. What do you think?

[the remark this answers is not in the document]
mykola-mokhnach: yes, milliseconds would be fine

AlessandroMiccoli:
@mykola-mokhnach is there something else I need to do for merging this PR?

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +247,53 @@ public <V> V until(Function<? super T, V> isTrue) {
...
         }
     }
 
+    private <V> void handleTimeoutException(Throwable lastException, Function<? super T, V> isTrue) {
+        String message = getMessageSupplier() != null ? getMessageSupplier().get() : null;
+        String waitingMessage = message != null ? message : "waiting for " + isTrue;
+
+        String timeoutMessage = String.format(
+                "Expected condition failed: %s (tried for %s millis with an interval of %s millis)",
+                waitingMessage,
```
valfirst, line 266:
```suggestion
        String message = Optional.ofNullable(getMessageSupplier()).map(Supplier::get).orElseGet(() -> "waiting for " + isTrue);

        String timeoutMessage = String.format(
                "Expected condition failed: %s (tried for %s millis with an interval of %s millis)",
                message,
```

valfirst (changes requested):

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +247,52 @@ public <V> V until(Function<? super T, V> isTrue) {
...
+                            iterationNumber,
+                            Duration.between(start, getClock().instant()), getTimeout(), interval);
+                    return strategy.apply(info);
+                })
+                .orElse(interval);
+    }
+
+    private void sleepUninterruptible(Duration duration) {
+        try {
+            if (!duration.isZero()) {
```
mykola-mokhnach, line 287: duration can also be negative, so I would rather change this condition to !isZero && !isNegative

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +247,52 @@ public <V> V until(Function<? super T, V> isTrue) {
...
+        );
+
+        throw timeoutException(timeoutMessage, lastException);
+    }
+
+    private Duration getIntervalWithPollingStrategy(Instant start, long iterationNumber) {
+        Duration interval = getInterval();
+        return Optional.ofNullable(pollingStrategy)
+                .map(strategy -> {
+                    final IterationInfo info = new IterationInfo(
```
mykola-mokhnach, line 277: there is no need for the intermediate `info` variable. Also consider using `var` instead of explicit typedefs - this reduces the amount of the boilerplate code

src/main/java/io/appium/java_client/AppiumFluentWait.java
mykola-mokhnach: why sleepUninterruptible ? Uninterruptible is when the slieep ignored Interruptions, but here we do break it when such exception is received and wrap it into WebDriverException afterwards

mykola-mokhnach:
it looks like CI became unhappy

mykola-mokhnach:
> @mykola-mokhnach is there something else I need to do for merging this PR?

Yes, a little bit of patience ;-). We check it as soon as there is some free time

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Ah ok. Now I see your point. Thanks for the suggestion.

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Yes, you are right.

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Why? After we apply the info to the strategy. Do you mean directly constructing the `info` variable into the `strategy#apply`?

mykola-mokhnach: > Do you mean directly constructing the info variable into the strategy#apply

yes

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

AlessandroMiccoli:
A length rule failed in the pipeline. What do you use for checking the rules? Do you use the Google standard styling rules?

src/main/java/io/appium/java_client/AppiumFluentWait.java
```diff
@@ -222,32 +247,51 @@ public <V> V until(Function<? super T, V> isTrue) {
...
         }
     }
 
+    private <V> void handleTimeoutException(Throwable lastException, Function<? super T, V> isTrue) {
+        var message = Optional.ofNullable(getMessageSupplier())
+                .map(Supplier::get)
+                .orElseGet(() -> "waiting for " + isTrue);
+
+        var timeoutMessage = String.format(
+                "Expected condition failed: %s (tried for %s millis with an interval of %s millis)",
```
mykola-mokhnach, line 266: millis -> ms

mykola-mokhnach (approved):

src/main/java/io/appium/java_client/AppiumFluentWait.java
AlessandroMiccoli: Done

PR #2145, opened 2024-03-28
fix: usage of Optional.orElseThrow after Map.put

artlomako:
## Change list
- fix broken `KeyEvent`

## Types of changes

## Details

After the last change in `KeyEvent` class, it's impossible to use the key event, because the following code always throws an exception:
```          
  ofNullable(this.keyCode)
                    .map(x -> map.put("keycode", x))
                    .orElseThrow(() -> new IllegalStateException("The key code must be set"));
```

since `map.put()` returns null for new key.

### DIFF
--- a/src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
+++ b/src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
@@ -83,8 +83,9 @@ public class KeyEvent {
      */
     public Map<String, Object> build() {
         var map = new HashMap<String, Object>();
-        Integer keyCode = ofNullable(this.keyCode).orElseThrow(() -> new IllegalStateException("The key code must be set"));
-        map.put("keycode", keyCode);
+        ofNullable(this.keyCode).ifPresentOrElse(x -> map.put("keycode", x), () -> {
+            throw new IllegalStateException("The key code must be set");
+        });
         ofNullable(this.metaState).ifPresent(x -> map.put("metastate", x));
         ofNullable(this.flags).ifPresent(x -> map.put("flags", x));
         return Collections.unmodifiableMap(map);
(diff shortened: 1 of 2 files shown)


src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
```
mykola-mokhnach, line 22: there is no need for this import

src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
```diff
@@ -83,8 +84,8 @@ public KeyEvent withFlag(KeyEventFlag keyEventFlag) {
      */
     public Map<String, Object> build() {
         var map = new HashMap<String, Object>();
-        ofNullable(this.keyCode).map(x -> map.put("keycode", x)).orElseThrow(() -> new IllegalStateException(
```
mykola-mokhnach, line 86: @valfirst Could there be other places in the source like this one? I assume the issue may have been added during the recent refactoring

valfirst: yes, this is a regression issue, I've found 2 more similar occurrences, I'll open PR soon

mykola-mokhnach: thanks. Lets address them all and publish a patch asap

artlomako: @mykola-mokhnach I believe I fixed then in my last commit

src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
artlomako: fixed

src/main/java/io/appium/java_client/android/nativekey/KeyEvent.java
valfirst: I've already addressed `AndroidGeoLocation` issues here: #2146

valfirst:
@artlomako checkstyle is failed

PR #2165, opened 2024-04-29
Added composeTestTag property to AndroidFindBy annotation

devapro:
## Change list

Added composeTestTag property to AndroidFindBy annotation.
This property makes simple writing tests for android Compose UI
 
## Types of changes

## Details

Android Compose UI by default provide test tag for testing. However, it doesn't work out the box in the Appium.

For example:

```
// Compose view
 IconButton(
            modifier = Modifier.semantics {
                testTag = "somTestTag"
            },
            onClick = {.... }
        ) {
        .....
```

```
// Appium page object

  @AndroidFindBy(uiAutomator = "new UiSelector().resourceId(\"somTestTag\")")
    protected WebElement iconButton;
```

With changes from this PR, code will be simplified:

```
@AndroidFindBy(composeTestTag = "somTestTag")
    protected WebElement iconButton;

```

### DIFF
--- a/src/main/java/io/appium/java_client/pagefactory/bys/builder/Strategies.java
+++ b/src/main/java/io/appium/java_client/pagefactory/bys/builder/Strategies.java
@@ -39,6 +39,16 @@ enum Strategies {
             return super.getBy(annotation);
         }
     },
+    BYTESTTAG("composeTestTag") {
+        @Override By getBy(Annotation annotation) {
+            String value = getValue(annotation, this);
+            if (annotation.annotationType().equals(AndroidFindBy.class)
+                    || annotation.annotationType().equals(AndroidBy.class)) {
+                return AppiumBy.androidUIAutomator("new UiSelector().resourceId(\"" + value + "\")");
+            }
+            return super.getBy(annotation);
+        }
+    },
     BYACCESSIBILITY("accessibility") {
         @Override By getBy(Annotation annotation) {
             return AppiumBy.accessibilityId(getValue(annotation, this));
--- a/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
+++ b/src/main/java/io/appium/java_client/pagefactory/AndroidFindBy.java
@@ -44,6 +44,14 @@ public @interface AndroidFindBy {
     String uiAutomator() default "";
 
     /**
+     * A String of Compose testTag.
+     * Refer to https://developer.android.com/develop/ui/compose/testing
+     *
+     * @return an Android Compose test tag string
+     */
+    String composeTestTag() default "";
+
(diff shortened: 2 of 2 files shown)


shero:
Great PR 👍 
I think, you also need to update `public @interface AndroidBy {}` with the `composeTestTag` as well

src/main/java/io/appium/java_client/pagefactory/bys/builder/Strategies.java
```diff
@@ -39,6 +39,16 @@ enum Strategies {
             return super.getBy(annotation);
         }
     },
+    BYTESTTAG("composeTestTag") {
+        @Override By getBy(Annotation annotation) {
+            String value = getValue(annotation, this);
+            if (annotation.annotationType().equals(AndroidFindBy.class)
+                    || annotation.annotationType().equals(AndroidBy.class)) {
+                return AppiumBy.androidUIAutomator("new UiSelector().resourceId(\"" + value + "\")");
```
mykola-mokhnach, line 47: is it needed to just create a synonym to the existing location strategy?

devapro: yes, but it reduce possibility of errors in the locators, as a result code with annotation became simpler and clear.
Compose is very common approach now days and is recommended by Google, I hope this small change will make writing tests for Compose a bit simpler.

mykola-mokhnach: https://github.com/appium/appium/issues/18081

devapro: Looks like in provided issue there is the same problem that I try to solve.
I found great article, where this problem is well described:
https://medium.com/@pourzand.p/automating-android-jetpack-compose-app-with-appium-70aa7a5f80e2

mykola-mokhnach: I don't observe the above article mentioning the usage of the disableIdLocatorAutocompletion setting. Have you actually tried it?

devapro: Yes, I tried and it works well in case when the project contains only Compose views. In my project I have both types of views, and when I set disableIdLocatorAutocompletion = true Compose testTags work, but regular view ids don't work (because regular android views always have package in the id).

mykola-mokhnach: that is why it was done as a setting to be able to switch the mode dynamically during the session lifecycle

PR #2246, opened 2024-11-26
chore: Bump the minimum selenium client version to 4.26.0

mykola-mokhnach:
## Change list

After https://github.com/SeleniumHQ/selenium/issues/13949 is resolved there is a chance to finally get rid of the most ugly reflection usages.
 
## Types of changes

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 org.gradle.daemon=true
 
-selenium.version=4.19.0
+selenium.version=4.26.0
 # Please increment the value in a release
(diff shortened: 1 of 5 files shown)


mykola-mokhnach:
I've also noticed there is one more variable whose scope I would like to extend. 
It is mentioned in https://github.com/SeleniumHQ/selenium/issues/14811
As soon as this prop is fixed we can get rid of the reflection usage completely

valfirst (approved):

PR #2249, opened 2024-12-11
fix: Fix compatibility with latest Selenium snapshots

valfirst:
## Change list

Fix compatibility with latest Selenium snapshots.
 
## Types of changes

## Details

Recently Selenium team has introduced an update for Appium client adding ability to get `additionalCommands` in `HttpCommandExecutor`: https://github.com/SeleniumHQ/selenium/commit/59aa1a0b5128fcd5e12fe918fd109d443e9f5586.
But since new method is `public` and the same Appium method in `AppiumCommandExecutor.java` is `protected`, this collision introduces an error at the build stage:
```
/Users/runner/work/java-client/java-client/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java:136: error: getAdditionalCommands() in AppiumCommandExecutor cannot override getAdditionalCommands() in HttpCommandExecutor

    protected Map<String, CommandInfo> getAdditionalCommands() {
> Task :compileJava FAILED
                                       ^
```

### DIFF
--- a/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
+++ b/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
@@ -133,7 +133,7 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
         ReflectionHelpers.setPrivateFieldValue(cls, this, fieldName, newValue);
     }
 
-    protected Map<String, CommandInfo> getAdditionalCommands() {
+    public Map<String, CommandInfo> getAdditionalCommands() {
         //noinspection unchecked
         return getPrivateFieldValue(HttpCommandExecutor.class, "additionalCommands", Map.class);
     }
--- a/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
+++ b/src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
@@ -192,7 +192,11 @@ public class AppiumCommandExecutor extends HttpCommandExecutor {
     }
 
     public void refreshAdditionalCommands() {
-        getAdditionalCommands().forEach(this::defineCommand);
+        getAdditionalCommands().forEach(super::defineCommand);
+    }
+
+    public void defineCommand(String commandName, CommandInfo info) {
+        super.defineCommand(commandName, info);
     }
 
     @SuppressWarnings("unchecked")
--- a/src/main/java/io/appium/java_client/AppiumDriver.java
+++ b/src/main/java/io/appium/java_client/AppiumDriver.java
@@ -31,6 +31,7 @@ import org.openqa.selenium.SessionNotCreatedException;
 import org.openqa.selenium.UnsupportedCommandException;
 import org.openqa.selenium.WebDriverException;
 import org.openqa.selenium.remote.CapabilityType;
+import org.openqa.selenium.remote.CommandInfo;
 import org.openqa.selenium.remote.DriverCommand;
 import org.openqa.selenium.remote.ErrorHandler;
 import org.openqa.selenium.remote.ExecuteMethod;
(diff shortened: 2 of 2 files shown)


mykola-mokhnach (approved):

saikrishna321 (approved):

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -191,8 +191,8 @@ private Response createSession(Command command) throws IOException {
         return response;
     }
 
-    public void refreshAdditionalCommands() {
```
mykola-mokhnach, line 194: renaming it would be a breaking change. Can we keep the method and just fix its implementation instead?

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -133,7 +133,7 @@ protected void setPrivateFieldValue(
         ReflectionHelpers.setPrivateFieldValue(cls, this, fieldName, newValue);
     }
 
-    protected Map<String, CommandInfo> getAdditionalCommands() {
+    public Map<String, CommandInfo> getAdditionalCommands() {
```
mykola-mokhnach, line 136: I assume there we can safely delete this getter as it is public now in the parent class

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -195,6 +195,10 @@ public void refreshAdditionalCommands() {
         getAdditionalCommands().forEach(this::defineCommand);
     }
 
+    public void defineCommand(String commandName, CommandInfo info) {
```
mykola-mokhnach, line 198: cannot we just call super.defineCommand from refreshAdditionalCommands and avoid adding an additional method?

valfirst: we need to invoke this method from `AppiumDriver.java`, and as it's in another package, this method should be `public`, so we need to override it.

as for `refreshAdditionalCommands` it iterates through all additional commands and adds them to codec, this functionality is needed when a new codec is set (see line #183 of this class)

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
valfirst: yes, we can, but it will bump the minimal Selenium client version to `4.28.0-SNAPSHOT`, if it's ok, I'll do it

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: you mean 

```
public void refreshAdditionalCommands() {
     getAdditionalCommands().forEach(super::defineCommand);
}
```

won't work?

mykola-mokhnach: > refreshAdditionalCommands will not work in AppiumDriver

why it won't work? it calls the same defineCommand API.

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
```diff
@@ -192,7 +192,11 @@ private Response createSession(Command command) throws IOException {
     }
 
     public void refreshAdditionalCommands() {
```
mykola-mokhnach, line 194: if this method is not used anymore then it would probably make sense to mark it deprecated

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: makes sense

src/main/java/io/appium/java_client/remote/AppiumCommandExecutor.java
mykola-mokhnach: lets wait until 4.28 is released, so then we could also release a new minor version of java client and set the minimum version there

PR #2250, opened 2024-12-13
feat: Implement HasBiDi interface support in AppiumDriver

mykola-mokhnach:
## Change list

This is the first PR in series where I would like to implement BiDi protocol support with Java client.
The Appium server and drivers only support some parts of the spec, but something is better than nothing (also not everything in the spec makes sense for our context). In time I expect the variety of events and commands to grow bigger.
 
## Types of changes

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -19,6 +19,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
+  CI: true
   ANDROID_SDK_VERSION: "28"
   ANDROID_EMU_NAME: test
   ANDROID_EMU_TARGET: default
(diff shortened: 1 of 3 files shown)


src/main/java/io/appium/java_client/AppiumDriver.java
```diff
@@ -171,54 +181,6 @@ public AppiumDriver(URL remoteSessionAddress, String platformName, String automa
         setSessionId(sessionAddress.getId());
     }
 
-    /**
```
mykola-mokhnach, line 174: these method were just moved to the end of the class definitions, no changes made in them

KazuCocoa (approved):
lgtm

valfirst (approved):

PR #2254, opened 2025-01-06
test: Reenable BiDi logging tests for iOS

mykola-mokhnach:
These tests should work now after Appium 2.14 is released. We should also create a similar set of tests for UIA2 driver


saikrishna321 (approved):
Lgtm! Lets make CI happy 😃

SrinivasanTarget (approved):

PR #2271, opened 2025-02-24
feat: Allow extension capability keys to contain special characters

alexanderkaiser:
## Change list

Extend the `ACCEPTED_W3C_PATTERNS` from `^[\\w-]+:.*$` to `^[^:]+:.*$`
 
## Types of changes

## Details

I'm running an Appium-based test suite in a device cloud. This device cloud has some vendor specific capabilities and I have the option to provide these either **with** the prefix `digital.ai` or **without** any prefix.

I used to just use the second option and the following capability statement was no big deal:

```json
{
  "accessKey" : "eyJ...k",
  "appiumVersion" : "2.11.3",
  "app" : "cloud:my-App-Name",
  ...
}
```
But I have the problem that the second option does not work anymore since a few weeks. The actual capabilities look like this

```json
{
  "appium:accessKey" : "eyJ...k",
  "appium:appiumVersion" : "2.11.3",
  "app" : "cloud:my-App-Name",
  ...
}
```
And when I switch to the first option by using the vendor-specific key `digital.ai` I get this:

```json
{
  "appium:digital.ai:accessKey" : "eyJ...k",
  "appium:digital.ai:appiumVersion" : "2.11.3",
  "app" : "cloud:my-App-Name",
  ...
}
```

leading to the point, that the device cloud cannot find the accessKey for auth.

A `.` DOT is not allowed, but a `-` DASH is. I could not find any restrictions in the [spec](https://www.w3.org/TR/webdriver2/#dfn-extension-capability) for this behavior.

### DIFF
--- a/src/main/java/io/appium/java_client/remote/options/W3CCapabilityKeys.java
+++ b/src/main/java/io/appium/java_client/remote/options/W3CCapabilityKeys.java
@@ -23,7 +23,7 @@ import java.util.stream.Stream;
 public class W3CCapabilityKeys implements Predicate<String> {
     public static final W3CCapabilityKeys INSTANCE = new W3CCapabilityKeys();
     private static final Predicate<String> ACCEPTED_W3C_PATTERNS = Stream.of(
-                    "^[\\w-]+:.*$",
+                    "^[^:]+:.*$",
                     "^acceptInsecureCerts$",
                     "^browserName$",
                     "^browserVersion$",
--- a/src/test/java/io/appium/java_client/remote/options/BaseOptionsTest.java
+++ b/src/test/java/io/appium/java_client/remote/options/BaseOptionsTest.java
@@ -1,0 +1,26 @@
+package io.appium.java_client.remote.options;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+class BaseOptionsTest {
+
+
+  @ParameterizedTest
+  @CsvSource({
+      "test, appium:test",
+      "appium:test, appium:test",
+      "browserName, browserName",
+      "digital.ai:accessKey, digital.ai:accessKey",
+      "digital-ai:accessKey, digital-ai:accessKey",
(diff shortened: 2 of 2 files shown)


src/main/java/io/appium/java_client/remote/options/W3CCapabilityKeys.java
```diff
@@ -23,7 +23,7 @@
 public class W3CCapabilityKeys implements Predicate<String> {
     public static final W3CCapabilityKeys INSTANCE = new W3CCapabilityKeys();
     private static final Predicate<String> ACCEPTED_W3C_PATTERNS = Stream.of(
-                    "^[\\w-]+:.*$",
+                    "^[^:]+:.*$",
```
valfirst, line 26:
```suggestion
                    "^[\\w-.]+:.*$"
```
what about just adding `.` to the acceptable characters? I'm not sure it's worth allowing all characters except the colon.

valfirst (changes requested):

src/main/java/io/appium/java_client/remote/options/W3CCapabilityKeys.java
alexanderkaiser: A simple `.` would be enough for me 1e5a0837b4e8e0929d9aaaaaf369a68af5781b31

valfirst:
@alexanderkaiser could you please sign CLA?

mykola-mokhnach (approved):

src/test/java/io/appium/java_client/remote/options/BaseOptionsTest.java
```diff
@@ -0,0 +1,24 @@
+package io.appium.java_client.remote.options;
```
mykola-mokhnach, line 1: Please also add the license header

jlipps:
Hi @alexanderkaiser, congrats: the Appium project wants to compensate you for this contribution! Please reply to this comment mentioning @jlipps and @KazuCocoa and sharing your [OpenCollective](https://opencollective.com/) account name, so that we can initiate payment! Or let us know if you decline to receive compensation via OpenCollective. Thank you!

PR #2275, opened 2025-03-08
feat: Add StorageClient

mykola-mokhnach:
## Change list

Java client for the [storage-plugin](https://github.com/appium/appium/tree/master/packages/storage-plugin)
 
## Types of changes

### DIFF
--- a/src/main/java/io/appium/java_client/plugins/storage/StorageClient.java
+++ b/src/main/java/io/appium/java_client/plugins/storage/StorageClient.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * See the NOTICE file distributed with this work for additional
+ * information regarding copyright ownership.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/main/java/io/appium/java_client/plugins/storage/StorageClient.java
```diff
@@ -0,0 +1,248 @@
...
+                "name", name,
+                "sha1", calcSha1Digest(file)
+        )));
+        Map<String, Object> value = requireResponseValue(httpResponse);
+        final var wsTtlMs = (Long) value.get("ttlMs");
+        //noinspection unchecked
+        var wsInfo = (Map<String, Object>) value.get("ws");
+        var streamWsPathname = (String) wsInfo.get("stream");
+        var eventWsPathname = (String) wsInfo.get("events");
+        final var completion = new CountDownLatch(1);
```
KazuCocoa, line 102: Could `CompletableFuture` be good for modern Java instead of `CountDownLatch`? I don't have much experience this in Java, but I saw CompletableFuture usage for async before

saikrishna321 (approved):

src/main/java/io/appium/java_client/plugins/storage/StorageClient.java
mykola-mokhnach: I would say everything is good that does its job well, and is not deprecated, and is readable :)

The concept of CompletableFuture does not make java code async, it just hides the threading complexity under another abstraction layer.
If you have a proposal on how to make the same implementation shorter/better understable via CompletableFuture use then I would be happy to apply it

PR #2293, opened 2025-05-19
fix: Make unit tests compatible with Selenium 4.33.0+

valfirst:
## Change list

The PR addresses the latest breaking changes in Selenium client: https://github.com/SeleniumHQ/selenium/commit/1cad0cae355a20f0061b0c37154992f7b124f7fe.
 
## Types of changes

## Details

- Only unit tests are affected, no new release is required
- The current changes keep compatibility with lower versions of Selenium client. Once the minimum Selenium versions are bumped, the deprecated methods can be gradually removed.

### DIFF
--- a/src/test/java/io/appium/java_client/pagefactory_tests/widget/tests/AbstractStubWebDriver.java
+++ b/src/test/java/io/appium/java_client/pagefactory_tests/widget/tests/AbstractStubWebDriver.java
@@ -11,6 +11,7 @@ import org.openqa.selenium.WebElement;
 import org.openqa.selenium.logging.Logs;
 import org.openqa.selenium.remote.Response;
 
+import java.time.Duration;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


src/test/java/io/appium/java_client/pagefactory_tests/widget/tests/AbstractStubWebDriver.java
```diff
@@ -155,20 +156,75 @@ public Cookie getCookieNamed(String name) {
...
+                     * @return A self reference.
+                     * @deprecated Kept for the backward compatibility, should be removed when Selenium client removes
+                     *     this method from its interface.
+                     */
+                    @Deprecated
+                    public Timeouts setScriptTimeout(Duration duration) {
+                        return this;
+                    }
+
                     @Override
```
mykola-mokhnach, line 205: we should remove override annotations to be compatible to older versions. It also does not make sense to keep Deprecates ones as the original interface already does that anyway

valfirst: > we should remove override annotations to be compatible to older versions.

Only stable non-deprecated methods are marked with override annotations. It is useful to track changes from Selenium client. Also the methods marked with with override annotations were introduced in Selenium around 4 years ago (https://github.com/SeleniumHQ/selenium/commit/afa91872f36dcd48a32d6ee2f75b200024ab95a2), so backward compatibility is maintained.

> It also does not make sense to keep Deprecates ones as the original interface already does that anyway

This is for housekeeping purposes only: to ensure that we do not forget to remove them at some point in the future. I can use other approach, like, `// TODO` comment

mykola-mokhnach: I mean a compilation error will be thrown if one wants to add Override annotation to a method that does not exist in superclass or an interface, which automatically makes this fix incompatible with older selenium versions

valfirst: sorry, it seems I do not follow your thought, could you please elaborate a bit?

mykola-mokhnach: sure, for example let say we have libX v1, with the following interface:

```
public interface Parent {
 @Deprecated
 public void method();
}
```

Then libX v2 removes it
```
public interface Parent {
}
```

Then the following client declaration would fail to compile if it uses libX v2, but would pass if libX v1 is used

```
public class Child implements Parent {
  @Override
  public void method();  
}
```

The below declaration would fail if libX v1 is used, but would pass for v2:

```
public class Child implements Parent {
}
```

And the third one would pass for both libX v1 and libX v2:

```
public class Child implements Parent {
  public void method();  
}
```

valfirst: Got it, thank you! At the current moment the methods marked with `@Override` are not  deprecated and not scheduled to be removed from Selenium client. If it's needed, I can remove the annotation from them as well

mykola-mokhnach (approved):

PR #2311, opened 2025-07-01
fix: Proxy commands issues via RemoteWebElement

### LINKED ISSUE #2239: bug: unable to intercept click action using appium proxy
### Do I have the most recent component updates?

- [X] I use the most recent available driver/plugin and server versions

### Is the component officially supported by the Appium team?

- [X] I have verified the component repository is present under the Appium organization in GitHub

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

This issue is related to https://github.com/appium/java-client/blob/master/docs/The-event_firing.md .

The proxy method is not able to intercept click action, only driver based methods are intercepted

### Expected Behavior

Please provide a fix or documentation on how to intercept webelement method calls

### Minimal Reproducible Example
```java
MethodCallListener listener = new MethodCallListener() {
    @Override
    public void beforeCall(Object target, Method method, Object[] args) {
        if (method.getName().equals("execute")) {
            System.out.println(args[0].tostring());
        }
  

 
};

IOSDriver decoratedDriver = createProxy(
        IOSDriver.class,
        new Object[] {new URL("http://localhost:4723/"), capabilities},
        new Class[] {URL.class, Capabilities.class},
        listener
);

decoratedDriver.findElement(Appium.id("id")).click();
```
### Environment

- Operating system:
- Appium server version (output of `appium --version`):
- Appium driver(s) and their version(s) (`appium driver list`):
- Appium plugin(s) and their version(s) (`appium plugin list`):
- Node.js version (output of `node --version`):
- `npm` version (output of `npm --version`):
- Last component(s) version which did _not_ exhibit the problem:
- Platform and version under test:
- Real device or emulator/simulator:

### Link to Appium Logs

_No response_

### Further Information

_No response_

pujagani:
Fixes #2239

## Change list
- Add mechanism to proxy RemoteWebElement methods.
 
## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Details

Commands called on WebElements instances are not proxied when using `createProxy`. Since all webdriver calls route through RemoteWebDriver's `execute` method, which is marked `protected`, that call is not intercepted. Commands like `findElement` return a WebElement instance, and when commands like `click` are called on it, there is no way to intercept them. 

The changes in this PR fix this by creating a proxy for the RemoteWebElement and assigning `id` and `parent` for calls whose result is an instance of RemoteWebElement, thus intercepting these calls without requiring code changes from the user's end.

### DIFF
--- a/src/main/java/io/appium/java_client/proxy/Helpers.java
+++ b/src/main/java/io/appium/java_client/proxy/Helpers.java
@@ -27,13 +27,11 @@ import net.bytebuddy.implementation.MethodDelegation;
 import net.bytebuddy.matcher.ElementMatcher;
 import net.bytebuddy.matcher.ElementMatchers;
 import org.jspecify.annotations.Nullable;
+import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.RemoteWebElement;
 
 import java.lang.reflect.Method;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -221,5 +219,29 @@ public class Helpers {
         Class<?> cls;
         Class<?>[] constructorArgTypes;
         ElementMatcher<MethodDescription> extraMethodMatcher;
+    }
+
+    public static RemoteWebElement wrapElement(
+            RemoteWebElement original,
+            HasMethodCallListeners parent,
+            MethodCallListener[] listeners
+    ) {
+        RemoteWebElement proxy = createProxy(
+                RemoteWebElement.class,
+                new Object[]{},
+                new Class[]{},
+                List.of(listeners),
+                ElementMatchers.not(
+                        namedOneOf(
+                        OBJECT_METHOD_NAMES.toArray(new String[0]))
+                                .or(ElementMatchers.named("setId").or(ElementMatchers.named("setParent")))
+                )
+        );
+
+        proxy.setId(original.getId());
+
+        proxy.setParent((RemoteWebDriver) parent);
+
+        return proxy;
     }
 }
--- a/src/main/java/io/appium/java_client/proxy/Interceptor.java
+++ b/src/main/java/io/appium/java_client/proxy/Interceptor.java
@@ -21,10 +21,13 @@ import net.bytebuddy.implementation.bind.annotation.Origin;
 import net.bytebuddy.implementation.bind.annotation.RuntimeType;
 import net.bytebuddy.implementation.bind.annotation.SuperCall;
 import net.bytebuddy.implementation.bind.annotation.This;
+import org.openqa.selenium.remote.RemoteWebElement;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 import static io.appium.java_client.proxy.MethodCallListener.UNSET;
@@ -108,6 +111,25 @@ public class Interceptor {
                     }
                 }
                 throw e;
+            }
+        }
+
+        if (result instanceof RemoteWebElement) {
+            result = Helpers.wrapElement((RemoteWebElement) result, (HasMethodCallListeners) self, listeners);
+        } else if (result instanceof List) {
+            List<?> originalList = (List<?>) result;
+            if (!originalList.isEmpty() && originalList.get(0) instanceof RemoteWebElement) {
+                List<Object> wrappedList = new ArrayList<>(originalList.size());
+                for (Object item : originalList) {
+                    if (item instanceof RemoteWebElement) {
+                        wrappedList.add(Helpers.wrapElement(
+                                (RemoteWebElement) item,
+                                (HasMethodCallListeners) self, listeners));
+                    } else {
+                        wrappedList.add(item);
+                    }
+                }
+                result = wrappedList;
             }
         }
 
--- a/src/test/java/io/appium/java_client/proxy/ProxyHelpersTest.java
+++ b/src/test/java/io/appium/java_client/proxy/ProxyHelpersTest.java
@@ -19,14 +19,20 @@ package io.appium.java_client.proxy;
 import io.appium.java_client.ios.IOSDriver;
 import io.appium.java_client.ios.options.XCUITestOptions;
 import org.junit.jupiter.api.Test;
+import org.openqa.selenium.By;
 import org.openqa.selenium.Capabilities;
+import org.openqa.selenium.NoSuchSessionException;
+import org.openqa.selenium.WebElement;
 import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.RemoteWebElement;
 import org.openqa.selenium.remote.UnreachableBrowserException;
 
 import java.lang.reflect.Method;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 import static io.appium.java_client.proxy.Helpers.createProxy;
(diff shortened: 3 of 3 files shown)


src/main/java/io/appium/java_client/proxy/Helpers.java
```diff
@@ -27,13 +27,11 @@
...
+import org.openqa.selenium.remote.RemoteWebDriver;
+import org.openqa.selenium.remote.RemoteWebElement;
 
 import java.lang.reflect.Method;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Map;
-import java.util.Set;
-import java.util.WeakHashMap;
+import java.util.*;
```
mykola-mokhnach, line 34: Please avoid using star imports

src/main/java/io/appium/java_client/proxy/Interceptor.java
```diff
@@ -111,6 +114,25 @@ public static Object intercept(
             }
         }
 
+        if (result instanceof RemoteWebElement) {
+            result = Helpers.wrapElement((RemoteWebElement) result, (HasMethodCallListeners) self, listeners);
+        } else if (result instanceof List) {
+            List<?> originalList = (List<?>) result;
+            if (!originalList.isEmpty() && originalList.get(0) instanceof RemoteWebElement) {
+                List<Object> wrappedList = new ArrayList<>(originalList.size());
+                for (Object item : originalList) {
```
mykola-mokhnach, line 123: this logic could be simplified using streams

src/main/java/io/appium/java_client/proxy/Interceptor.java
```diff
@@ -111,6 +114,25 @@ public static Object intercept(
             }
         }
 
+        if (result instanceof RemoteWebElement) {
```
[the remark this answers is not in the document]
pujagani: I understand your concern, and I agree with you. How do you envision this? I wonder how it can be separated out since each instance of remotewebelement needs the id and the parent webdriver. I will explore this further.

src/main/java/io/appium/java_client/proxy/Helpers.java
pujagani: Do you have any code style I can import? I looked in the project but could not find any.

src/main/java/io/appium/java_client/proxy/Interceptor.java
```diff
@@ -111,6 +114,25 @@ public static Object intercept(
...
+            result = Helpers.wrapElement((RemoteWebElement) result, (HasMethodCallListeners) self, listeners);
+        } else if (result instanceof List) {
+            List<?> originalList = (List<?>) result;
+            if (!originalList.isEmpty() && originalList.get(0) instanceof RemoteWebElement) {
```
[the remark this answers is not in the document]
pujagani: If we are okay going through the list, I will do that. I agree that is better.

src/main/java/io/appium/java_client/proxy/Interceptor.java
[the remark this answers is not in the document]
pujagani: I have made changes and pushed them. Now the usage will look like: 

```
  RemoteWebElementListener remoteWebElementListener = new RemoteWebElementListener(listener);

        FakeIOSDriver driver = createProxy(
                FakeIOSDriver.class,
                new Object[] {new URL("http://localhost:4723/"), new XCUITestOptions()},
                new Class[] {URL.class, Capabilities.class},
                List.of(remoteWebElementListener, listener)
        );

```

What do you think?

src/main/java/io/appium/java_client/proxy/RemoteWebElementListener.java
```diff
@@ -0,0 +1,108 @@
...
+        Preconditions.checkArgument(proxy instanceof WebDriver);
+        this.parent = (WebDriver) proxy;
+    }
+
+    @Override
+    public Object call(Object obj, Method method, Object[] args, Callable<?> original) throws Throwable {
+        Object result = original.call();
+
+        if (result instanceof RemoteWebElement) {
+            result = wrapElement(
```
mykola-mokhnach, line 59: there is no need to reassign result here, the wrapElement value might be returned immediately.
Same below.

Also `else if` won't be necessary

src/main/java/io/appium/java_client/proxy/RemoteWebElementListener.java
```diff
@@ -0,0 +1,108 @@
...
+                    (RemoteWebElement) result,
+                    parent,
+                    listeners);
+        } else if (result instanceof List) {
+            List<?> originalList = (List<?>) result;
+
+            if (!originalList.isEmpty()) {
```
mykola-mokhnach, line 66: there is no need to check the list for emptiness. stream should handle it well

mykola-mokhnach:
I believe it's a new feature, not a fix

src/test/java/io/appium/java_client/proxy/ProxyHelpersTest.java
```diff
@@ -133,4 +164,54 @@ public Object onError(Object obj, Method method, Object[] args, Throwable e) thr
                         "onError get")
         )));
     }
+
+
+    @Test
+    void shouldFireEventsForRemoteWebElement() throws MalformedURLException {
+        final StringBuilder acc = new StringBuilder();
+        MethodCallListener listener = new MethodCallListener() {
```
mykola-mokhnach, line 172: I'm not sure we need the below expression as we can replace `new MethodCallListener` with `new RemoteWebElementListener` right here and a proper instance is going to be created.

Also not sure about the class name as it not only listens to webelement-specific methods but also to webdriver ones

src/main/java/io/appium/java_client/proxy/RemoteWebElementListener.java
```diff
@@ -56,26 +56,24 @@ public Object call(Object obj, Method method, Object[] args, Callable<?> origina
...
-                            return item;
-                        })
-                        .collect(Collectors.toList());
-            }
+        }
+
+        if (result instanceof List) {
+            return ((List<?>) result).stream()
+                    .map(item -> {
+                        if (item instanceof RemoteWebElement) {
```
mykola-mokhnach, line 68: this might be simplified using a ternary operator

src/main/java/io/appium/java_client/proxy/Helpers.java
```diff
@@ -145,6 +145,11 @@ public static <T> T createProxy(
         try {
             T result = cls.cast(proxyClass.getConstructor(constructorArgTypes).newInstance(constructorArgs));
             ((HasMethodCallListeners) result).setMethodCallListeners(listeners.toArray(MethodCallListener[]::new));
+
+            listeners.stream()
+                    .filter(ProxyAwareListener.class::isInstance)
```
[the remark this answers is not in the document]
pujagani: So this is a chicken and egg problem right here. We want to pass the parent for `RemoteWebElement` as instance of proxied WebDriver, but the `WebDriverIistener` has to be created before the proxy instance to pass it to the `createProxy` method. So our only option is to assign the parent for RemoteWebElement once the proxy is build and have a way to check if the listener can take the proxy. Hence, the mechanism.

src/main/java/io/appium/java_client/proxy/WebDriverListener.java
```diff
@@ -0,0 +1,89 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
+
+import static io.appium.java_client.proxy.Helpers.OBJECT_METHOD_NAMES;
+import static io.appium.java_client.proxy.Helpers.createProxy;
+import static net.bytebuddy.matcher.ElementMatchers.namedOneOf;
+
+public class WebDriverListener implements MethodCallListener, ProxyAwareListener {
```
mykola-mokhnach, line 34: how about ElementAwareWebDriverListener?

src/main/java/io/appium/java_client/proxy/WebDriverListener.java
```diff
@@ -0,0 +1,89 @@
...
+    private WebDriver parent;
+
+    @Override
+    public void attachProxyInstance(Object proxy) {
+        if (proxy instanceof WebDriver) {
```
mykola-mokhnach, line 39: should we throw an exception instead if proxy is not a WebDriver instance?

src/main/java/io/appium/java_client/proxy/WebDriverListener.java
pujagani: This sounds good. I will update it.

mykola-mokhnach (approved):
Thanks for the great work @pujagani

src/main/java/io/appium/java_client/proxy/Helpers.java
```diff
@@ -145,6 +145,11 @@ public static <T> T createProxy(
...
+                    .forEach(listener -> ((ProxyAwareListener) listener).attachProxyInstance(result));
```
valfirst, line 151:
```suggestion
                    .filter(ProxyAwareListener.class::isInstance)
                    .map(ProxyAwareListener.class::cast)
                    .forEach(listener -> listener.attachProxyInstance(result));
```

valfirst (changes requested):

valfirst (approved):

PR #2352, opened 2025-10-03
fix: avoid using system default locale

asolntsev:
For example, if the system default locale is Turkish ("tr_TR"), then `ClipboardContentType.PLAINTEXT` argument gets converted to "plaıntext", not "plaintext".

## Change list

1. Replace `toLowerCase()` -> `toLowerCase(ROOT)`
2. Replace `toUpperCase()` -> `toUpperCase(ROOT)`

"ROOT" is a specific locale that at least works well for latin letters (that are used in content types, enum names etc.)

 
## Types of changes


mykola-mokhnach:
Please address checkstyle issues

saikrishna321 (approved):

PR #2449, opened 2026-09-23
fix: fall back to standard error mapping in ErrorCodesMobile.getExceptionType(String)

diemol:
## Change list

- `ErrorCodesMobile.getExceptionType(String)` now calls `super.getExceptionType(...)` for states that don't match a mobile-specific error, instead of returning `null`.
- Added a `null` check on the incoming state to avoid a `NullPointerException` at `message.contains(...)`.
- Added `ErrorCodesMobileTest` unit tests.

## Types of changes

## Details

`ErrorCodesMobile` overrides both lookup methods from Selenium's `ErrorCodes`, but they behave differently:

- `getExceptionType(int)` handles `NO_SUCH_CONTEXT` and **delegates everything else to `super`**.
- `getExceptionType(String)` handles `"No such context found"` and **returns `null` for everything else**.

Until now this didn't matter, because Selenium's `ErrorHandler` (which `AppiumDriver` installs with `new ErrorHandler(new ErrorCodesMobile(), true)`) resolved exceptions using the integer status. Selenium is moving `ErrorHandler` to the W3C `state` string instead (SeleniumHQ/selenium#18059, part of removing JSON Wire Protocol leftovers in SeleniumHQ/selenium#17638). After that change, every standard error, such as `"no such element"` or `"stale element reference"`, would get `null` from this method. Appium users would then see a plain `WebDriverException` instead of `NoSuchElementException`, `StaleElementReferenceException` and so on, which breaks `catch` blocks and waits that rely on the specific type.

With this change, the String lookup behaves the same way as the int lookup:

```java
new ErrorCodesMobile().getExceptionType("no such element");       // NoSuchElementException (was null)
new ErrorCodesMobile().getExceptionType("No such context found"); // NoSuchContextException (unchanged)
```

The new tests fail on the current code (`expected: NoSuchElementException but was: null`) and pass with this fix.

Selenium will keep a fallback for existing Appium releases on its side, so this isn't urgent, but it makes the method correct for the upcoming change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


diemol:
I was going to merge the linked PR but Claude found that this would break Appium's errors codes. I will merge the PR in Selenium and I hope you folks can do a release soon with this change.

mykola-mokhnach:
Thanks @diemol 

It looks like the compatibility with the latest snapshot is already broken: https://github.com/appium/java-client/actions/runs/35845286236/job/107136595938, so only this change won't be enough to keep it compatible

src/main/java/io/appium/java_client/ErrorCodesMobile.java
```diff
@@ -51,21 +51,22 @@ public Class<? extends WebDriverException> getExceptionType(int statusCode) {
...
+     * Returns the exception type that corresponds to the given {@code message}. Messages that do not
+     * match a mobile exception are resolved by the standard W3C error mapping.
      *
      * @param message message An error message returned by Appium server
-     * @return The exception type that corresponds to the provided error message or {@code null} if
-     *     there are no matching mobile exceptions.
+     * @return The exception type that corresponds to the provided error message.
      */
     @Override
     public Class<? extends WebDriverException> getExceptionType(String message) {
```
mykola-mokhnach: please mark message arg as nullable

mykola-mokhnach (approved):

